PR #55, opened 2016-12-09
Add branch name to build-info

luketheobscure:
Found myself looking for this info quite a bit today, and got tired of hitting “back” to figure out what branch I was looking at.

I tried to follow the conventions that I saw in the project. If you would rather I just add it to the template (instead of creating a component), I don’t mind updating.

### DIFF
--- a/app/components/build-branch.js
+++ b/app/components/build-branch.js
@@ -1,0 +1,5 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  build: null
+});
--- a/app/styles/components/_build_header.sass
+++ b/app/styles/components/_build_header.sass
@@ -37,6 +37,7 @@
     .BuildInfo-started,
     .BuildInfo-duration,
     .BuildInfo-comparison,
+    .BuildInfo-branch
       float: left
       padding-left: 12px
 
--- a/app/templates/components/build-branch.hbs
+++ b/app/templates/components/build-branch.hbs
@@ -1,0 +1,1 @@
+{{fa-icon 'code-fork'}} {{build.branch}}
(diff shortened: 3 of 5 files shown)


fotinakis:
This is great, it makes me very happy that you can contribute to this directly. :)

After reviewing the diffs in Percy (yayy perfect use case) — I noticed that we already have branch info on the build page. :/
  
screen shot 2016-12-08 at 11 09 21 pm

BUT, it doesn't show up if the build doesn't have a commit (Percy builds will always have a branch, but might not have a commit).

In `build-header.hbs` you'll find this:
```hbs
{{#if build.commit}}
  {{commit-info build=build}}
{{/if}}
```
...it's safe to remove that if-statement and always show the commit-info, which should give you the info you need. Feel like remaking this PR to just remove that if-statement?

luketheobscure:
Interesting. Looks like Percy isn't getting commit info from non-github projects. Removing that `if` (or updating the `if` to test for `build.branch`) is probably a step in the right direction. 

Thoughts on the icon? Seems weird to have the octocat next to a non-github branch name. Update it to the [code-fork](http://fontawesome.io/icon/code-fork/) icon, or add some conditional logic?

Thanks for making this open source!

fotinakis:
In some cases we get commit info from non-GitHub sources (percy-capybara collects some commit info client-side). So, let's keep the icon for now (if you see in `commit-info.hbs`, we don't show octocat unless the build is from a github source). As we add more support for other services we'll keep that in mind and maybe update to something more generic.

fotinakis:
I'll go ahead and remove the if-statement now, should see branch info on builds in our next release.

fotinakis:
https://github.com/percy/percy-web/commit/3cda6c54f3aa57951114aca02100a851468890a3

luketheobscure:
Thanks for the help. I updated my TeamCity setup, and now I'm getting commit info. If anyone else needs it, here's how to set it up for TeamCity 10:

Team City Setup

fotinakis:
Awesome, thanks! We can actually put together real TeamCity support across our client libraries based on that as well, will put it on the list.

PR #58, opened 2016-12-10
Remove iframe from auth code, use the new /user endpoint instead

mfazekas:


fotinakis:
This is awesome, thanks! 👍  So much simpler. I did a bunch of QA locally, confirmed that all the redirects and OAuth flows still work, I'll do some staging auth QA before fully rolling this out.

PR #59, opened 2016-12-10
Do not care about Safari in private mode

bobisjan:
Hi,

I’ve encountered this when using percy.io on Safari in private mode.

Hope this helps.

Jan


fotinakis:
What a weird behavior! I can confirm that Safari in private mode makes localStorage.setItem throw "QuotaExceededError: Dom exception 22: An attempt was made to add something to storage that exceeded the quota".

Thanks for the fix! 👍

PR #60, opened 2016-12-11
Removed test authenticator

mfazekas:
- use custom authenticator in acceptance tests, no need for test  authenticator
- increased default mocha timeout from 2000 to 4000 ms as it's sometimes failing


tests/helpers/setup-acceptance.js
```diff
@@ -28,6 +27,13 @@ export default function setupAcceptance() {
   });
 }
 
+function authenticateSession(app, sessionData) {
```
mfazekas: Yes we can do that, or just skip it. Then either visit ('/login') route is let authenticated auto redirect to it.

fotinakis: Yeah let's try that, would be nice to not have this duplicated. Another option is you could use [Ember's dependency injection](https://guides.emberjs.com/v2.10.0/applications/dependency-injection/) to copy the authenticator (instead of actually creating test.js and inheriting).  I think it would be fine if some things in tests hit "authenticator:custom" and some hit "authenticator:test" because they'll be the same thing—so we can probably keep the changes to config/environment.js where you drop the APP.authenticator config.

For dependency injection, setupAcceptance could do something like this (untested):
```js
let authenticator = Ember.getOwner(application).lookup('authenticator:custom');
application.register('authenticator:test', authenticator);
```
... to copy the factory to `authenticator:test` so that the simple-auth helpers still work.

Anyway, whatever you think is best! Whatever is the simplest solution works for me, my main goal is just to avoid having to copy/customize the simple-auth helpers here.

fotinakis:
LGTM 👍

PR #61, opened 2016-12-11
Link to comparison

mfazekas:
- added ember-route-action
- used [Intrinsic Placeholders](http://daverupert.com/2015/12/intrinsic-placeholders-with-picture/) via new `image-with-reserved-space` component, so layout does not change as images are loaded - so we can scroll in Ember's didRender.
- keyboard navigation updates queryParams in url
- the title comparison is now a link so url can be copied from there


fotinakis:
@mfazekas check out the percy diffs here, some ratios look weird.

binhums:
I'm getting `Forbidden` on the diffs here.

fotinakis:
Hey @binhums — that's correct, currently Percy builds are access control limited to only specified people, sorry! It's on our roadmap to allow toggling public access.

binhums:
NP. I'm looking forward to this feature :)

mfazekas:
@fotinakis i think this is complete and ready for review now.

app/components/image-spacer.js
```diff
@@ -0,0 +1,26 @@
...
+  image: null,
+  imageClass: '',
+
+  action: null,
+  bubbles: true,
+
+  classNames: ['ImageSpacer'],
+  attributeBindings: ['style'],
+  style: Ember.computed('image.width', 'image.height', function() {
+    let scale = this.get('image.height') * 100.0 / this.get('image.width');
```
fotinakis: Nice job on this! It's deceptively simple, but handles it really well.

mfazekas:
@fotinakis i've tested and seemd to work fine for me.

When clicking on title it did change the url, but it did not change the selectedIndex, so it was kind of split brain situation where url showed the clicked item, but it was not focused and keyboard navigation wasn't consistent with the url, so i've added a (commit)[https://github.com/percy/percy-web/pull/61/commits/38029bdfb53deef24187bf99bb59ee52b94782b9] to address that, pls review.

fotinakis:
A few more bugs I noticed:
- Clicking on the title once works, but if you click again on the same title it doesn't (it removes the border, though the selectedIndex seems to be right).
- When in Overview mode, clicking on a comparison should jump to it in full-width mode, but right now it just jumps to the first one (this was something I added, I think it was working before the last commit but not sure).
- Links don't work when on the 375px view or anything other than the default (because we don't store the state of the responsive dropdown, so when the new page opens it defaults to 1280px and the selected comparison doesn't exist).

fotinakis:
@mfazekas this is fantastic, thanks for implementing this and working through the issues! I'll ship soon, people will be able to link directly to any diff!

PR #62, opened 2016-12-12
Fix times in tests

mfazekas:
* use window.moment instead of moment es6 shim in freeze-moment.js
* use moment() calls in dynamic attributes in mirage factories so they pick up the frozen times

Should fix the time related flakyness in percy screenshots


mirage/factories/build.js
```diff
@@ -1,10 +1,11 @@
 import { Factory } from 'ember-cli-mirage';
+import moment from 'moment';
 
 export default Factory.extend({
   branch: 'master',
   state: 'finished',
   totalComparisonsDiff: 8,
   totalComparisonsFinished: 12,
-  createdAt: new Date().getTime(),
+  createdAt() { return moment() },
```
fotinakis: For compatibility, I think all factory methods should call `.toDate()` on the Moment objects, so that they return Date objects instead of Moments. That way, we can still use the `freezeMoment` helper but won't have to worry about moment compatibility with ember-data or how we reference those attributes in templates.

Here and all date fields, including the `startedProcessingAt` fields below.

fotinakis: Ah, yes, that makes total sense. Good point!

PR #68, opened 2017-01-11
ESLint config to braces

mfazekas:


mfazekas:
> 👍 Nice! Let's also set "space-in-parens": ["error", "never"] to do the same thing for whitespace padding inside parens.

i've added that too

fotinakis:
Awesome, thanks! 👍

PR #69, opened 2017-01-13
fastboot

mfazekas:
Summary of changes:

- added _FASTBOOT variables to adapter, as in fastboot there is no `http-proxy` so we're routing our requests to api hosts directly.
- `restore` in authenticator does not call user endpoint in api anymore. In non fastboot we do a re-render once the promise returned by `restore` resolves. On fastboot there is no rerender so `restore` should be an immediate thing.
-  `currentUser` is now based on ember-simple-auth's [currentUser patterns](https://github.com/simplabs/ember-simple-auth/blob/master/guides/managing-current-user.md), this is needed as session does not always have user immediately (see on `restore`). This was the biggest change replacing `currentUser: Ember.computed.alias('session.data.authenticated.user')` with `currentUser: Ember.inject.service()` and we also need to use `currentUser.user` instead of just `user`
-  some refactorings in authentication
   - new `github-login` route to initiate redirect to github oauth flow as we cannot use `window.location=...` in case of fastboot
   - new `auth/failure` route to catch failures in auth
   - new `auth/success` route which is called on sucessfull authentication
   - we also redirect save ESA `attemptedTransition` to cookie which is used by ESA at `auth/sucess` to redirect to the url before the auth

### DIFF
--- a/app/authenticators/custom.js
+++ b/app/authenticators/custom.js
@@ -5,68 +5,73 @@ import BaseAuthenticator from 'ember-simple-auth/authenticators/base';
 export default BaseAuthenticator.extend({
   store: Ember.inject.service(),
   analytics: Ember.inject.service(),
-  restore() {
-    let store = this.get('store');
+  fastboot: Ember.inject.service(),
+
+  restore(data) {
     return new Ember.RSVP.Promise((resolve, reject) => {
-      store.queryRecord('user', {}).then((userRecord) => {
-        if (window.Intercom) {
-          window.Intercom('update', {
-            name: userRecord.get('name'),
-            email: userRecord.get('email'),
-            created_at: userRecord.get('createdAt').getTime() / 1000,
-          });
+        if (!this._validate(data)) {
+          reject();
+        } else {
+          resolve(data);
         }
-        this.get('analytics').identifyUser(userRecord);
-        resolve({user: userRecord});
-      }, reject);
-    });
+      });
   },
   authenticate(options) {
     let store = this.get('store');
     return new Ember.RSVP.Promise((resolve) => {
       store.queryRecord('user', {}).then((userRecord) => {
-        if (window.Intercom) {
-          window.Intercom('update', {
-            name: userRecord.get('name'),
-            email: userRecord.get('email'),
-            created_at: userRecord.get('createdAt').getTime() / 1000,
-          });
+        if (!this.get('fastboot.isFastBoot')) {
+          if (window.Intercom) {
+            window.Intercom('update', {
+              name: userRecord.get('name'),
+              email: userRecord.get('email'),
+              created_at: userRecord.get('createdAt').getTime() / 1000,
+            });
+          }
         }
         this.get('analytics').identifyUser(userRecord);
-        resolve({user: userRecord});
+        resolve({_user: userRecord});
       }, () => {
         // Build params if given a custom final redirect location.
         var finalRedirect;
         options = options || {};
-        if (options.redirectTo) {
-          var parser = document.createElement('a');
-          parser.href = window.location.href;
-          parser.pathname = '/login';
-          parser.search = '?redirect_to=' + encodeURIComponent(options.redirectTo);
-          finalRedirect = parser.href;
-        } else {
-          finalRedirect = '/login';
+
+        if (!this.get('fastboot.isFastBoot')) {
+          if (options.redirectTo) {
+            var parser = document.createElement('a');
+            parser.href = window.location.href;
+            parser.pathname = '/login';
+            parser.search = '?redirect_to=' + encodeURIComponent(options.redirectTo);
+            finalRedirect = parser.href;
+          } else {
+            finalRedirect = '/login';
+          }
+          // Redirect to GitHub auth.
+          window.location = utils.buildApiUrl('login', {params: {redirect_to: finalRedirect}});
         }
-        // Redirect to GitHub auth.
-        window.location = utils.buildApiUrl('login', {params: {redirect_to: finalRedirect}});
       });
     });
   },
   invalidate() {
     return new Ember.RSVP.Promise((resolve, reject) => {
-      Ember.$.ajax({
-        type: 'GET',
-        url: utils.buildApiUrl('logout')
-      }).done((data) => {
-        // If a user clicks Logout, make sure we clear all the persistent storage locations.
-        this.get('analytics').invalidate();
-        if (window.localStorage) {
-          window.localStorage.clear();
-        }
-        resolve(data);
-      }).fail(function(xhr) {
-        reject(xhr);
-      });
+      if (!this.get('fastboot.isFastBoot')) {
+        Ember.$.ajax({
+          type: 'GET',
+          url: utils.buildApiUrl('logout')
+        }).done((data) => {
+          // If a user clicks Logout, make sure we clear all the persistent storage locations.
+          this.get('analytics').invalidate();
+          if (window.localStorage) {
+            window.localStorage.clear();
+          }
+          resolve(data);
+        }).fail(function(xhr) {
+          reject(xhr);
+        });
+      }
     });
+  },
+  _validate(data) {
+    return data._user && data._user.login;
   }
 });
--- a/app/authenticators/custom.js
+++ b/app/authenticators/custom.js
@@ -48,6 +48,8 @@ export default BaseAuthenticator.extend({
           }
           // Redirect to GitHub auth.
           window.location = utils.buildApiUrl('login', {params: {redirect_to: finalRedirect}});
+        } else {
+          return Ember.getOwner(this).lookup('router:main').transitionTo('github-login');
         }
       });
     });
--- a/app/components/forms/organization-user-edit.js
+++ b/app/components/forms/organization-user-edit.js
@@ -9,10 +9,9 @@ export default BaseFormComponent.extend({
   classNameBindings: ['classes'],
 
   session: Ember.inject.service(),
-  currentUser: Ember.computed.alias('session.data.authenticated.user'),
-  isCurrentUser: Ember.computed('organizationUser', 'currentUser', function() {
-    // TODO: why can't we use Ember.computed.equal for this?
-    return this.get('organizationUser.user.id') === this.get('currentUser.id');
+  currentUser: Ember.inject.service(),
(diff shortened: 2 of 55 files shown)


config/environment.js
```diff
@@ -6,6 +6,9 @@ module.exports = function(environment) {
     environment: environment,
     rootURL: '/',
     locationType: 'auto',
+    fastboot: {
+      hostWhitelist: ['percy.io', /^localhost:\d+$/]
```
mfazekas: we should add PERCY_WEB_API_HOST here

app/mixins/reset-scroll.js
```diff
@@ -3,9 +3,11 @@ import Ember from 'ember';
 var ResetScrollMixin = Ember.Mixin.create({
   actions: {
     didTransition() {
-      window.scrollTo(0, 0);
+      if (window.scrollTo) {
```
fotinakis: Check for existence of `window` first? `window && window.scrollTo`

app/routes/application.js
```diff
@@ -13,6 +14,19 @@ export default Ember.Route.extend(ApplicationRouteMixin, {
     }
   },
 
+  beforeModel() {
+    return this._loadCurrentUser();
+  },
+
+  sessionAuthenticated() {
+    this._super(...arguments);
+    this._loadCurrentUser().catch(() => this.get('session').invalidate());
```
fotinakis: What is this line for?

app/routes/application.js
```diff
@@ -21,11 +35,14 @@ export default Ember.Route.extend(ApplicationRouteMixin, {
       this.transitionTo('login');
     },
     redirectToDefaultOrganization() {
-      let lastOrganizationSlug = localStorage.getItem('lastOrganizationSlug');
+      let lastOrganizationSlug;
+      if (!this.get('fastboot.isFastBoot')) {
```
fotinakis: Can you test for existence of `localStorage` instead?

app/authenticators/custom.js
```diff
@@ -4,37 +4,32 @@ import BaseAuthenticator from 'ember-simple-auth/authenticators/base';
 
 export default BaseAuthenticator.extend({
   store: Ember.inject.service(),
-  restore() {
-    let store = this.get('store');
+  fastboot: Ember.inject.service(),
```
fotinakis: Needs a careful merge of conflicts in this file.

config/environment.js
fotinakis: Also `canary.percy.io` and `dev.percy.local` could be hard-coded?

mfazekas:
@fotinakis i don't know much about production hosting other than it obviously requires a node server. I think your best fit is probably as a node.js middleware
https://ember-fastboot.com/docs/deploying

We should also have fastboot hosted from same domain (like percy.io) as our api. If the domains are different cookies are not passed around without some extra headers - see https://www.youtube.com/watch?v=jcAgi7fpTw8&t=14m25s

app/routes/application.js
mfazekas: So we've used to have get the currentUser in session restore. Now session restore just restores the session, assuming the auth is still valid. But if we cannot get the currentUser then we need to invalidate the session.

mfazekas: hmm actually i need to revisit this. The code is frome here:

https://github.com/simplabs/ember-simple-auth/blob/master/guides/managing-current-user.md#loading-the-current-user

but `sessionAuthenticated()` doesn't seems to be called on restore just on full auth, i'll look into it a bit more.

mfazekas:
@fotinakis pls review this should be complete now.

I've tweaked `currentUser` pattern a bit, so if restore is succesful, but backed returns error for `currentUser` we still invalidate the sessions. (i've also submitted a PR to improve ESA doc https://github.com/simplabs/ember-simple-auth/pull/1176)

i've also added some new percy + acceptance test which should cover 90% of `session.data.authenticated.user` => `currentUser.user` changes

app/routes/application.js
mfazekas: I think the pattern described at [ESA](https://github.com/simplabs/ember-simple-auth/blob/master/guides/managing-current-user.md#loading-the-current-user) wasn't complete. I've submitted a PR (https://github.com/simplabs/ember-simple-auth/pull/1176) and also changed our code accordingly.

Now i've placed `.catch(() => this.get('session').invalidate())` logic to `_loadCurrentUser()`. The idea is that even if we think we have a valid session, backed might disagree and then we need to invalidate our session as well.

app/authenticators/custom.js
mfazekas: `this.get('analytics').identifyUser(userRecord)` now called from `currentUser`

app/components/forms/organization-user-edit.js
```diff
@@ -9,10 +9,9 @@ export default BaseFormComponent.extend({
   classNameBindings: ['classes'],
 
   session: Ember.inject.service(),
-  currentUser: Ember.computed.alias('session.data.authenticated.user'),
-  isCurrentUser: Ember.computed('organizationUser', 'currentUser', function() {
-    // TODO: why can't we use Ember.computed.equal for this?
-    return this.get('organizationUser.user.id') === this.get('currentUser.id');
+  currentUser: Ember.inject.service(),
+  isCurrentUser: Ember.computed('organizationUser', 'currentUser.user', function() {
+    return this.get('organizationUser.id') === this.get('currentUser.user.id');
```
fotinakis: This references the relationship's ID, not the user id. It should be `organizationUser.user.id`

app/components/nav-menu.js
```diff
@@ -5,7 +5,6 @@ export default Ember.Component.extend(TargetApplicationActionsMixin, {
   user: null,
   classes: null,
 
-  session: Ember.inject.service(),
```
fotinakis: Don't you need to inject the `currentUser` service here? Not sure how the nav is working currently.

app/routes/login.js
```diff
@@ -6,18 +6,21 @@ import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-
...
 
   afterModel() {
     // Right now location.href is the URL before the transition to the login route has completed.
     // Make sure to pass redirectTo into the authenticator so that we come back to that URL.
     let options = {};
-    if (!window.location.href.endsWith('/login')) {
-      options['redirectTo'] = window.location.href;
-    } else {
-      options['redirectTo'] = '/';
+    if (!this.get('fastboot.isFastBoot')) {
```
fotinakis: Drop the fast-boot specific check, I think you can just do a `window` specific check like `if (window && window.location && !window.location.href.endsWith('/login')) {`

app/services/current-user.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+
+export default Ember.Service.extend({
```
fotinakis: By convention, add `user: null,` as a property on this service to indicate that `user` is something that exists on this object.

tests/acceptance/organization-test.js
```diff
@@ -141,6 +171,29 @@ describe('Acceptance: Organization', function() {
...
+
+    it('pricing', function() {
+      visit('/login');
+      visit('/pricing');
+      andThen(() => expect(currentPath()).to.equal('pricing'));
+      percySnapshot(this.test);
+
+      click('.PricingSection-bucket:contains("Small") a:contains("Select Plan")');
+      andThen(() => expect(currentPath()).to.equal('pricing'));
+      percySnapshot(this.test + '| select organization');
```
fotinakis: Needs to be `this.test.fullTitle() +`

tests/acceptance/organization-test.js
```diff
@@ -141,6 +171,29 @@ describe('Acceptance: Organization', function() {
...
+
+      /* disabled as sometimes fails with
```
fotinakis: Probably just drop this entire comment block then.

app/components/nav-menu.js
mfazekas: we're calling nav menu like this, so no currentUser needed:

```es6
  {{nav-menu user=currentUser.user}}
```

app/components/forms/organization-user-edit.js
mfazekas: good catch, sorry for that i'm sure i've verified with mirage in debugger the only issue is that in mirage most of the stuff has id=1...

app/services/current-user.js
mfazekas: i've added user: null

tests/acceptance/organization-test.js
```diff
@@ -141,6 +171,17 @@ describe('Acceptance: Organization', function() {
         return percySnapshot(this.test.fullTitle() + ' | invalid modification');
       });
     });
+
+    it('pricing', function() {
```
fotinakis: By convention, this should have a verb in it so it reads like a sentence — like "it can select a plan on the pricing page" or something like that.

mfazekas:
@fotinakis i've removed factory-guy as we seems to have not used it at all, and it is not compatbile with fastboot - https://github.com/percy/percy-web/commit/f6d21dd3eee52700f4da72786013051698dd77a7

mfazekas:
Ready for review now. Not sure what's the issue with percy, as buildkite seems to have built just fine.

timhaines:
@mfazekas Thanks for the updates here.  I've tried exercising it a little today.  It works very nicely on the public pages.  I've found it fails to load the pages on the signed in pages though.  Are those loading for you okay?

Here's an example stacktrace:
image

mfazekas:
@timhaines i think i've addressed that issue. I'm a bit unsure as i cannot get this code path ran from fastboot now.

timhaines:
@mfazekas That error has gone now yes.  Nice one!   Thanks for your work on this.

It seems to hang if you try and load a build page when you're signed out, and doesn't redirect to the login page/process as you'd expect.
 
There's another error (a complaint about localstorage) when viewing a build.

image

mfazekas:
@timhaines thanks for looking into it. My last commit should solve these two issues. There is one more issue - might be caused by my fix is that when you log in you have to click the "Sign in with github" twice

fotinakis:
Unfortunately I think we need to close this off for now as it's not something we're going to tackle right now. Going to leave it available for reference here for one day when we come back to it.

PR #79, opened 2017-02-23
Add a snapshot showing a new project

timhaines:


tests/acceptance/project-test.js
```diff
@@ -25,6 +25,23 @@ describe('Acceptance: Project', function() {
     });
   });
 
+  context('waiting for first snapshot', function() {
+    setupSession(function (server) {
+      let organization = server.create('organization', 'withUser');
+      let project = server.create('project', {name: 'My Project', organization});
+      server.create('token', {token: 'abc', role: 'write_only', is_active:true, project});
```
fotinakis: These are reasonable defaults for test tokens, you could move the attributes up into the `token` model itself, then this line becomes just `server.create('token', {project})`

PR #81, opened 2017-02-26
Ping intercom with the company id

timhaines:
@fotinakis There's a few things to discuss with this one.  
* We want to ping intercom with the specific company id of the organization a user is looking at.  Intercom's docs say we need to include either an email address or user id with every update call where we change data, so I've included the email address.
* I've chosen to do add the Intercom update in the Organization route.  I *think* this means the event is triggered whenever someone is outside of the organization route (and it's nested routes) and they enter it.  So - on a new page load inside the tree, or going from home page -> organization, but not when moving between builds that are already inside the organization route tree.  Does this seem like a good place to you?  One limitation of this is if the user has a Percy tab open for days, and is bouncing around between builds, we're not pinging Intercom over that period.  Is this okay?  Is there a way to address this?
* Intercom isn't pinged with an org id if the user only visits the home page or docs even if signed in.  If we wanted to do this, we could perhaps look at the lastOrganizationSlug and try and convert that to an organization id, then fire the event.  This seems of dubious value though, as the user's not really looking at the organization (hmmm).


app/routes/organization.js
```diff
@@ -2,11 +2,23 @@ import Ember from 'ember';
...
     try {
       localStorage.setItem('lastOrganizationSlug', model.get('slug'));
     } catch (_) {
       // Safari throws errors while accessing localStorage in private mode.
     }
+
+    let email = this.get('currentUser.email');
+    if (email && window.Intercom) {
+      window.Intercom('update', {
+        email: email,
```
fotinakis: Per discussion, let's include in this PR moving to real user IDs and not relying on email addresses anymore.

timhaines:
@fotinakis I switched from emails to user_ids when talking to intercom.  Take a look at the questions in the description too if you didn't already.  Maybe they're all 👍  ?

fotinakis:
LGTM 👍 Your understanding of the route handling is correct, and this will handle 90% of the cases we care about. You could potentially add this same code (or a service abstraction of it) into the `organizations.organization` route as well, to capture times when someone goes to the org settings page and not to a build.

PR #82, opened 2017-02-28
GitHub Integration - Deprecate Bot User

timhaines:


fotinakis:
Nice! I think you can hide the save button entirely in this case (unless the user is a bot user, or the form is modified moving from github to no access -- you can use `changeset.isDirty` to check if the form is dirty).

PR #83, opened 2017-03-01
Stop sending analytics events for admin users

timhaines:


timhaines:
@fotinakis there might be some discussions to have around naming here, but I think the structure is about right.  Could potentially refactor more intercom type stuff into the intercom service, but -1 on doing that right now.

app/services/analytics.js
```diff
@@ -25,8 +23,18 @@ export default Ember.Service.extend({
     this.organizationInstance.init(config.APP.AMPLITUDE_ORGANIZATIONS_PROJECT_ID);
   },
 
+  isEnabled() {
```
fotinakis: How about:

```
return window.amplitude && !this.get('adminMode').excludeAnalytics();
```

app/services/admin-mode.js
```diff
@@ -0,0 +1,22 @@
+import Ember from 'ember';
+
+export default Ember.Service.extend({
+
+  get() {
```
fotinakis: A slightly more Ember-idiomatic way to do this service would be to expose computed properties rather than getters, so that callers can do things like `this.get('adminMode.excludeFromAnalytics')` and `this.get('adminMode.currentMode')` instead of `this.get('adminMode').get()`. I guess the `localStorage` dependency makes it slightly more complicated because state is not entirely stored in the service — but the answer there is to implement the [init](http://emberjs.com/api/classes/Ember.Service.html#method_init) constructor to read state from on boot, rather than having `get()` compute it every time.

So, the API could be:

```js
this.get('adminMode.currentMode')
this.get('adminMode.excludeAnalytics')

this.get('adminMode').setMode('admin')
```

PR #91, opened 2017-03-17
Add note on github integration screen about owner role

timhaines:


timhaines:
Interesting that we have no snapshot for this.

PR #92, opened 2017-03-17
Adds anchor links to documentation headers and handles jumping (#182)

rosschapman:


app/components/docs-renderer.js
```diff
@@ -2,8 +2,27 @@ import Ember from 'ember';
...
 export default Ember.Component.extend({
   markdown: null,
+  isDocsNav: null,
+
+  didInsertElement() {
+    this._super(...arguments);
+
+    if (!this.get('isDocsNav')) {
+      Ember.run.next(this, function() {
+        if (window.location.hash) {
```
fotinakis: Optional: could combine this `if` conditional with the other `if` above ^.

app/components/docs-renderer.js
```diff
@@ -14,9 +33,19 @@ export default Ember.Component.extend({
       return false;
     }
   },
+  _scrollToAnchor(hash) {
+    // Manually resetting the hash seems necessary to "hijack" normal browser anchor jumping
```
fotinakis: Wow, nice find on combining this plus scrollTo — I tried for a minute to simplify this by removing one or the other, but it definitely doesn't work. I'm 👍  on this hack, it seems to work cross browser as well.

rosschapman: Yeah, don't ask me why this works :man_facepalming:

rosschapman:
@fotinakis [this commit](https://github.com/percy/percy-web/pull/92/commits/dba669c2e61f6a2cd4f9738c0e4e8e84754ce4af) adds h3 support and fixes the bug that broke some titles.

timhaines:
@rosschapman I've tried out this PR in Firefox, Safari, and Chrome - works well with the simple testing I did.  Nice one - this is going to be a very helpful addition!

What text editor are you using?  I'm wondering if we can help you get it setup to highlight linting mismatches with the .eslintrc.js and .editorconfig rules we have in place.  This will mean we'll stop seeing CI failures due to linting - and will all be using a consistent style.

Looking forward to getting this merged and shipped once @fotinakis has taken another look. 👍

fotinakis:
@rosschapman going to merge this so I can include it in a release

rosschapman:
@timhaines yep that was on my todo list but hadn't got around to it yet. Just fixed that so I've got Sublime configured properly now to read the linter configs for this repo.

PR #93, opened 2017-03-18
added github-int + pricing page percy tests

mfazekas:


timhaines:
Interestingly it surfaces the fact we're showing the org picker when there's only one organization too (not sure if this is maybe a Mirage issue?) - but if this happens in reality we probably should probably just do the redirection when they only have 1 org, rather than show the picker.  (This is outside the scope of this pull request of course - just noting it as an observation)

PR #94, opened 2017-03-23
Update ember from 2.10 to 2.12.1. Also updates other stale packages.

timhaines:


ember-cli-build.js
```diff
@@ -8,13 +8,15 @@ module.exports = function(defaults) {
...
-      includePolyfill: true,
     },
     'ember-cli-mocha': {
       useLintTree: false
     },
     sassOptions: {
       extension: 'sass',
+    },
+    'ember-cli-babel': {
+      includePolyfill: true,
```
fotinakis: Does this need to have `comments: false` like the `babel` section had above? Can we drop the `babel` section now?

package.json
```diff
@@ -23,46 +23,54 @@
...
+    "ember-source": "^2.12.0",
+    "ember-truth-helpers": "^1.3.0",
+    "ember-watson": "^0.8.5",
+    "glob": "^7.1.1",
+    "http-proxy": "^1.16.2",
+    "loader.js": "^4.2.3",
+    "morgan": "^1.8.1",
     "percy-docs": "git+https://github.com/percy/percy-docs.git#8efc10256a7f3605ea06354165170f0df57123fb"
+  },
+  "dependencies": {
```
fotinakis: Just noting, we haven't had a `dependencies` block before, everything has been `devDependencies`.

tests/helpers/destroy-app.js
```diff
@@ -3,4 +3,7 @@ import Ember from 'ember';
 export default function destroyApp(application) {
   server.shutdown();
```
fotinakis: Drop this then if adding the block below (I think added automatically by ember-cli-mirage).

package.json
timhaines: These were added as a result of 'yarn upgrade package-name' rather than manually.  I think technically the dependencies section should be the default section, and devDependencies should be for packages we only need in dev - but will read up on this.

timhaines:
@fotinakis Cleaned up the shutdown call and package.json dependencies.  The babel comments: false prevents comments from being included into the built js.  If I remove that line, they're included.  We don't want them, so leaving comments: false in place.

PR #99, opened 2017-03-26
Update to ember-2.12.0.  

timhaines:
Tests pass with warnings currently.

App does not let you update the billing plan right now.  Some errors to track down.


timhaines:
The remaining warning when running tests about ember-cli-test-loader is coming from ember-cli-mocha I believe.  Perhaps another update of ember-cli-mocha will be released soon to address this.

timhaines:
The problem with updating billing plans comes from a combo of a couple of different things.  

We have an existing subscription with an id 'xyz' that's an obfuscated organization id.  When the plan is changed, we're creating a new subscription, and when it's saved, it's receiving the id 'xyz' (the same obfuscated organization id).

Ember-Data 2.12.0 is throwing an error when it sees we've now got a second model with the same id as the existing id.

@Runspired tells me this is a known bug in Ember-Data, that usually surfaces in different ways to what we're seeing.  He mentioned that Ember-Data should tolerate the new record coming back with an existing id, and merge/overwrite the existing internal model.

Will look into some options for how to deal with this soon.  Some options off the top of my head: 
- Don't upgrade until the bug is fixed.
- Switch from creating a new record to updating the existing one (as far as ED is concerned).
- Perhaps delete the old record before creating the new one?
- Switch to using a different, unique id system for subscriptions on the client.

timhaines:
Have locked ember-data to 2.11.3.  All tests pass, no visual diffs, have had no problems with viewing and changing data.

PR #103, opened 2017-03-27
Batches "no diff" snapshots in comparison list [ch98]

rosschapman:


timhaines:
Couple of early thoughts - there's an unintended ordering different on snapshots right now - alphabetical vs number of pixels changed (discussed briefly in Slack).

Would be good to handle the '1 unchanged snapshot' vs '2 unchanged snapshots' pluralization of snapshot too.

image

app/components/comparison-list.js
```diff
@@ -18,8 +18,14 @@ export default Ember.Component.extend({
     return hasNoDiffs && (activeComparisonIsNoDiff == undefined);
   }),
   diffComparisons: Ember.computed.filterBy('sortedComparisons', 'isDifferent'),
-  noDiffComparisonsCount: Ember.computed(function() {
-    return this.get('comparisons').filterBy('isSame').length;
+  noDiffComparisonsCountTxt: Ember.computed(function() {
```
fotinakis: Alternatively: you could drop this and just use the ``{{pluralize`` helper in the template. Check out the other usages around.

app/components/comparison-list.js
```diff
@@ -13,7 +13,7 @@ export default Ember.Component.extend({
 
   hideNoDiffs: Ember.computed('comparisons', 'activeComparisonId', function() {
     let noDiffs = this.get('comparisons').filterBy('isSame');
-    let hasNoDiffs = noDiffs.length >= 0;
+    let hasNoDiffs = noDiffs.length > 0;
```
timhaines: good catch

fotinakis:
(I broke the tests, one min)

PR #104, opened 2017-03-30
[DON'T MERGE] Find the breaking upgrade

timhaines:
Trying to locate the package upgrade that prevented 90 snapshots from running last week.


timhaines:
Deleting

PR #109, opened 2017-04-14
Adds responsive UI improvements for comparison viewing 

rosschapman:


rosschapman:
@fotinakis I'm still working out some test failures but I think this is a good point to get you et al  involved in code review as well as start visual QA of the behavior to ensure it conforms to all the specs from https://docs.google.com/document/d/1DAGDU6HXBicwcvXyvmh0fFNG3Z6gET-771WstIWQqyY/edit#heading=h.hw5j5d8bz0hm. Let me know how you want to proceed with visual QA -- maybe we sit down sometime next week and click through together?

timhaines:
@rosschapman Just gave it a spin Ross.  Very nice!  It definitely had a "Awesome, now I know what's going on at all widths!" immediate impact on me.

This message feels like it could do with a little tweaking now when there's multiple widths and potentially more snapshots being revealed when you click it.  What do you think?
image

I wonder if it would make sense to call a snapshot 1 snapshot, regardless of how many widths it includes?  So previously percy-web had 106 snapshots - now it maybe has 53 snapshots, and each snapshot has 2 comparisons?  The total summary could still state how many snapshots have changed..

timhaines:
image

Also wondering now if the red dot on the mostly white bar actually gives us much useful info.  Maybe the whole bar should just be red if changes, white if no changes, or grey if no comparison at that width?

fotinakis:
Yeah I kind of agree the progress bar doesn't really seem to provide useful info now that we see it, maybe solid colors are more clear (but we can always adjust that more later, since it may change the design of this more generally).

rosschapman:
Keep the feedback coming. I'm at Gray Area all day for a conference. How about we regroup Monday?

I also just found a bug where changing the width in the header doesn't force all snapshots to reset to that selection -- need to fix up the computed dependency chain.

timhaines:
Maybe we could do something like this rather than the progress-like bars:

image

timhaines:
@rosschapman Have been clicking around on some customer's build using this, and it is very very awesome.  I'm super excited about shipping this out to all of our users.  :)

app/components/snapshot-list.js
```diff
@@ -0,0 +1,162 @@
...
+  buildContainerSelectedWidth: null,
+  buildWidths: [],
+  lastSnapshotIndex: null,
+  selectedNumColumns: null,
+  selectedSnapshotIndex: -1,
+  snapshotComponents: null,
+  snapshots: [],
+  updateActiveSnapshotId: null,
+
+  sortedSnapshots: Ember.computed('snapshots', function() {
```
fotinakis: You could use `Ember.computed.sort('snapshots', function(a, b) {` to slightly simplify things here.

app/components/snapshot-viewer-width-selector-control.js
```diff
@@ -0,0 +1,25 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  selectedWidth: null,
+  width: null,
+  comparisons: [],
```
fotinakis: Since this is the single control selector, is there a way we can refactor this component to only accept a single `comparison` instead of all of them? That would simplify a bunch of the logic below as well because you do a bunch of re-finding the correct comparison by width.

app/components/snapshot-viewer-width-selector.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
```
fotinakis: Possible component renaming:

* `comparison-switcher` (instead of `snapshot-viewer-width-selector`)
* `comparison-switcher-item` (instead of `snapshot-viewer-width-selector-control`)

I'm happy to quickly do this renaming work if you 👍  it.

app/components/snapshot-viewer.js
```diff
@@ -0,0 +1,76 @@
...
+  snapshot: null,
+  
+  classNames: ['SnapshotViewer'],
+  buildContainerSelectedWidth: null,
+  snapshotSelectedWidth: Ember.computed.oneWay('buildContainerSelectedWidth'),
+  selectedComparison: Ember.computed('snapshot', 'snapshotSelectedWidth', function() {
+    let width = this.get('snapshotSelectedWidth');
+    return this.get('snapshot.comparisons').findBy('width', parseInt(width, 10));
+  }),
+  totaldiffPercentage: Ember.computed('snaphot', function() {
```
fotinakis: `totaldiffPercentage` --> `totalDiffPercentage`

app/models/build.js
```diff
@@ -41,16 +41,11 @@ export default DS.Model.extend({
   snapshots: DS.hasMany('snapshot', {async: true}),
 
   comparisonWidths: Ember.computed('comparisons', function() {
-    // TODO(fotinakis): use a JavaScript Set when the world has advanced.
```
fotinakis: lol thanks for taking care of this, I guess the world has advanced now

app/components/snapshot-viewer-width-selector.js
fotinakis: @rosschapman I pushed this rename up, you'll want to pull this PR locally before adding more things.

timhaines:
This screenshot is from what's currently in production:
image

What happens with new snapshots now?  Do they just appear in red text like any other changes?  (Would be reasonable I think)

PR #120, opened 2017-05-11
New Project Screen Updates

djones:
* Simplifies project setup screen.
* Improves the visuals of the setup screen
* Splits out the demo project instructions and better explains what is going on.
* Code format UI updated

### DIFF
--- a/app/components/project-container.js
+++ b/app/components/project-container.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   project: null,
-
+  showQuickstart: false,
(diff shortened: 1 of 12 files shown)


timhaines:
@djones Can you add a test and snapshot that shows the Demo Project Instructions expanded please?  So we can capture that in Percy too.

fotinakis:
@djones there's already a good acceptance test for this state too, you could easily extend to click the button and take another snapshot: https://github.com/percy/percy-web/blob/master/tests/acceptance/project-test.js#L36

PR #121, opened 2017-05-12
Upgrade ember to 2.13.2.

fotinakis:


ember-cli-build.js
```diff
@@ -6,7 +6,9 @@ module.exports = function(defaults) {
   var app = new EmberApp(defaults, {
     babel: {
       comments: false,
-      includePolyfill: true,
+    },
+    'ember-cli-babel': {
+      disableDebugTooling: true
```
rwjblue: I'm curious why this needs to be disabled.  Just wondering if there may be a bug that we need to fix...

fotinakis: @rwjblue going to revert this, this was me experimenting to fix a perf regression bug.

For some reason starting around 2.11 and continuing until now, our build time has gone from ~10s to 150s+ (on all machines in all environments) and I haven't been able to trace the cause yet — I thought upgrading to the new ember with babel 6 might help, but no joy, so I was experimenting here. @runspired suggested I run things through the [heimdalljs-visualizer](https://github.com/rwjblue/heimdalljs-visualizer) and I learned that babel is 95% of the time spent, but I'm not sure what the root cause is yet.

rwjblue: Yikes, thats horrible! Please let me know if you need help figuring out what is going on...

fotinakis: @rwjblue thanks for the offer! Was about to take you up on it but we able to fix the slowdown, though still not sure what package was causing it. Upgrading all our transitive deps fixed it (#123).

rwjblue: Awesome!

fotinakis:
@timhaines Ember 2.13 upgrade is ready to go. Only caveat is that upgrading ember-data to 2.13 breaks some flows (in the way we chatted about this week with ID re-use). It's fine to leave ember-data at a lower version for the moment, but we'll want to get that fixed and upgraded at some point.

fotinakis:
Oof actually not ready for review here.

timhaines:
@fotinakis The new production build stuff seems to slow the test run down considerably.  Is it easy to reorder it so it happens after the test run in build kite?

image

fotinakis:
@timhaines Thanks for refreshing this, I did a full QA pass as well and things look great. Merging! (We'll need to catch up ember-data at some point and fix the actual issues with our use of it).

PR #123, opened 2017-05-16
Fix build times and upgrade packages.

fotinakis:
We had some package that was causing our builds to be 20X slower, but still unsure which one it is so this recreates the yarn.lock and our dependency tree from scratch and upgrades all of our transitive dependencies.

Tested: I've done a QA pass through the app, all is well. This does surface a bunch of new deprecation warnings, but they can be safely ignored for the moment.


timhaines:
Takes 23 seconds for a test run now.  Fab.

PR #124, opened 2017-05-18
Fix Deprecations

timhaines:


timhaines:
This PR fixes the deprecations that were getting logged in test output.  The output is less noisy now - but still contains a bunch of warnings about how we're using mirage.

PR #125, opened 2017-05-18
Package updates

timhaines:
Updating ember-cli-htmlbars still causes problems.  Might have to dig into that one sometime.


package.json
```diff
@@ -26,11 +26,11 @@
     "ember-ajax": "^2.5.6",
     "ember-changeset": "^1.2.3",
     "ember-changeset-validations": "^1.2.6",
-    "ember-cli": "2.12.0",
+    "ember-cli": "^2.13.1",
```
fotinakis: Let's not upgrade ember-cli independently of upgrading ember source — would love to keep them in sync so we never have to worry about any incompatibilities. I think same with the ember-cli deps below you've changed, I think it's fine to upgrade them now if you want to, but it may make more sense to wait for #121

PR #126, opened 2017-05-23
New decoupled asset deployment pipeline.

fotinakis:
Fixes [ch350] and deployment pipeline.

* For production, compile assets with a prepended `/static/` path. Also add fingerprints to SVG filenames.
* Configured production load balancer to serve `percy.io/static/*` from GCS with Google Cloud CDN enabled for another level of caching.
* Add `deploy-assets` makefile target that uses `gsutil` to upload files to storage.


fotinakis:
@djones would love to show you this working in person for review.

fotinakis:
@djones Ready for review here! I've verified that this will work in production and in our new more automated build pipelines. Will book some time with you tomorrow to go through this.

timhaines:
I'm interested in staying up to date with how this works.  Please include me when you schedule the discussion @fotinakis.

fotinakis:
Sounds good, added you to the meeting.

ember-cli-build.js
```diff
@@ -15,7 +15,11 @@ module.exports = function(defaults) {
     },
     sassOptions: {
       extension: 'sass',
-    }
+    },
+    fingerprint: {
+      extensions: ['js', 'css', 'png', 'jpg', 'gif', 'map', 'svg'],
```
fotinakis: Per discussion, let's fingerprint everything including fonts.

fotinakis: Turns out we can't do this because ember-font-awesome has it's own internal cache busting mechanism with query params (`src: url(/fonts/fontawesome-webfont.eot?v=4.4.0);`), so we need to not fingerprint those yet (but we can still update them and it'll be fine since the CDNs will all cache including the query param).

PR #129, opened 2017-05-29
[WIP] Story 111 view full

rosschapman:
Bugs/Design:
- [x] Tweak header bar spacing
- [x] When `New snapshot, nothing to compare yet` text needs to be centered and styled
- [x] Hide expand button in Overview Mode
- [x] Fix header to top
- [x] Design parity between current snapshot header
- [x] Don't reuse snapshot viewer header, but abstract out shared styles
- [x] Change "screenshot" to "comparison mode" or "mode"
- [x] Change "new, base, diff" to "base, head, diff"
- [x] Preserve mode state from snapshot-viewer to full-view 
  In snapshot-viewer only switch between "head" and "diff" mode)
- [x] One liner `cycleMode` with mod
- [x] Figure out if smaller width screenshots should be full-screen (probably)
- [x] Don't duplicate but create new component for comparison viewer templates (ie snapshotmissing)
 
Features:
- [x] `Esc` should exit full view 
- [x] Arrow keys should cycle screenshots
- [x] Add width to url for hyperlinking all the way

From June 8:

- [x] When clicking fullscreen button, "Diff" should be the default state.
- [x] Missing image: Diff mode should overlay the diff on the head image in fullscreen mode (same as current UI).
- [x] Missing interaction: clicking on the image should rotate through to the next mode. Since the image doesn't exist in some states ("new snapshot"), I think the best UX is clicking anywhere in the entire lower section (via a transparent overlay) should rotate through the images. The cursor should also be a pointer.
- [x] Missing style: the same style of simple shadow border should surround the images.
- [x] Broken scrollbar: for some reason, the horizontal scrollbar is always at the bottom of the page (ie. you have to scroll down to see it) when the window is smaller.
- [x] State issue: escaping the mode should drop you back to the same scroll position as you were in before you entered it. Think of this flow: "what is that diff? ... zoom in ... ok got it ... zoom out"
- [x] UI nitpick: the fullscreen button should be square, and be all the way to the right of the box and have even margins.
- [x] Ordering and naming: How about in the UI we call them "Original" "Diff" and "New" (instead of "Base" "Diff" and "New Version" — just talking about the user-centric UI words here, in the codebase we should consistently call them `base`/`head`/`diff`). Also we'd like to swap the order in the UI to be "Original" on the left, "Diff" in the middle, and "New" on the right. Reason being, when you land on "Diff" to start you click and get to "New" immediately, which maps with what've trained people to do in the build UI.


rosschapman:
@fotinakis this is ready for a test drive. Let's chat after you take a look.

app/templates/components/snapshot-viewer-full.hbs
```diff
@@ -0,0 +1,38 @@
+<div class="Viewer-header SnapshotViewerFull-header row">
```
rosschapman: @fotinakis what do you think about declaratively composing classes like this? I think it may be a departure from your "single-class" encapsulation paradigm but because there are multiple elements that need sharing it actually seemed like less confusing code to create multiple mixins for `Viewer-header` and `Viewe-header-title` vs just being able to declare those classes in the template.

timhaines:
@rosschapman This is feeling awesome.  Had a quick play and it's easy to move around through the various options.  I noticed pushing `esc` results in scrolling to the to top of the list when the full screen closes, which is unintended behavior I think.  Closing full screen with the close icon seems to preserve the scroll position correctly.

Can we get some Percy snapshots added to the tests that show the pop-up in a few different states please?

timhaines:
image

Two lines of text fit okay in the header.

rosschapman:
@timhaines yep I'll work on some tests. And yeah, long titles are not looking terrible.

fotinakis:
Nice!

Clicking through the UI, some comments:
- When clicking fullscreen button, "Diff" should be the default state.
- Missing image: Diff mode should overlay the diff on the head image in fullscreen mode (same as current UI).
- Missing interaction: clicking on the image should rotate through to the next mode. Since the image doesn't exist in some states ("new snapshot"), I think the best UX is clicking anywhere in the entire lower section (via a transparent overlay) should rotate through the images. The cursor should also be a pointer.
- Missing style: the same style of simple shadow border should surround the images.
- Broken scrollbar: for some reason, the horizontal scrollbar is always at the bottom of the page (ie. you have to scroll down to see it) when the window is smaller.
- State issue: escaping the mode should drop you back to the same scroll position as you were in before you entered it. Think of this flow: "what is that diff? ... zoom in ... ok got it ... zoom out"
- UI nitpick: the fullscreen button should be square, and be all the way to the right of the box and have even margins. Ie closer to this spacing in the header:
screen shot 2017-06-08 at 2 24 56 pm
vs.  
screen shot 2017-06-08 at 2 25 12 pm

- Ordering and naming: talked through the naming with @djones and I think we came up with some more intuitive names for the modes. How about in the UI we call them "Original" "Diff" and "New" (instead of "Base" "Diff" and "New Version" — just talking about the user-centric UI words here, in the codebase we should consistently call them `base`/`head`/`diff`). Also we'd like to swap the order in the UI to be "Original" on the left, "Diff" in the middle, and "New" on the right. Reason being, when you land on "Diff" to start you click and get to "New" immediately, which maps with what've trained people to do in the build UI.

app/templates/components/snapshot-viewer-full.hbs
fotinakis: Yeah this is a nice way to share the classes between the two states 👍

app/components/comparison-viewer-full.js
```diff
@@ -3,10 +3,13 @@ import Ember from 'ember';
 export default Ember.Component.extend({
   classNames: ['ComparisonViewerFull'],
   comparison: null,
-  isBase: Ember.computed.equal('comparisonMode', 'base'),
+  isOriginal: Ember.computed.equal('comparisonMode', 'original'),
```
fotinakis: This may be counter-intuitive but I think we should not have the word "original" in the codebase at all, except for where it's presented to users in the UI from the template. We already consistently use "base" and "head" everywhere to refer to the comparison's before/after states, so even though we say "Original" and "New" in the UI we should keep this as `isBase`, here and elsewhere, in the `galleryMap`, etc. This naming decoupling also gives us the flexibility to rename the modes in the UI later from "Original" to something else, without having to refactor the underlying code.

rosschapman:
@fotinakis a few clarifications. 

- By "shadow border" you mean box-shadow on the images when in full screen mode? So the images would not be entirely flush with viewport? 
- Can you send me a url where I can test the broken scrollbar? Not sure what you mean exactly.

timhaines:
Heads up that 16 of the existing snapshots aren't running.

rosschapman:
@fotinakis fixed broken tests and added a few more. There's a bit of strange routing behavior I had to compensate for in build-test.js. Maybe you have some thoughts about it. Otherwise ready for another review.

timhaines:
@rosschapman Found a wee bug.  In the 2 column view, if you scroll down and try and rotate through diff and new screenshot, you get scrolled to the top of the list.

timhaines:
@rosschapman Pushing the `esc` key when viewing a single snapshot presents the header and a blank screen, rather than the list of snapshots.

timhaines:
@rosschapman @fotinakis I noticed that if you open the single snapshot view, scroll down, close it, then scroll down the list and reopen the single snapshot view for another snapshot, the single snapshot view is already partially scrolled down.  I think the expected behavior would probably be to scroll the single snapshot view to top each time it's reopened?

app/components/build-container.js
```diff
@@ -1,38 +1,23 @@
...
-export default Ember.Component.extend({
+export default Ember.Component.extend(ScrollableMixin, {
   build: null,
   activeSnapshotId: null,
   updateActiveSnapshotId: null,
   classNames: ['BuildContainer'],
   maxWidth: Ember.computed.max('build.comparisonWidths'),
   buildContainerSelectedWidth: Ember.computed.oneWay('maxWidth'),
   noWidthSelected: false,
+  currentPos: null,
```
fotinakis: `currentPos` --> `currentPosition`

app/mixins/scrollable.js
```diff
@@ -0,0 +1,29 @@
...
+    };
+
+    Ember.$(document).on('touchmove.scrollable', onScroll);
+    Ember.$(window).on('scroll.scrollable', onScroll);
+  }.on('didInsertElement'),
+
+  unbindScrolling: function() {
+    Ember.$(window).off('.scrollable');
+    Ember.$(document).off('.scrollable');
+  }.on('willDestroyElement'),
```
fotinakis: Use `Ember.on` like: `unbindScrolling: Ember.on('willDestroyElement', function() { ...` instead of prototype extensions. Here and elsewhere in this file.

app/mixins/scrollable.js
```diff
@@ -0,0 +1,29 @@
+import Ember from 'ember';
+
+export default Ember.Mixin.create({
+  scrollingTimeout: 100,
+  bindScrolling: function() {
+    let self = this,
```
fotinakis: Could remove the `self` scope hack by using an arrow function.

app/models/build.js
```diff
@@ -38,7 +38,12 @@ export default DS.Model.extend({
   commit: DS.belongsTo('commit', {async: false}),  // Might be null.
   baseBuild: DS.belongsTo('build', {async: false, inverse: null}),
   comparisons: DS.hasMany('comparison', {async: true}),
-  snapshots: DS.hasMany('snapshot', {async: true}),
+
+  snapshots: Ember.computed('comparisons', function() {
```
fotinakis: Nice, this is a fair thing to do until we have a real snapshots API.

app/router.js
```diff
@@ -49,7 +49,9 @@ Router.map(function() {
     this.route('project', {path: '/:project_id'}, function() {
       this.route('settings');
       this.route('builds', {}, function() {
-        this.route('build', {path: '/:build_id'});
+        this.route('build', {path: '/:build_id'}, function() {
+          this.route('snapshot', {path: '/view/:snapshot_id/:width'});
```
fotinakis: Is it easy to make this `:snapshot_id/:comparison_id` instead of `:width`? Using the comparison will be more future proof.

app/routes/organization/project/builds/build.js
```diff
@@ -6,6 +6,7 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
     activeSnapshotId: {as: 'snapshot', replace: true}
   },
   afterModel(model) {
+    debugger
```
fotinakis: Drop.

app/routes/organization/project/builds/build.js
```diff
@@ -37,6 +38,16 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
     },
     updateActiveSnapshotId(snapshotId) {
       this.set('controller.activeSnapshotId', snapshotId);
-    }
+    },
+    showSnapshotFullModal(snapshotId, snapshotSelectedWidth) {
+      Ember.$('.BuildContainer').hide();
```
fotinakis: Per discussion, use a variable here instead of jquery DOM manipulation.

app/routes/organization/project/builds/build.js
```diff
@@ -37,6 +38,16 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
...
+      this.controller.set('trackScroll', false);
+      this.transitionTo(
+        'organization.project.builds.build.snapshot',
+        snapshotId,
+        snapshotSelectedWidth,
+        {queryParams: {m: 'diff'}}
```
fotinakis: Let's use `mode` in the query param instead of `m`, for clarity. No need to obfuscate that.

app/routes/organization/project/builds/build/snapshot.js
```diff
@@ -0,0 +1,56 @@
...
+      this.transitionTo(
+        routeName,
+        build.id,
+        snapshot.id,
+        width,
+        {queryParams: {comparisonMode: comparisonMode}}
+      );
+    },
+    closeSnapshotFullModal(buildId, snapshotId) {
+      Ember.$('.BuildContainer').show();
```
fotinakis: Same here.

app/router.js
rosschapman: It can be done. I went this direction since it's makes the url a bit more meaningful which I think was @timhaines's instinct as well.

rosschapman:
@fotinakis see https://github.com/percy/percy-web/pull/129/commits/8a5c8f2f25c6b9036cea39e269c287bfe3a7ea8f as stop-gap while you explain to me how broccoli works. :)

PR #132, opened 2017-06-14
Add snapshots of width selector, pending builds, and processing builds

timhaines:


app/components/build-container.js
```diff
@@ -45,10 +45,14 @@ export default Ember.Component.extend({
     while (this.get('numPollRequests') < MAX_UPDATE_POLLING_REQUESTS) {
       this.incrementProperty('numPollRequests');
       this.get('build').reload().then((build) => {
-        if (this.get('build.isPending') || this.get('build.isProcessing')) {
-          build.get('comparisons').reload();
-        } else {
-          build.get('comparisons').reload();
+        build.get('comparisons').reload();
+        if (!(this.get('build.isPending') || this.get('build.isProcessing'))) {
```
fotinakis: I believe this will change the logic here — if a build is processing, we WANT the task to keep running and reload until the build is finished.

timhaines: @fotinakis That may mean this is too confusing to read (I don't think it changes the logic).  Would it be better as `if (!this.get('build.isPending') && !this.get('build.isProcessing'))`  ?  Or perhaps even abstracting the `this.get('build.isPending') || this.get('build.isProcessing')` as a `continueReloadingAttempts()` function and then using if (!continueReloadingAttempts())   ?

timhaines:
@fotinakis Refactored to make clearer.  👍

app/components/build-container.js
```diff
@@ -16,6 +16,10 @@ export default Ember.Component.extend({
   selectedNumColumns: 1,
   showComparisons: Ember.computed.or('build.isPending', 'build.isProcessing', 'build.isFinished'),
 
+  buildHasReloadableState() {
```
fotinakis: Ah yes I didn't see the outer parens around `!(foo || bar)`, thanks for refactoring to be easier to read.

A slightly more Ember idiomatic way would be to use an [Ember.computed](https://www.emberjs.com/api/classes/Ember.computed.html) property if you have simple AND/OR logic:

```javascript
shouldPollForUpdates: Ember.computed.or('build.isPending', 'build.isProcessing'),
```

Then reference `this.get('shouldPollForUpdates')` in the conditional. This will be more efficient as well because Ember's bindings won't recompute the value of `shouldPollForUpdates` until the build properties change.

PR #137, opened 2017-06-18
Move button text up slightly

timhaines:


timhaines:
@djones @fotinakis Talked about this a while ago - moving the button text up by a pixel or two.

timhaines:
I think it has a large positive impact on some of the buttons.  Settings and Refresh at the top of the project page, and the Approve button on a build.  Other buttons, it feels like a small change without making them better or worse.

fotinakis:
Not sure why, probably the icons, but certain ones are pretty off center now:
screen shot 2017-06-18 at 2 49 55 pm

timhaines:
@fotinakis Ahh yeah, that one in your screenshot looks worse.  I'll go through each of the screens.

PR #139, opened 2017-06-22
Update docs

timhaines:
Almost tempted to push these simple docs update PRs straight to master.


fotinakis:
Definitely 👍  with pushing these straight to master, we could even automate those commits at some point.

PR #142, opened 2017-07-02
Update non-core packages for minor version bumps

timhaines:


timhaines:
@fotinakis Does the logos jumping sideways on the desktop version here (in Percy snapshots) make sense to you?

timhaines:
@fotinakis No visual diffs now, as expected.

PR #143, opened 2017-07-04
Improve build and docs page analytics.

fotinakis:
Some new events for the build page so that we can have more information about what UI features people are using the most:

- "Docs Page Viewed"
- "Snapshot Fullscreen Selected" - when the fullscreen button is clicked
- "Snapshot Fullscreen Viewed" - when a snapshot is viewed fullscreen (might happen by sending a link)
- "Build Width Selected" - width selector dropdown
- "Build Zoom Level Selected" - tracks Overview mode on/off (to be more future proof the name for this follows the internal implementation where we simple switch the number of number of columns currently viewed)

Also added to our [event taxonomy](https://docs.google.com/spreadsheets/d/16X1DnPSHCCr6J65dsisMT_fNU2n4FhdHquzRtBOLQ3I/edit#gid=0).

Tested: manually verified each event tracks data correctly and at the right moment.


app/components/build-mode-picker.js
```diff
@@ -5,14 +5,16 @@ export default Ember.Component.extend({
   selectNumColumns: null,  // Action to pass in.
   selectedNumColumns: 1,
 
-  classNames: ['ComparisonModePicker'],
+  classNames: ['BuildModePicker'],
```
fotinakis: Sneaking in a tiny refactor to fix some naming inconsistency.

timhaines:
@fotinakis `Build Mode Columns Selected` seems a confusing/unintuitive name for me at first blush.  Can you explain your thinking on that in a little more detail pls?

fotinakis:
@timhaines Agreed it's a confusing name, I'll try to think of something shorter. The way it's architected underneath, we support showing any number of columns at a time — "overview mode" just happens to be one instance of that where we pick 4 columns. @djones and I were chatting about the header redesign and think we may use 🔎+ and 🔎- icons and get rid of the word "overview" in the UI, so I wanted to make an event name that was a little bit closer to the implementation "selecting the number of columns to view" rather than the UI "overview" wording.

fotinakis:
Per discussion, changed "Build Mode Columns Selected" to "Build Zoom Level Selected".

PR #144, opened 2017-07-06
Add team members to billing plans on settings page. …

timhaines:
Also update older `parallel workers` to `concurrent renderers` in some places it was left over.


app/services/subscription-data.js
```diff
@@ -14,7 +14,8 @@ export default Ember.Service.extend({
       name: 'Free',
       monthlyPrice: 0,
       numDiffs: 500,
-      numWorkersTitle: '2 parallel workers',
+      numTeamMembersTitle: '1 team member',
```
fotinakis: Optional: looking through the screenshots, might want to include the words "Up to" to make it slightly clearer, like "Up to 5 team members"

timhaines: @fotinakis Thanks for the suggestion.  I tried it to take a look, and decided against -- weak preference of not doing it, as it's easier to parse without it, and the meaning is somewhat obvious (I don't think there's room for ambiguity).

PR #147, opened 2017-07-12
Build overview header

djones:


tests/acceptance/build-test.js
```diff
@@ -137,8 +137,8 @@ describe('Acceptance: Build', function() {
     });
     percySnapshot(this.test.fullTitle() + ' on the build page');
 
-    click('.BuildModePicker button');
-    percySnapshot(this.test.fullTitle() + ' in overview mode');
+    //click('.BuildModePicker button');
```
djones: these commented lines are removed in another PR.

tests/integration/components/build-overview-test.js
```diff
@@ -0,0 +1,81 @@
+/* jshint expr:true */
```
fotinakis: :man_dancing: 🍰

fotinakis:
One more comment: I think we should say "in 15 snapshots" rather than "from 15 snapshots"

app/styles/components/_build_overview.sass
```diff
@@ -0,0 +1,58 @@
+$statusBarWidth: 10px
```
djones: Good catch. Interestingly this approach fixes the global scope issue:

```
.BuildOverview
  $statusBarWidth: 10px
  width: $statusBarWidth // OK to use it here

body
  width: $statusBarWidth // Undefined variable error here.
```

djones: Fixed with 8df73b2b70e95a7ad4226e7311a2cacf05e2d925

PR #149, opened 2017-07-13
New fixed build header

djones:
This PR brings quite substantial updates, more so than the title suggests.

Generally speaking I've tried to approach the new toolbar by reusing as much styling as possible from the existing one. So we reduce the amount of customizations introduced for the build-specific toolbar. This approach has resulted in some changes to buttons and menu items that are not seen in the build-specific toolbar.

### Removes Overview/Detail Mode

Based on usage data less than 2% of users use that.

### Significant Mobile Improvements

Below is the biggest change. You can see a larger snapshot review area. New toolbar that takes up way less space. Example titles are hidden on mobile to fix breaking the layout of the browser size pills. It should be very usable on mobile now.

screen shot 2017-07-18 at 3 05 27 pm

### Cross Browser Dropdown Styling

Currently in Safari, the browser width dropdown looks like this:

screen shot 2017-07-18 at 3 24 03 pm

That's fixed in this PR. It looks like this now in Safari:

screen shot 2017-07-18 at 3 25 06 pm

### Settings Spacing Fix

Turns out nav item spacing wasn't even before around settings, now it is.

screen shot 2017-07-13 at 11 40 11 am

### Build Header styling Applied to Full Screen Header

Drop shadow gone.

screen shot 2017-07-18 at 3 33 38 pm

Todo:

- [x] Flesh out the approval text once approved.
- [x] Include "Percy" wording on the marketing site
- [x] Add back in trial expiry info
- [x] Make approved more obvious


timhaines:
Looking good David.  Was just taking another peek through the screenshots in Percy.  I'm guessing this one may have been an accidental side effect?  https://percy.io/percy/percy-web/builds/286647/view/8966984/1280?mode=head&snapshot=8966984

app/styles/_buttons.sass
```diff
@@ -53,17 +54,6 @@ input[type="button"],
...
-    background-color: $dark-purple
-
-    &:hover,
-      color: $white
-      background-color: $purple
-
-    &:active
-      box-shadow: inset 3px 3px 120px -40px rgba(0,0,0,0.75)
-
   &.Button--onDark.Button--primary,
```
fotinakis: Since you're killing `.Button--onDark` and moving it into `.Header--dark .Button` styles below, can you move this one too?

app/styles/_header.sass
```diff
@@ -39,42 +38,42 @@
...
+    a:not(.Button)
+      color: $white
+      
+      &:hover
+        color: $lightest-purple
+
+    .NavMenu-item
+      line-height: 4.0rem
+
+  &.Header-light
```
fotinakis: I think this code might be dead? Or `Header-light` should be `Header--light` and moved up ^^, right?

app/styles/app.sass
```diff
@@ -51,6 +54,7 @@
 @import 'components/snapshot_list'
 @import 'components/forms/billing_edit'
 @import 'components/forms/organization_user_edit'
+@import 'components/fixed_top_build_header'
```
fotinakis: Possible renaming since this will be with us for a while, we could call this something shorter like `BuildBar` or `ReviewBar`. I don't think we can call it `BuildHeader` since that's something else?

app/styles/components/_build_overview_icon.sass
```diff
@@ -0,0 +1,20 @@
+$icon-size: 22px
```
fotinakis: These will be globals, unfortunately, and `icon-size` is a pretty common name. Probably need to "namespace" them like `$build-overview-icon-size`, or actually make them globals and move them into layouts.sass or something,

app/styles/components/_build_overview_info.sass
```diff
@@ -0,0 +1,18 @@
+.BuildOverviewInfo
+  margin-top: 5px
+  text-align: right
+
+  h2
+    margin-bottom: 5px
+
+  .BuildOverviewInfo-PrNumber
```
fotinakis: Should be `-prNumber` not `-PrNumber`

app/templates/components/trial-bar.hbs
```diff
@@ -0,0 +1,18 @@
+{{#if organization.subscription.isTrialOrFree}}
+  <div class="Header-trialBar">
```
fotinakis: By convention, can't have a `Header-trialBar` inside of a `TrialBar` component, this component and it's own internals shouldn't know anything about the fact that it is sitting in the `Header`.

tests/factories/repo.js
```diff
@@ -0,0 +1,11 @@
+import FactoryGuy from 'ember-data-factory-guy';
+
+FactoryGuy.define('repo', {
+  traits: {
+    percyWeb: {
```
fotinakis: This looks more like a type/subclass of repo, not a "trait" of the repo. You could pull it out of the traits block to make it it's own type, but ideally we should just be able to call `make('repo')` and have it have some nice defaults. So maybe just make these `default` attributes too like in #147 ?

```js
default: {
  name: 'test',
  slug: 'percy/test',
  htmlUrl: 'https://github.com/percy/test',
  isPrivate: false,
  description: '',
  createdAt: () => new Date(),
  updatedAt: () => new Date(),
}
```

tests/integration/components/build-overview-info-test.js
```diff
@@ -0,0 +1,40 @@
...
+  it('renders with pull request', function() {
+    let build = make('build', 'isGithubLinked', 'hasPullRequest');
+    this.set('build', build);
+
+    this.render(hbs`{{build-overview-info build=build}}`);
+    percySnapshot(this.test);
+  });
+
+  it('renders with github branch', function() {
+    let build = make('build', 'isGithubLinked', 'hasBranch');
```
fotinakis: "branch" is a required attribute of builds, `hasBranch` is redundant and should just be one of the defaults.

I think the three states here are these?

- `make('build', 'isGithubLinked', 'hasPullRequest')`
- `make('build', 'isGithubLinked')`
- `make('build')`

app/styles/app.sass
djones: It was based on the naming of our existing "fixed-top-header". But putting that aside, something like `BuildToolbar` feels pretty good to me. Will update.

app/styles/_buttons.sass
djones: Decided to leave this one.

app/styles/app.sass
```diff
@@ -77,6 +78,7 @@
 @import 'components/saving_indicator'
 @import 'components/supported_cis_image'
 @import 'components/sync_repos_section'
+@import 'components/trialbar'
```
fotinakis: consistency nitpick: this should be `trial_bar`

djones: @fotinakis yup it's fixed with 4ecbc38a3d229f8a5bc286fbfce4ce8e4322835f probably committed while you were reviewing

PR #151, opened 2017-07-15
Upgrade packages.

timhaines:


timhaines:
@fotinakis Have looked at the organization picker in the header and on the pricing page, and at the width selector on the build header.  Let me know if the power-selects are used in another location I should check?

PR #153, opened 2017-07-19
GitHub install with redirect instead of popup

timhaines:
Requires [API update](https://github.com/percy/percy-api/pull/143) to be deployed.  Also need to coordinate the deploy with updating the URL for the production github app.

Remaining todos:

- [x] Confirm non-signed user and users without access to org with matching installation_id are handled.
- [x] Handle redirect to setup process or settings
- [x] Handle webhook not received yet?
- [x] Check to see what happens after editing repos the app has access to.
- [x] Clean up unneeded code on Adding GitHub page.
- [x] Add tests
- [x] Poll for webhook coming in. 
- [x] Add UI to show polling/waiting is taking place.


timhaines:
Need to discuss test strategy with @fotinakis.  
Perhaps consider waiting a second or two if an org isn't found with the installation_id, and reloading orgs then - to allow time for the webhook to be received and processed.   This is somewhat mitigated by choosing the users last org.

timhaines:
Want to double check logic for fetching slug from local storage and matching.

timhaines:
@fotinakis This is ready for re-review.  Polling and tests have been added.

app/router.js
```diff
@@ -29,6 +29,7 @@ Router.map(function() {
   this.route('privacy');
   this.route('security');
   this.route('admin');
+  this.route('github-app-setup');
```
fotinakis: Per discussion, would be good to have a nice namespace here we can re-use instead of a top-level route. Maybe `/install/github-app`?

```js
   this.route('install', function() {
     this.route('github-app');
   });
```

app/routes/github-app-setup.js
```diff
@@ -0,0 +1,68 @@
...
+
+
+export default Ember.Route.extend(AuthenticatedRouteMixin, {
+  currentUser: Ember.computed.alias('session.data.authenticated.user'),
+
+  queryParams: {
+      installation_id: ''
+  },
+
+  installation_id: null,
```
fotinakis: Can rename this from the camel_case query param to this route property like this:
https://guides.emberjs.com/v2.14.0/routing/query-params/#toc_map-a-controller-s-property-to-a-different-query-param-key

```js
  queryParams: {
    installationId: {as: 'installation_id', replace: true},
  },
  installationId: null,
```

Would need to update gets/sets as well.

fotinakis:
Also just noticed we say this in the UI: "Installation will open in a popup—come back here after it's installed!"

mirage/factories/github-integration.js
```diff
@@ -1,6 +1,9 @@
 import {Factory} from 'ember-cli-mirage';
 
 export default Factory.extend({
+  github_installation_id() {
```
timhaines: I have no idea how this was working previously.  Ember is magic.

app/templates/github-app-setup.hbs
```diff
@@ -0,0 +1,14 @@
+{{title "Completing GitHub Integration Setup"}}
+{{fixed-top-header}}
+<div class="container container--fixedWidth">
+  <div class="container container--forContent">
+    <div class="Card Card--extraPadded u-text-center">
+      <h3>GitHub integration processing...</h3>
+      <p>
+        {{fa-icon 'fa-spinner fa-pulse'}}
```
timhaines: image
That seems to be pretty inconsistent across all of our code.

fotinakis: Yes unfortunately that's true right now, need to get a linter for HBS.

app/router.js
```diff
@@ -29,6 +29,9 @@ Router.map(function() {
   this.route('privacy');
   this.route('security');
   this.route('admin');
+  this.route('setup', {path: '/setup'}, function() {
+    this.route('github-app-setup');
```
fotinakis: Can we do `github-app` here? so it's `/setup/github-app` instead of `/setup/github-app-setup`

timhaines:
I've done a manual run through a couple of times since renaming stuff, just to be sure.  All good.  going to merge once the tests have rerun with updates from master.

PR #154, opened 2017-07-20
Build info dropdown

djones:


tests/helpers/setup-acceptance.js
```diff
@@ -78,3 +78,11 @@ export function setupSession(createData) {
     this.loginUser = undefined;
   });
 }
+
+// By default `ember-modal-dialog` creates modals at the bottom of the body tag, which is outside
+// the testing container and so it doesn't render.
+// This moves the modal into the testing container so we can view it in our acceptance tests when
+// running Percy snapshots.
+export function moveModalIntoTestContainer() {
+  Ember.$('#ember-testing').prepend(Ember.$('.BuildInfoDropdown-modal'));
```
fotinakis: Make this not specific to BuildInfoDropdown?

app/templates/components/build-info-dropdown.hbs
```diff
@@ -0,0 +1,72 @@
...
+                <span class="BuildOverviewInfo-prNumber">
+                  #{{build.pullRequestNumber}}
+                </span>
+              </td>
+            </tr>
+        {{/if}}
+      {{/if}}
+
+      {{#if build.user_agent}}
+        <tr class='BuildInfoDropdown-fullWidthLabel'>
```
fotinakis: Style for consistency: use double quotes in HTML/HBS.

tests/helpers/setup-acceptance.js
djones: @fotinakis because I had a feeling it might screw up full screen mode. But maybe it won't... will check it out.

app/templates/components/build-info-dropdown.hbs
djones: Fixed with 04a050b275f924f4db8f575aa11a152cd7d65cdf

tests/factories/commit.js
```diff
@@ -0,0 +1,20 @@
+import FactoryGuy from 'ember-data-factory-guy';
+
+FactoryGuy.define('commit', {
+  default: {
+    committerName: 'Mike Fotinakis',
```
fotinakis: Don't love all this hard-coding of real names and commit SHAs and such. Could you take a pass at making the test data here more generic?

Ideally I think we should add an explicit new dev dependency on faker.js and do something like:

```js
committerName: faker.name.findName(),
```

We'll also need to run `faker.seed(123);` (where 123 is some static number) in order to make our tests deterministic for Percy. This could be run as part of `setupAcceptance` and `setupComponentTest`.

We already depend on `faker.js` because it comes in with mirage, we could import it from there:
http://www.ember-cli-mirage.com/docs/v0.3.x/factories/#using-fakerjs

We should especially drop the reference to `getsentry/sentry` below and just put some generic no space data in there.

tests/factories/build.js
```diff
@@ -10,9 +11,17 @@ FactoryGuy.define('build', {
     branch: 'master',
     createdAt: () => new Date(),
     updatedAt: () => new Date(),
+    commit: {commit: FactoryGuy.belongsTo('commit')},
   },
   traits: {
-    finished: {state: 'finished'},
+    longHeadCommitMessage: {commit: FactoryGuy.belongsTo('commit', 'longMessage')},
```
fotinakis: A common factory naming convention and what we do on the backend: when a trait influences a relationship, rather than just attributes, it should be named `withSomething` like `withLongHeadCommitMessage` or `withBaseBuild`. That way it's really clear when using it that you're affecting the relationship.

tests/factories/build.js
```diff
@@ -10,9 +11,17 @@ FactoryGuy.define('build', {
...
+    noSpacesMessageCommitMessage: {commit: FactoryGuy.belongsTo('commit', 'noSpacesMessage')},
+    longBranch: {branch: 'david_jones/ch699/build-header-qa/super-long-branch-name-here'},
```
fotinakis: Use faker here too, like:

```js
> faker.lorem.slug(10)
"cupiditate-odio-accusamus-ex-quo-fugiat-veritatis-quidem-eos-expedita"
```

tests/factories/build.js
```diff
@@ -10,9 +11,17 @@ FactoryGuy.define('build', {
     branch: 'master',
     createdAt: () => new Date(),
     updatedAt: () => new Date(),
+    commit: {commit: FactoryGuy.belongsTo('commit')},
```
fotinakis: Why is this `commit: {commit:`?

I think it should just be:

```js
  commit: FactoryGuy.belongsTo('commit'),
```

Unless you're defining a separate "type" of build, outside of the `default` block, which then you'd do something like:

```
build_with_bad_commit: {
  commit: FactoryGuy.belongsTo('commit', 'bad')
}
```

(but defining subtypes is something we want to avoid most of the time in preference for traits)

tests/factories/commit.js
```diff
@@ -0,0 +1,20 @@
...
+    authorName: 'David Jones',
+    committedAt: () => new Date(),
+    createdAt: () => new Date(),
+    updatedAt: () => new Date(),
+    sha: '01cb4be6f5dc5a3d19d57bbf840328fd0eb3a01f',
+    shaShort: '01cb4be',
+    message: 'Add commit factory to test suite.',
+  },
+  traits: {
+    longMessage: {message: 'Add snapshots of width selector, pending builds, and processing builds\
```
fotinakis: Let's use faker here too:

```js
> faker.lorem.sentence(30)
"Aut exercitationem qui id qui provident eius architecto et quibusdam illo praesentium incidunt atque unde blanditiis illo iste facere laboriosam tempora reiciendis doloremque delectus omnis nemo perspiciatis qui nisi aut."
```

tests/factories/commit.js
```diff
@@ -0,0 +1,20 @@
...
+                            .\n\n* Update tests to include width selector in snapshots\n\n* Capture\
+                             snapshots of pending and processing builds'},
+    noSpacesMessage: {message: 'getsentry/sentry@2d3385cf00283cf822a49261e6f38c8cf9003405'}
```
fotinakis: Could use `faker.lorem.slug` here as well.

tests/factories/build.js
```diff
@@ -33,6 +42,6 @@ FactoryGuy.define('build', {
       pullRequestNumber: '123',
       pullRequestTitle: 'New Build Header Design',
     },
-    isGithubLinked: {repo: FactoryGuy.belongsTo('repo', 'percyWeb')}
+    isGithubLinked: {repo: FactoryGuy.belongsTo('repo')}
```
fotinakis: `isGithubLinked` --> `withRepo`.

tests/integration/components/build-info-dropdown-test.js
```diff
@@ -0,0 +1,114 @@
...
+describe('Integration: BuildInfoDropdownComponent', function() {
+  setupComponentTest('build-info-dropdown', {
+    integration: true
+  });
+
+  beforeEach(function() {
+    manualSetup(this.container);
+  });
+
+  it('renders in pending state', function() {
```
fotinakis: (untested)

```js
let states = [
  ['pending'],
  ['processing'],
  ['finished', 'hasDiffs'],
  ...
]

states.forEach((state) => {
  it(`renders in state: ${state}`, function() {
    let build = make.apply(this, ['build'].concat(state));
    this.set('build', build);
    this.render(hbs`{{build-info-dropdown build=build isShowingModal=true renderInPlace=true}}`);
    percySnapshot(this.test);
  });
})
```

app/styles/components/_build_container.sass
```diff
@@ -1,3 +1,10 @@
 .BuildContainer
-  overflow-y: scroll
-  height: calc(100vh - #{$header-height})
+  &.BuildContainer--snapshotModalOpen
+    display: none
+
+  .BuildContainer-Body
```
fotinakis: `-Body` --> `-body` :)

app/templates/components/build-info-dropdown.hbs
```diff
@@ -0,0 +1,76 @@
...
+      <tr>
+        <td>Build time</td>
+        <td>{{build-duration build=build}}</td>
+      </tr>
+      <tr>
+        <td>Baseline build</td>
+        <td>{{build-comparison-info build=build}}</td>
+      </tr>
+
+      {{#if build.isGithubLinked}}
```
fotinakis: Two ways to simplify these 3 if statements into one: either use an `Ember.computed.and(` on this component to compose all the dependent keys together, or you can use the `and` operator from ember-truth-helpers (which we have installed already): https://github.com/jmurphyau/ember-truth-helpers

app/templates/components/build-info-dropdown.hbs
```diff
@@ -0,0 +1,76 @@
...
+                {{/if}}
+              </td>
+            </tr>
+          {{/if}}
+        {{/if}}
+      {{/if}}
+
+      {{#if build.user_agent}}
+        <tr class="BuildInfoDropdown-fullWidthLabel">
+          <td colspan='2'>User agent</td>
```
fotinakis: Double quotes here and below

tests/factories/build.js
```diff
@@ -10,17 +12,24 @@ FactoryGuy.define('build', {
     branch: 'master',
     createdAt: () => new Date(),
     updatedAt: () => new Date(),
+    commit: FactoryGuy.belongsTo('commit'),
   },
   traits: {
-    finished: {state: 'finished'},
+    withLongHeadCommitMessage: {commit: FactoryGuy.belongsTo('commit', 'longMessage')},
+    withNoSpacesMessageCommitMessage: {commit: FactoryGuy.belongsTo('commit', 'noSpacesMessage')},
```
fotinakis: One more potential refactor here — I don't think we should have this as part of the build factory, since it's not really re-usable (ie. we will probably only care about no-space commit messages in one specific place in one specific test, ever). It'd probably make more sense to do something like this in the specific test you care about this state:

```js
let commit = make('commit', 'noSpacesMessage');
let build = make('build', {commit});
```

tests/factories/commit.js
```diff
@@ -0,0 +1,19 @@
...
+FactoryGuy.define('commit', {
+  default: {
+    committerName: faker.name.findName(),
+    authorName: faker.name.findName(),
+    committedAt: () => new Date(),
+    createdAt: () => new Date(),
+    updatedAt: () => new Date(),
+    sha: '01cb4be6f5dc5a3d19d57bbf840328fd0eb3a01f',
+    shaShort: '01cb4be',
+    message: 'Add commit factory to test suite.',
```
fotinakis: `faker.lorem.sentence` here?

tests/helpers/setup-acceptance.js
fotinakis: Can we rename this to something like `prepareBuildInfoModalForSnapshot` so it's more specific to the build info modal?

tests/index.html
```diff
@@ -17,7 +17,7 @@
 
     <!-- Testing-only style fix that makes snapshots in Percy full-height. -->
     <style>
-      .BuildContainer {
+      .BuildContainer-Body {
```
fotinakis: `-body`

PR #155, opened 2017-07-20
Run tests in Chrome headless - tests run 5 seconds faster

timhaines:


timhaines:
Aborting due to errors reported in https://app.clubhouse.io/percy/story/667/update-percy-web-tests-to-use-chrome-headless-instead-of-phantomjs

PR #158, opened 2017-07-22
Upgrade to Ember 2.14 ember-cli ember-cli-eslint ember-simple-auth ember-source

timhaines:
Have also made some minor changes to some of the supporting files to reduce noise in diffs during ember init.


timhaines:
This introduces 2 warnings along the lines of `Warning: ignoring input sourcemap for vendor/ember/ember.debug.js`.  These are generated from ember-cli, which has an open issue for them.  Right now I think it's safe to ignore them.

PR #160, opened 2017-07-25
Build header QA

djones:
A few smaller fixes after the initial launch.


tests/factories/build.js
```diff
@@ -38,7 +39,11 @@ FactoryGuy.define('build', {
     hasPullRequest: {
       isPullRequest: 'true',
       pullRequestNumber: '123',
-      pullRequestTitle: 'New Build Header Design',
+      pullRequestTitle: () => faker.lorem.sentence(5),
+    },
+    hasPullRequestWithoutTitle: {
+      isPullRequest: 'true',
```
fotinakis: Should be a boolean, not a string?

tests/factories/build.js
```diff
@@ -38,7 +39,11 @@ FactoryGuy.define('build', {
...
+      pullRequestNumber: '456',
```
fotinakis: Should be a number.

PR #168, opened 2017-08-01
Remove mirage models

timhaines:
Mirage now auto-discovers app/models and [uses those if they exist](http://www.ember-cli-mirage.com/blog/2017/06/05/polymorphic-associations-and-auto-discovery-of-ember-data-models/).

We should remove the mirage models so they don't introduce different models/attributes from what we have in the real app.


timhaines:
3 more mirage models to be removed.  Comparison, snapshot, and organization-user.  They all cause tests to fail when removed.

timhaines:
@fotinakis I propose merging this now.  I removed all but one of the mirage models (snapshot) and fixed a few minor bugs along the way.

I'm having trouble removing the snapshot mirage model, due to what I suspect is a bug in mirage.  It doesn't seem to prevent infinite loops when resolving foreign keys when using the real model rather than the mirage model.

I doubt this is going to be an easy bug to fix in mirage, and looking at the issue and PR backlog there, I'm not confident a submitted fix would be merged quickly either.

PR #170, opened 2017-08-02
Show any errors that occurred when creating an organization

timhaines:
Have manually tested the success and failure paths.

image


app/components/forms/organization-new.js
```diff
@@ -32,4 +32,17 @@ export default BaseFormComponent.extend({
     });
   }),
   validator: OrganizationNewValidations,
+  actions: {
+    saving(promise) {
+      this._super(...arguments);
+
+      this.set('errorMessage', null);
```
fotinakis: Ideally handling base errors not related to a specific field would be shared logic for every form, since we definitely will care about this in other places. This is where that would happen: https://github.com/percy/percy-web/blob/master/app/components/forms/base.js#L68

-- I think all you'd have to do that `catch` is add the `errors` argument to it, and then do something like `changeset.set('baseError', errors.errors[0].detail)` and then reference `{{changeset.baseError}}` in the template. As far as I can tell, ember-changeset does not have a concept of 'base' errors that aren't related to a field, so that might require some thought, but it would be nice to have a common interface there.

djones: hm @fotinakis does this PR directly relate to what we discussed yesterday about the fact that we restrict the use of reserved words on `name` in addition to the `slug` just because the `slug` errors wouldn't make it to this form?

fotinakis: It sort of does, but in that case the errors are both `fields` we just happen to not show one of the fields in the UI when an error happens on it. There's sort of an ideal API/web separation we should have here, let's chat about it sometime.

PR #177, opened 2017-08-22
Parallelize percy-web test suite.

fotinakis:
This happens to speed up our tests, as well as test new ember-exam compatibility.


package.json
```diff
@@ -11,7 +11,7 @@
     "start": "ember server",
     "start-production-for-testing": "ember server --environment=production",
     "build-production": "yarn run clean && ember build --environment=production",
-    "test": "ember test",
+    "test": "PERCY_PARALLEL_TOTAL=4 ember exam --split=4 --parallel",
```
timhaines: I think you'll need to specify a PERCY_PARALLEL_NONCE here as well, otherwise PERCY_PARALLEL_TOTAL will be ignored?  Running snapshots from dev excludes any after the first finalize call.

fotinakis: We don't, it's set automatically by percy-js when running in buildkite or other supported CIs, we just can't infer the correct parallel total.

timhaines: So to capture snapshots with a local run of this now, I need to run something like `PERCY_PARALLEL_NONCE=abc yarn test` from the command line.  Would be good to get this into the README if it's the planned approach for local runs.

fotinakis: Per discussion, moving this to its own `test:parallel` entrypoint.

fotinakis:
back to you @timhaines

timhaines:
LGTM.  🍰

PR #181, opened 2017-09-03
MERGE AFTER BACKFILL: Replace Pdiff model with fields on Comparison

timhaines:
This PR replaces use of the Pdiff model with the fields that have been moved onto the Comparison model.  Some mentions of pdiff still exists in the codebase, but they relate to the diff/image itself rather than the model, so I've left those is place.

Related to this story: https://app.clubhouse.io/percy/story/895/datamodel-refactor-pdiff-fields-into-comparisons

To be deployed after [data backfill is done](https://github.com/percy/percy-api/pull/204).  Deploy at the same time as [backend update to read from comparisons](https://github.com/percy/percy-api/pull/205).


fotinakis:
Also can delete `mirage/serializers/pdiff.js`

PR #183, opened 2017-09-15
Upgrade packages, and add lint command.

timhaines:
You can now run `yarn lint` and it finishes in < 3 seconds.


package.json
```diff
@@ -6,16 +6,17 @@
     "test": "tests"
   },
   "scripts": {
-    "clean": "rm -rf dist/* tmp/*",
     "build": "ember build",
+    "build-production": "yarn run clean && ember build --environment=production",
+    "clean": "rm -rf dist/* tmp/*",
+    "lint": "./node_modules/eslint/bin/eslint.js .",
```
timhaines: Adding lint is the only adjustment to scripts here.  The other lines are all the same, just alphabetically ordered.

PR #214, opened 2017-10-21
Auth0 Integration - Github Only

djones:
Integrate auth0 login into percy-web.
This implementation is the frontend piece of [Option 3.5 from this document](https://docs.google.com/document/d/1mSlGlaNF8m6x-98KHRab1YZMTEYhe8jjnHkpIlK9Pnc/edit#heading=h.2uhsjrsbekov). 

The biggest change is that the custom authenticator has been removed in favor of auth0 authenticator. The rest of the updates in this branch follow from that. 

Logic flow:
- User visits any percy page
- Application route `beforeModel` fires. This sets off some inherited `beforeModel` actions in `ember-simple-auth` and `ember-simple-auth-auth0` that initiates the authentication process. If the user is already logged in, it requests the current user from the backend. 
If the user is not logged in it:
    - redirects the user to the login page if they are visiting an authentication-required route OR
    - does nothing if the user is on a publicly available route
- Before the login modal displays, the app sends a request to `api/auth/session` to get a state token. The state token is inserted into the `lockOptions` hash which is sent to auth0, and is then sent back to the backend for a check to make sure they match. It is also kept in the ember-simple-auth localstorage session so the state parameter can be consistent for all of that user's session.
-  On the login page or with the login modal, the user types in their info and sends it off to auth0. Once the chat is complete, the backend redirects to the `callback` route with the correct tokens. The callback route is an _unauthenticated_ route which means that you can _only_ reach it when there is no authenticated user. 
This is ok because:
- When the `ember-simple-auth`/`ember-simple-auth-auth0` libraries are finished processing the tokens in the url sent back from backend to the callback route, it logs the user in and fires the `sessionAuthenticated` method. The default behavior of this method executes a set of redirect logic that does not work for us, so I've implemented custom redirect logic that starts in this method.
- User gets redirected 

TODO:
- [x] in progress [check/handle login failures](https://app.clubhouse.io/percy/story/1092/handle-login-failure-cases-gracefully) (in progress here #243 )
- [x] [check/handle extended github integration](https://app.clubhouse.io/percy/story/1089/fix-test-extended-github-integration) (this works)
- [x] in progress [replace intercom and analytics setup removed in custom authenticator](https://app.clubhouse.io/percy/story/1117/replace-intercom-and-analytics-setup-that-was-removed-by-removing-custom-authenticator) (in progress here #243 )
- [x] [test bot user compatibility](https://app.clubhouse.io/percy/story/1090/test-bot-user-login)
- [ ] Tests?
- [ ] If login cancelled on /login page, don't leave the user on a blank page


app/services/analytics.js
```diff
@@ -94,7 +94,7 @@ export default Service.extend({
     if (organization) {
       this.organizationInstance.setUserId(organization.get('id'));
       let organizationEventProperties = Object.assign(
-        {user_id: this.get('currentUser.id')},
+        {user_id: this.get('currentUser.user_metadata.id')},
```
cadeParade: check this...Looks like before currentUser was a POJO from ember simple auth, but now currentUser would be a user model

cadeParade: I've checked this out, and indeed, this property is now undefined.
@djones, looks like you updated this line last, do you know which id this is supposed to be? Presumably not their percy user id...
maybe this id? `github|3179218`

cadeParade: ok, looking back in history, looks like it was originally `currentUser.get('id')` so I'll replace it with that.

app/components/organizations/github-integrator.js
```diff
@@ -13,8 +13,6 @@ export default Component.extend({
   classes: null,
 
   store: service(),
-  session: service(),
```
cadeParade: These were not being used in this component AFAICT

app/styles/_auth0.sass
```diff
@@ -0,0 +1,27 @@
+.auth0-lock-header-bg, .auth0-lock-submit
```
djones: all subject to change here via a sub-feature PR.

config/environment.js
```diff
@@ -42,6 +42,14 @@ module.exports = function(environment) {
         integration: 'https://github.com/apps/percy/installations/new',
       },
     },
+
+    'ember-simple-auth': {
+      auth0: {
+        clientID: 'xFX5PgBn_v8iGt97dDsk36h8GjQ1WqC2',
```
djones: realising now that we'll need a different client ID per environment.

app/authenticators/custom.js
```diff
@@ -1,87 +0,0 @@
...
-
-export default BaseAuthenticator.extend({
-  store: service(),
-  analytics: service(),
-  restore() {
-    let store = this.get('store');
-    return new EmberPromise((resolve, reject) => {
-      store.queryRecord('user', {}).then(userRecord => {
-        if (window.Intercom) {
-          window.Intercom('update', {
```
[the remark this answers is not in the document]
cadeParade: Yes, covered in TODO's in the description... wanted to keep the scope of this PR from expanding further... but lmk if you think it should go in this PR

app/lib/lock-settings.js
```diff
@@ -0,0 +1,18 @@
+export default {
+  auth: {
+    redirect: true,
+    redirectUrl: 'http://dev.percy.local:4200/api/auth/auth0/callback',
```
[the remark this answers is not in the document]
cadeParade: Oh yeah, definitely, good catch.

config/environment.js
cadeParade: why is this?

cadeParade:
@djones ok, your comments are address and the tests are passing, so this PR is ready for final review. I'll do the followups listed above (which should be similar to those in clubhouse) as followup PRs to be merged into this one once its approved.

app/components/login-button.js
```diff
@@ -2,19 +2,18 @@ import Component from '@ember/component';
 
 export default Component.extend({
   classes: null,
-  text: 'Sign in with GitHub',
+  text: 'Login or Sign Up',
```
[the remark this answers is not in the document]
cadeParade: Actually, we should probably keep this button as it was ("sign in with github")  since right now we are only shipping github login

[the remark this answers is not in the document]
djones: Yup, good call.

app/authenticators/custom.js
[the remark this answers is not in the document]
cadeParade: This is now done in services/session

app/components/user-avatar.js
```diff
@@ -5,7 +5,7 @@ export default Component.extend({
   user: null,
   classes: null,
   linked: false,
-  href: readOnly('user.githubUrl'),
+  href: readOnly('user.avatarUrl'),
```
fotinakis: The `href` here is the link for the avatar when `linked=true` and when the avatar is click, so it should be a relevant destination not the avatar URL itself. I think we can leave this as is for now, and handle it when decoupling frontend from GitHub.

We use linked=true in a two places:

- `app/templates/components/organizations/github-bot-integrator.hbs`
- `app/templates/components/organizations/user-card.hbs`

fotinakis: Sorry when I said "leave as is" I mean revert it to using `user.githubUrl`

app/lib/localstorage.js
```diff
@@ -0,0 +1,30 @@
+export default {
```
fotinakis: Nice centralization of this logic.

app/mixins/ensure-stateful-login.js
```diff
@@ -0,0 +1,21 @@
+import Mixin from '@ember/object/mixin';
+import lockOptions from 'percy-web/lib/lock-settings';
+import $ from 'jquery';
+
```
fotinakis: I think this needs a brief comment to explain what this mixin does and why.

app/router.js
```diff
@@ -18,12 +20,13 @@ const Router = EmberRouter.extend({
...
 
 Router.map(function() {
-  this.route('login');
   this.route('join', {path: '/join/:invite_code'});
   this.route('auth-failure', {path: 'auth/failure'});
   this.route('docs', {path: '/docs'}, function() {
     this.route('page', {path: '*path'});
   });
+  this.route('login');
+  this.route(AUTH_CALLBACK_ROUTE);
```
fotinakis: I think we should make this not a top-level route, particular because top-level routes that we use need to be reserved in the backend as well so that they cannot be claimed as an org name.

Can we rename this to `auth-callback` and make it path: `auth/callback` like the auth-failure route above?

app/routes/application.js
```diff
@@ -1,65 +1,119 @@
...
+  beforeModel(transition) {
+    this._super(...arguments);
+    if (!this.get('session.isAuthenticated')) {
+      this._storeTargetTransition(transition);
     }
+    return this._loadCurrentUser();
+  },
+
+  sessionAuthenticated() {
+    // This method is called after the session is authenticated by ember-simple-auth.
```
fotinakis: Great comment thank you!

app/routes/application.js
```diff
@@ -1,65 +1,122 @@
...
+    redirectToDefaultOrganization() {
+      this._redirectToDefaultOrganization();
+    },
+
     navigateToProject(project) {
       let organizationSlug = project.get('organization.slug');
       let projectSlug = project.get('slug');
       this.transitionTo('organization.project.index', organizationSlug, projectSlug);
     },
+
```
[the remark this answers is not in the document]
fotinakis: 😶

app/services/session.js
```diff
@@ -0,0 +1,83 @@
...
+    this._clearSentry();
+    this._clearAnalytics();
+
+    if (window.localStorage) {
+      window.localStorage.clear();
+    }
+  },
+
+  _setupSentry(user) {
+    Raven.setUserContext({id: user.get('id')});
```
fotinakis: Should we `if (window.Raven)` here?

ember-cli-build.js
```diff
@@ -22,6 +22,7 @@ module.exports = function(defaults) {
     fingerprint: {
       extensions: ['js', 'css', 'png', 'jpg', 'gif', 'map', 'svg'],
       prepend: '/static/',
+      exclude: ['images/percy.svg'],
```
fotinakis: I don't think we should exclude any assets here — we set aggressive Cache-Control headers on all static assets, so if we did this then when we end up changing `images/percy.svg` it will not get updated across clients potentially for a long time. Keeping fingerprinting ensures that any assets we ever ship and rely on can be cached aggressively because we can assume they won't change.

tests/unit/services/session-test.js
```diff
@@ -0,0 +1,144 @@
+/* eslint-disable no-unused-expressions */
+import {expect} from 'chai';
+import {it, describe, beforeEach} from 'mocha';
+import {setupTest} from 'ember-mocha';
+import {make, manualSetup} from 'ember-data-factory-guy';
+import {resolve, reject} from 'rsvp';
+import sinon from 'sinon';
+
+describe('SessionService', function() {
```
fotinakis: 🎉 💯 🥇

app/router.js
cadeParade: This is done, and I've opened a corresponding backend PR here: https://github.com/percy/percy-api/pull/265

app/mixins/ensure-stateful-login.js
cadeParade: 👍  done

ember-cli-build.js
cadeParade: The reason we decided to do this is because we must give a full path to lock0 for the logo url: https://github.com/percy/percy-web/pull/214/files#diff-9abe2a49cb3e9a7fd4594ca3530a668bR14

If we didn't exclude this from fingerprinting, we would have to update this url each time the fingerprint changed.
But maybe you know a ✨  magic secret ✨ ?

app/services/session.js
cadeParade: yes, great call, this is done

PR #230, opened 2017-11-03
Add prettier-eslint CLI to improve editor linting

faun:
Add CLI to parse `eslint` configuration and apply `prettier`-related configuration to make it easier to auto-format in a text editor.

Minor:
- Add `eslint-check` npm script to check `eslint` configuration issues


package-lock.json
```diff
@@ -0,0 +1,1193 @@
+{
```
cadeParade: I think we don't need this `package-lock` file, since we should be using `yarn install` which generates `yarn.lock` rather than `npm` to install stuff

faun: Oops. Removed.

PR #233, opened 2017-11-06
Snapshot titlebar adjustments

PixelJanitor:
- [x] Title truncates with ellipsis in favor of header controls getting layout priority
- [x] Title is returned as a link when the `snapshotUrl is valid` and a div when not
- [x] General CSS refactoring

image


fotinakis:
Yayayay!

PR #242, opened 2017-11-09
Auth0 UI Updates

PixelJanitor:
image


PixelJanitor:
I think these updates are about the safest route we can go right now. If there is a piece of UI you particularly aren't fond of, I will wrangle what I can with it. Auth0 really pulled on the reigns with v10 when it comes to styling, but luckily we have Percy to keep our visual diffs in check ✅

PixelJanitor:
Ah you're totally right. They didn't show with a fully opaque background, and I added the .9 opacity back last second. Thanks

djones:
I did a little testing on mobile and found this on Safari

img_1ca4d6c0707e-1

Not sure if this is broken in Lock.js or our overrides but we should fix that one :-)

PixelJanitor:
Yea I noticed that too and i've been working on that currently. It's not part of our styles. I've commented everything out and it still has that layout. I'll let you know what I figure out.

PixelJanitor:
@cadeParade if you don't mind giving this a final once-over today we can get this merged in and you guys can take it to the finish line. The styling issues discussed above are caused by the passwordless issue you're working on, and will be resolved once you get that ripped out!

cadeParade:
So the styling issues in safari mobile are also caused by passwordless lock?

PixelJanitor:
Yep - here is safari mobile with passwordless styles removed (how it should look):
image

cadeParade:
sweeeeeeeeeeeet 👍

cadeParade:
Is the safari mobile problem only with the modal with username and password fields, or does it also look bad with just the github button?
I can't figure out how to reproduce it.

PixelJanitor:
Both. There are styles getting injected by passwordless that
affects it. I’ll hop on real quick and show you.

On November 13, 2017 at 9:08 PM, Lindsay
(notifications@github.com) wrote:

Is the safari mobile problem only with the modal with username
and password fields, or does it also look bad with just the
github button?
I can't figure out how to reproduce it.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub 
( https://github.com/percy/percy-web/pull/242#issuecomment-344122740 )
, or mute the thread

PixelJanitor:
Lindsay and I determined that the passwordless style tag issue is not affecting the GitHub-only login layout and shouldn't be a blocker to move forward with this week's deploy of that version of Auth 👍

PR #245, opened 2017-11-11
Avatar Overlap Fix

PixelJanitor:
A sass nesting issue caused:
screen shot 2017-11-11 at 4 27 28 pm

And this PR sets the layout to:
screen shot 2017-11-11 at 4 28 56 pm


djones:
Good catch 👍

PR #248, opened 2017-11-13
Update Normalize to 7.0.0

PixelJanitor:


app/styles/app.scss
```diff
@@ -1,88 +1,98 @@
...
 
 // Variables.
-@import "variables/colors";
-@import "variables/fonts";
-@import "variables/mixins";
-@import "variables/animations";
-@import "variables/shadows";
-@import "variables/code";
-@import "variables/layouts";
+@import
```
djones: Neat, I didn't know you could group imports like this.

djones: Neat, I didn't know you could group imports like this.

PixelJanitor: Yep! I think it's cleaner.

PR #253, opened 2017-11-16
Auth0 -- add state param

cadeParade:
This PR uses the recent backend updates related to an authentication state param. Each time the lock modal is triggered (this can be via the "Login via github button" on the home page, the "Sign in" button in the menu bar, or via the login route), the app makes a request to `api/auth/session` to get a unique state token. 
The state token is immediately inserted into the lock options hash which is then sent to auth0. The auth0 addon then stores the state token in the auth session in localstorage so it sticks around while navigating the site and reloading the page. This also takes care of removing the token whenever the session is invalidated, the user logs out, or there is some sort of error.


cadeParade:
tests are passing now -- it was failing from an uncaught promise rejection for the ajax call to api/auth/session
I've stubbed that out now but we should also handle that better.

djones:
👍

PR #254, opened 2017-11-16
Snapshot Titles

PixelJanitor:


app/styles/components/_comparison_mode_switcher.scss
```diff
@@ -1,34 +1,3 @@
...
-    &.active {
-      color: $default-link-hover;
-      background: $white;
-    }
-  }
-}
-
-.ComparisonModeSwitcher--hidden {
-  visibility: hidden;
+.is-hidden {
```
fotinakis: Big point to mention here: these styles are not scoped in any way to a specific component like they might be in react, they're just scooped up into a global list. So this will define a top-level `.is-hidden` class. Everything in side of a file in `components/` should be scoped inside of a BEM class by the same name, like `_comparisons_mode_switcher.scss` --> `.ComparisonModeSwitcher { }`

app/components/comparison-mode-switcher.js
```diff
@@ -2,12 +2,12 @@ import {alias, empty} from '@ember/object/computed';
 import Component from '@ember/component';
 
 export default Component.extend({
-  classNames: ['ComparisonModeSwitcher'],
-  classNameBindings: ['isHidden:ComparisonModeSwitcher--hidden'],
+  classNameBindings: ['isHidden:is-hidden'],
```
cadeParade: this is a completely reasonable and explicit way to do this, but if you want to get real fancy, it turns out that ember should auto-dasherize boolean properties that you put in classNameBindings, so I think if you just had `classNameBindings: ['isHidden']` it would render `is-hidden` if `isHidden` is true, and nothing if it is false.
[Documentation here](https://guides.emberjs.com/v2.13.0/components/customizing-a-components-element/) (grep for `isUrgent` to find the specific place)

That being said, I'm not sure that magic is better because it's less explicit, but I'll leave it up to you whether you want to change it and also you probably learned something about Ember :) ✨

app/components/comparison-mode-switcher.js
```diff
@@ -2,12 +2,12 @@ import {alias, empty} from '@ember/object/computed';
...
   wasAdded: alias('comparison.wasAdded'),
   isHidden: empty('comparison'),
   comparison: null,
   comparisonMode: null,
+
```
cadeParade: Dat


white


space


tho

app/styles/components/_comparison_mode_switcher.scss
PixelJanitor: Yea we should have an global `.is-hidden` class that can be shared throughout so that we're not duplicating styles - and that's what I created in `_utilities.scss` that file will become more files but is a home for these utilities for now. This file shouldn't even be here anymore and it slipped through the cracks. Thanks for catching!

app/components/comparison-switcher-item.js
```diff
@@ -5,21 +5,23 @@ export default Component.extend({
...
-  classNames: ['ComparisonSwitcherItem'],
   updateSelectedWidth() {},
-  classNameBindings: [
-    'isSelected:ComparisonSwitcherItem--selected',
-    'matchingComparison::ComparisonSwitcherItem--noComparison',
-    'matchingComparison.isDifferent:ComparisonSwitcherItem--hasDiffs',
-  ],
-  matchingComparison: computed('comparsions', 'width', function() {
+
+  matchingComparison: computed('comparisons', 'width', function() {
```
cadeParade: thanks for fixing this 👍

app/styles/components/_comparison_mode_switcher.scss
cadeParade: do we need two `is-hidden` classes?

app/styles/components/_snapshot_viewer.scss
```diff
@@ -53,10 +55,41 @@
...
+  grid-template-columns: 1fr auto 1fr;
 }
 
 .SnapshotViewer-header-controls {
   display: flex;
 }
+
+.maximize-snapshot,
+.minimize-snapshot {
+  path {
```
cadeParade: what is this magic?

oh...svg things...interesting

app/templates/components/snapshot-viewer-full.hbs
```diff
@@ -1,28 +1,47 @@
-<div class="Viewer-header SnapshotViewerFull-header row">
-  <div class="five columns" title="{{snapshot.name}}">
-    <strong class="Viewer-header-title">{{snapshot.name}}</strong>
+<div class="Viewer-header SnapshotViewer-header SnapshotViewer-header-full">
+  <div class="flex align-center truncate">
+    <div class="svg-container">
+      <svg width="23" height="26" viewBox="0 0 23 26" class="mr-2">
+        <g fill="none" fill-rule="evenodd">
+          <path fill="#1D8DDB" fill-opacity=".104" d="M13.8085593,24.4714932 L3.15372306,24.4714932 C2.41808653,24.4714932 1.82186854,23.882373 1.82186854,23.1554941 L1.82186854,7.36350472 C1.82186854,6.63662577 2.41808653,6.0475056 3.15372306,6.0475056 L13.8085593,6.0475056 C14.1612876,6.0475056 14.500511,6.1863014 14.7502337,6.4330407 C14.9999564,6.67978001 15.1404138,7.01494919 15.1404138,7.36350472 L15.1404138,23.1554941 C15.1404138,23.5040233 14.9999458,23.8392083 14.7502337,24.0859581 C14.5005216,24.3327079 14.1613143,24.4714932 13.8085593,24.4714932 Z" transform="rotate(-5 8.481 15.26)"/>
```
cadeParade: What does this make?  It's a bit terrible to read, so maybe we can stick this thing in a component, so we don't have to look at svg paths unless we _really_ want to?

cadeParade: Also, if it is an icon as i think it might be, then if it was in a component we could use the icon in other places

tests/acceptance/build-test.js
```diff
@@ -299,7 +299,7 @@ describe('Acceptance: Build', function() {
 
     visit(`/${this.project.fullSlug}/builds/${this.build.id}/view/${snapshot.id}/320?mode=diff`);
     andThen(() => {
-      expect(find('.ComparisonModeSwitcher').css('visibility')).to.equal('hidden');
+      expect(find('[data-test-comparison-mode-switcher]').css('visibility')).to.equal('hidden');
```
cadeParade: jooc why use `find(selector)` rather than `$(selector)`?

PixelJanitor:
OMG so much yes to svg image components

app/components/comparison-mode-switcher.js
PixelJanitor: Oh fancy! I'm like 50/50 on this. I like the ember trick there, but it's way more clear to a designer who's writing css classes what's happening here ¯\_(ツ)_/¯

app/templates/components/snapshot-viewer-full.hbs
PixelJanitor: Yup - totally an icon

tests/acceptance/build-test.js
PixelJanitor: Because the rest of the test suite does this ¯\_(ツ)_/¯ 

The test-selector documentation also doesn't uses the `find()` way

fotinakis: Yeah `find` selector is "better" overall, Ember is actually moving away from having jquery-explicit `$` dependencies mostly everywhere.

app/templates/components/snapshot-viewer-full.hbs
fotinakis: Or it could be just the `.svg` file stored in `public/images` and then sourced into this place. Our ember build pipeline fingerprints assets including SVGs so they can be forever cached at the edge in our CDN and in users browsers, so it would continue to be very fast.

PixelJanitor: Svg’s in img tags are non-interactive unfortunately, but for the ones that just need to be statically presentational, we can totally put them in an img tag. The comparison icon I created is static at the moment, but I want to add a subtle hover separating transition for when you hover the header. So I went ahead and embedded it for now so that it’s interactive. Believe me - I wish we could access an svg’s code through its img tag!

PixelJanitor:
FYI it doesn't appear that any of the percy tests include snapshots where there is a mode switcher in the middle of the snapshot header full screen view.

app/templates/components/comparison-mode-switcher.hbs
```diff
@@ -1,18 +1,13 @@
-<nav class="ComparisonModeSwitcher">
+<div class="btn-group" data-test-comparison-mode-switcher role="group" aria-label="Snapshot Comparison Switcher">
```
fotinakis: @PixelJanitor A weird (and slightly older) Ember thing, but there is actually a wrapping outer element that is outside of this div as well — ie. everything inside this file is actually wrapped in ANOTHER div, which is the actual ComparisonModeSwitcher element (the fact that the previous file had the `class="ComparisonModeSwitcher" on it is kind of incorrect). Ideally you would apply `data-test-comparison-mode-switcher` in the js file itself and not here in the template, so it just gets added to the most top-level div of this component. Happy to chat about this more.

app/templates/components/comparison-mode-switcher.hbs
```diff
@@ -1,18 +1,13 @@
...
   {{#if wasAdded}}
-    <span
-      class="ComparisonModeSwitcher-switch active"
-      {{action updateComparisonMode 'head'}}>New Snapshot</span>
+    <button type="button" class="btn is-active"
```
fotinakis: @PixelJanitor what do you think about `active` (and things like `disabled`,`enabled`, etc.) rather than `is-active`?

app/templates/components/comparison-mode-switcher.hbs
cadeParade: Putting `data-test` attrs in the component js files is terrible because you have to add two lines:
```
attributeBindings: ['data-test-foo'],
"data-test-foo": true
```

We used [this tool](https://github.com/AltSchool/ember-get-component) at altschool to get component instances in tests without having to put `data-test` attrs in component js. It was kind of the best, although I'm sure it's not actively maintained any more.

app/templates/components/comparison-mode-switcher.hbs
PixelJanitor: I've used both before, but having an `.is-` prefix shows it's a stylistic class added by javascript. This lets nextguy come in and see what styling is being added via logic and what styling is always on the component. Same with `.has-` .

app/templates/components/comparison-mode-switcher.hbs
PixelJanitor: Like I told Lindsay the other day - I want to so bad make sure that markup stays in the .hbs files as much as humanly possible so that we can keep the engineering logic in the js files. I would MUCH rather be able to keep static attributes like this in the hbs files when we add them from now on.

app/components/snapshot-viewer-full.js
```diff
@@ -8,19 +8,24 @@ export default Component.extend({
...
   }),
+
   buildWidths: alias('build.comparisonWidths'),
   selectedComparison: computed('snapshot', 'snapshotSelectedWidth', function() {
     let comparisons = this.get('snapshot.comparisons') || [];
     let width = parseInt(this.get('snapshotSelectedWidth'), 10);
     return comparisons.findBy('width', width);
   }),
+
+  hasComparisonAtSelectedWidth: notEmpty('selectedComparison'),
```
cadeParade: This is sort of a note for the future, but we have several places where we need to decide if there is a snapshot comparison for a certain width. I wonder if it makes sense to put on the snapshot model something like 
```
comparisonForWidth(width) {
  return this.get('comparisons').findBy('width', width);
},
hasComparisonforWidth(width) {
  return Ember.isPresent(this.comparisonForWidth(width));
}
```

This seems useful in many places

djones:
Merging... a follow up PR to fix the icon will be created at a later stage.

PR #258, opened 2017-11-19
Fix percy logo + callback redirect

cadeParade:
Do not exclude percy logo from asset fingerprinting

Fix bug for callback route redirect.


app/router.js
```diff
@@ -2,7 +2,7 @@ import {on} from '@ember/object/evented';
 import EmberRouter from '@ember/routing/router';
 import config from './config/environment';
 
-export const AUTH_CALLBACK_ROUTE = 'auth/callback';
+export const AUTH_CALLBACK_ROUTE = 'auth-callback';
```
djones: good subtle catch there!

djones: is this why we were seeing .../organizations/auth as the redirect URL?

PR #262, opened 2017-11-21
Redirect to `/` after closing login modal on `/login`

cadeParade:
Previously when on `/login` and closing the login modal, it would leave you on a blank login page.
Now, closing the modal will redirect you to `/`.

I couldn't figure out how to grab the actual instance of the lock js modal rendered by the auth0 addon to grab onto to put the `hide` event listener on. So instead I've taken the `showLock` code  from https://github.com/auth0-community/ember-simple-auth-auth0/blob/develop/addon/services/auth0.js#L93 and implemented the `show` manually so I could add the event listener. 

Additionally, before when you clicked "login" button on the home page twice, it would show two modals. Now it shows only once.


tests/acceptance/login-test.js
```diff
@@ -23,8 +23,6 @@ describe('Acceptance: Login', function() {
       this.server.create('user', {_currentLoginInTest: true});
     });
 
-    click('a.LoginButton');
```
cadeParade: After the updates to `ensure-stateful-login` mixin (calling `showLock` manually) , clicking this button was calling to the auth0 server to try and get lock information. This caused other tests down the line to fail with unrelated errors: see [here](https://buildkite.com/percy/test-web/builds/1509#ae899372-1093-4005-a444-105c954de1d3/22-111) for an example. 

So I've removed this line since it doesn't really add anything to the test since it authenticates on line 27.

config/environment.js
```diff
@@ -91,7 +91,7 @@ module.exports = function(environment) {
     ENV.percy = {
       defaultWidths: [375, 1280],
     };
-    ENV['ember-simple-auth']['auth0'].clientID = 'foo';
+    ENV['ember-simple-auth']['auth0'].clientID = 'drcZPIIVI3fP1giWuojIp6NQ6K3owYuC';
```
cadeParade: Undo this

PR #267, opened 2017-11-21
Remove bot user configuration.

fotinakis:


djones:
@fotinakis I think

* [This line](https://github.com/percy/percy-web/blob/master/app/styles/app.scss#L79)
* [This file](https://github.com/percy/percy-web/blob/master/app/styles/components/organizations/_github_bot_integrator.scss)

Can be removed too.

fotinakis:
@djones done

PR #269, opened 2017-11-24
Update README.md

twokul:
Looks like there's no need to run `bower install` since it's a part of `postinstall` step in `package.json`


timhaines:
@twokul Thanks Alex!  We added the `postinstall` script just a couple of days ago. It's nice to get this tidied up too.

PR #273, opened 2017-11-26
Switcher Item States

PixelJanitor:
This fixes the issue of the comparison switcher buttons having red text when there isn't a diff for that width.


PixelJanitor:
I can't test that this works on the current master branch since I can't run the most recent version of master with the Auth0 stuff locally yet.

PR #277, opened 2017-11-27
Remove passwordless lock css element

cadeParade:
Passwordless auth0 lock adds a style element to the page on [addon initializaiton](https://github.com/auth0/lock-passwordless/blob/master/src/index.js#L24)

Running `ember-simple-auth-auth0` without passwordless breaks the app, and writing a PR to separate their concerns is not trivial.

This issue is documented on the `ember-simple-auth-auth0` repo [here](https://github.com/auth0-community/ember-simple-auth-auth0/issues/76) where another developer has come to the same conclusion.

The auth0 team is working on combining passwordless into the rest of the auth modal, but for now I propose the way forward to be just deleting this rogue style div.

This solution works because the standard `lock.js` addon injects its styles into an [id'ed element](https://github.com/auth0/lock/blob/master/src/index.js#L7). So the jquery selector removes style elements with a certain style in it (specific to auth0) but rejects the one with the id that we want to keep.


PixelJanitor:
I'll let one of the reviewers give the final thumbs up, but good work on this. I know it was quite the chase to get where you are with it :)

PR #278, opened 2017-11-27
Dev + Production API Support Fixes

djones:
A quick hack that will trick the frontend into logging you in if you're running dev with the production API.


app/lib/dev-auth.js
```diff
@@ -0,0 +1,5 @@
+import config from '../config/environment';
+
+export default function() {
+  return config.APP.percyWebApiHost && config.APP.percyWebAuthToken;
```
fotinakis: Or just `config.APP.environment == 'development'` I think would work?

app/routes/application.js
```diff
@@ -5,6 +5,7 @@ import ApplicationRouteMixin from 'ember-simple-auth-auth0/mixins/application-ro
 import localStorageProxy from 'percy-web/lib/localstorage';
 import {AUTH_CALLBACK_ROUTE} from 'percy-web/router';
 import EnsureStatefulLogin from 'percy-web/mixins/ensure-stateful-login';
+import isDevWithProductionAPI from 'percy-web/lib/dev-auth';
```
fotinakis: Naming here is maybe a bit too explicit — maybe just call this `isDevelopment()`? Perhaps even drop the module entirely and just use `config.APP.environment == 'development'` check directly + a comment explaining.

config/environment.js
```diff
@@ -68,6 +68,9 @@ module.exports = function(environment) {
       integration: 'https://github.com/apps/percy-dev/installations/new',
     };
 
+    ENV.APP.percyWebApiHost = process.env.PERCY_WEB_API_HOST;
```
fotinakis: Could drop these if you use a `config.APP.environment` check.

app/lib/dev-auth.js
cadeParade: I don't think we want this to happen _always_ in development, just when you pass in an auth token and api host specifically

fotinakis: Ahh yes, totally. Whoops!

PR #281, opened 2017-11-28
Add flash message support for login errors

cadeParade:
If there is some problem with the login system on our end (between the time github pings our backend and the time the user object shows up in the frontend), this displays a flash message that persists across 1 full page reload, which happens after a session is invalidated.

flash-messages


cadeParade:
@PixelJanitor plz to help with nice flash messages 🙏

app/services/session.js
```diff
@@ -1,12 +1,13 @@
...
-import {resolve} from 'rsvp';
+import {resolve, reject} from 'rsvp';
 import {Promise as EmberPromise} from 'rsvp';
+import localStorageProxy from 'percy-web/lib/localstorage';
 
 export default SessionService.extend({
   store: service(),
   analytics: service(),
-
+  flashMessages: service(),
```
cadeParade: get rid of this

app/services/session.js
```diff
@@ -52,10 +54,8 @@ export default SessionService.extend({
   _clearThirdPartyUserContext() {
     this._clearSentry();
     this._clearAnalytics();
-
-    if (window.localStorage) {
-      window.localStorage.clear();
```
cadeParade: This was clearing all of localstorage which doesn't work if we need to persist flash message data in localstorage. So I've changed this to remove specific keys with strings related to their services.

app/styles/_flash_messages.scss
```diff
@@ -0,0 +1,34 @@
+/* stylelint-disable selector-no-qualifying-type */
+
+.flash-message {
```
djones: Let's run this by @PixelJanitor. We have some existing "alert" styles here https://github.com/percy/percy-web/blob/master/app/styles/_alerts.scss that would be a good starting point but that might not be visually acceptable in a flash message context.

app/lib/localstorage.js
```diff
@@ -27,4 +27,29 @@ export default {
       // Safari throws errors while accessing localStorage in private mode.
     }
   },
+
+  keys() {
+    try {
+      // const localStorageLength = localStorage.length;
```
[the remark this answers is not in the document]
cadeParade: yes i could, thanks for catching

app/routes/application.js
```diff
@@ -38,7 +39,20 @@ export default Route.extend(ApplicationRouteMixin, EnsureStatefulLogin, {
   },
 
   _loadCurrentUser() {
-    return this.get('session').loadCurrentUser();
+    return this.get('session')
+      .loadCurrentUser()
+      .catch(() => {
+        this.get('flashMessages').createPersistentFlashMessage(
+          {
```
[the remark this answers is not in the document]
cadeParade: Any thoughts about wording?

cadeParade:
@djones While writing up steps for Derek to reproduce the error, I realized it should show a flash message if a user runs into the same problem, so I've added a persistent flash message to the auth-failure route.

cadeParade:
@PixelJanitor to get a flash message you can:
1. put `this.get('flashMessages').success('message', {sticky: true});` inside the `beforeModel` hook in `app/routes/application`. This will always show a message whenever the page loads. You can change the `success` to `danger`, `info`, or `warning` and it will change the class. You can also get more than one to display on the page at once, if you edit the absolute position css.

2. To reproduce the actual error state, start your server with export `PERCY_WEB_API_HOST=https://percy.io` and _not_ the token, and try to login. It should display a flash message.

The flash messages are from `https://github.com/poteto/ember-cli-flash`, so feel free to look at other settings to make it nicer. You can attach animations to it! I tried briefly, but had no luck.

app/styles/_flash_messages.scss
PixelJanitor: are those alert styles used in a current alert somewhere?

djones: Yep, there's a couple of samples here https://percy.io/docs/clients/javascript/react

cadeParade:
Omg these are very better 😍

PixelJanitor:
This is good to go on my end if you want to do a final review 👍

app/routes/auth-failure.js
```diff
@@ -1,6 +1,19 @@
...
+  flashMessages: service(),
   beforeModel() {
+    this.get('flashMessages').createPersistentFlashMessage(
+      {
+        type: 'danger',
+        message:
+          'There was a problem with logging in. \
+        Please try again or contact us if the issue does not resolve.',
+      },
+      {persistentReloads: 0},
```
cadeParade: me: double check this

cadeParade:
@PixelJanitor The flash messages sort of disappear on mobile screen sizes -- they disappear very gracefully, but wondering if they should still display on mobile?

PixelJanitor:
My mobile support commit didn't make it for some reason, but I added it back in and added a small refactor to a sass loop. Let me know if you need anything else on this!

cadeParade:
Thanks @PixelJanitor ! I've confirmed the fix and merging now!

PR #282, opened 2017-11-29
Visual tests for power-selects

cadeParade:
Visual tests for the remaining power select menus so we don't have to manually check the menus whenever we upgrade powerselect.
Check out the first four snapshots: https://percy.io/percy/percy-web/builds/425139

(the others are from differences in faker data -- i'm guessing it's because there is now additional fake models being generated before those snapshots, so the seed for the generation is different?)


cadeParade:
@PixelJanitor The flash messages sort of disappear on mobile screen sizes -- they disappear very gracefully, but wondering if they should still display on mobile?

cadeParade:
@PixelJanitor oops this comment is in the wrong PR, deleting + moving now

PR #283, opened 2017-11-30
Remove power-select from org and project settings menu

cadeParade:
Power-selects take a lot of maintenance and also, the settings menu power select is broken in production. This PR removes the power-select that keeps breaking and replaces it with a more intuitive UI.

@djones and I have confirmed that the most projects a real user will see right now is 16, which fits on a full screen chrome browser on a mac book.

[Here is an export of the count of projects per org with a histogram](https://docs.google.com/spreadsheets/d/1AS7Po_HLHhcVbZerXbgSETdSXIcZ0bHFrUKLeiCNeDk/edit#gid=0).

# Old
broken-settings-menu

# New
new-settings-menu


PixelJanitor:
If there exists an action for this - Instead of not showing project settings if there are no projects, we should have an “add new project” button in that container.

PR #284, opened 2017-11-30
Convert js native alerts to flash messages

cadeParade:
Replace native JS `alerts` with flash messages.

cc @fotinakis 

~Depends on #281~ This is merged


app/services/subscriptions.js
```diff
@@ -16,9 +17,14 @@ export default Service.extend({
     savingPromise.then(
       () => {},
       () => {
-        alert(
-          'A Stripe error occurred! Your card may have been declined. Please try again or ' +
-            'contact us at hello@percy.io and we will help you get set up.',
+        this.get('flashMessages').createPersistentFlashMessage(
```
cadeParade: This is a persistent message because the next line (line 29) reloads the page immediately. Not sure why, but I've kept it

PR #285, opened 2017-11-30
upgrade ember-power-select from 1.9.11 -> 1.10.4

cadeParade:
Depends on #282


cadeParade:
I can't approve this because I opened it, but I've merged some visual tests for the powerselects before this, and there were no visual diffs detected, so this should be safe to merge!

PR #287, opened 2017-12-01
Custom CSS Color Properties

PixelJanitor:


app/styles/_panel.scss
```diff
@@ -12,11 +12,43 @@
...
     text-overflow: ellipsis;
     white-space: nowrap;
-    background: $lightest-gray;
-    border-bottom: $lighter-gray-border;
+    background: var(--lightest-gray);
+    border-bottom: var(--lighter-gray-border);
     overflow: hidden;
   }
 
+  .Panel-header--withSelect {
```
djones: some bonus changes here?

PixelJanitor: Hmm. Good catch. I didn't move these here so I'm not sure where these came from. @cadeParade did these styles get added here when you were playing with power-select styling the other day?

cadeParade: Sorry for getting to this so late!
These styles were removed in this PR : #238 
And are not present in master: https://github.com/percy/percy-web/blob/master/app/styles/_panel.scss

So it is safe to ditch these!

PR #302, opened 2017-12-02
Fix snapshot linking 

timhaines:
This PR fixes an edge case with linking to specific snapshots in builds.

If the snapshot didn't have a comparison at the selected width, and there were other snapshots that did but didn't have diffs, then a link to that snapshot would fail to scroll the page on load.

The fix was updating snapshot-list's sorting mechanism to prioritize snapshots with diffs at any widths over snapshots with no diffs at any widths.

The snapshot-list's `registerChild` method is the method that looks for new added snapshots to the list, and sets up the scroll if a snapshot with the id matching the url is added.  The `registerChild` method relies on `sortedSnapshots` to determine the snapshot id of the snapshot component that was just inserted (seems fragile - but out of scope of this fix).


app/components/snapshot-list.js
```diff
@@ -26,10 +26,26 @@ export default Component.extend({
     }
 
     return snapshots.sort(function(a, b) {
+      // Prioritize snapshots with diffs at any widths over snapshots with no diffs at any widths
+      function maxDiffRatioAnyWidth(comparisons) {
+        let comparisonWidths = comparisons.mapBy('smartDiffRatio').filter(x => x);
+        comparisonWidths.push(0); // Make sure there's at least one element in the array
```
timhaines: If there's not one element in the array, then Math.max.apply(null, []) returns -Infinity.

cadeParade: Why does this need to use `apply`? Why not `Math.max(comparisonWidths)`?
`Math.max([])` return 0 afaict.

Also not sure why we would need to make a copy of the map results on line 31 (I am guessing that's why the `.filter(x => x)` is there?). Even if we had to push 0 for the case of using `apply`, `comparisonWidths` isn't being used anywhere else or being altered, so I think not copying it is safe enough...

app/components/snapshot-list.js
```diff
@@ -41,13 +57,7 @@ export default Component.extend({
...
       }
 
-      // Second sort: by diff ratio across all widths.
-      function maxDiffRatioAnyWidth(comparisons) {
-        return Math.max.apply(null, comparisons.mapBy('smartDiffRatio').filter(x => x));
-      }
-      let maxComparisonDiffA = maxDiffRatioAnyWidth(a.get('comparisons'));
-      let maxComparisonDiffB = maxDiffRatioAnyWidth(b.get('comparisons'));
-
+      // Finally, sort by diff ratio across all widths.
```
cadeParade: I think this comment is not in the right place anymore

cadeParade:
Can we add a test for this logic? :)

app/components/snapshot-list.js
timhaines: @cadeParade Math.max doesn't like receiving arrays as params.  I think that's why the original code used apply.

app/components/snapshot-list.js
timhaines: @cadeParade I think this is still in the right location.  The return statement below does this.  Happy to discuss further.  Why do you think it's in the wrong location?

app/components/snapshot-list.js
timhaines: @cadeParade I'm not quite sure why the filter was there originally.  I guess it's removing 0's and nulls.  Perhaps unnecessary.

timhaines:
@cadeParade I suspect it was never right in the first place.  It would be great to have a series of tests for the sorting mechanism, yes.

timhaines:
It's also confusing for me (without researching it) as to why this sorting mechanism is separate from the one the decides on the on-page ordering.

And how is it consistent if something like 'snapshot name' isn't used as a second sort field for snapshots that otherwise compare as equals..

app/components/snapshot-list.js
cadeParade: Ahh, good catch with the array args. Not sure if this is any clearer, but it is shorter: 
`Math.max(0, ...comparisonWidths)` expands to `Math.max(1,values, of, comparison, widths)`, so would work.
Pushing a 0 just feels sort of weird to me.

app/components/snapshot-list.js
cadeParade: I misread the last line of the method, you're right.

app/components/snapshot-list.js
timhaines: @cadeParade Your suggestion definitely reads better to me.  Pushing a commit to update to that.  Thanks for the tip!

PR #312, opened 2017-12-08
Sticky Flash Message Styles

PixelJanitor:


config/environment.js
```diff
@@ -24,7 +24,7 @@ module.exports = function(environment) {
     flashMessageDefaults: {
       timeout: 5500,
       extendedTimeout: 500,
-      destroyOnClick: false,
```
cadeParade: Was removing this line on purpose?

app/templates/application.hbs
```diff
@@ -4,7 +4,7 @@
 {{#if flashMessages.queue}}
   <div class="flash-message-container">
     {{#each flashMessages.queue as |flash|}}
-      {{#flash-message flash=flash classNames=(concat "flash-message-" flash.type) as |component flash close|}}
+      {{#flash-message flash=flash classNames=(concat "flash-message-" flash.type (if flash.sticky ' flash-message-sticky')) as |component flash close|}}
```
cadeParade: do we still need `close` in the pipes?

app/styles/_flash_messages.scss
```diff
@@ -10,7 +10,11 @@ $flash-message-states: (
   position: fixed;
   bottom: 24px;
   left: 24px;
-  width: calc(100vw - 128px);
+  max-width: calc(100vw - 128px);
+
+  @media screen and (min-width: 768px) {
```
cadeParade: What are these two new media queries for?

PixelJanitor: for having the flash messages hang on the left corner when not on mobile and wider on mobile (but accounting for intercom's UI)

app/templates/application.hbs
PixelJanitor: we don't. Good catch. I've removed

config/environment.js
PixelJanitor: Yep! On mobile we hide the close button, so I decided to just allow the user to be able to click anywhere on the flash to close it

cadeParade: This will allow desktop users to also click anywhere to close, which seems fine to me, but just wanted to make sure we're all on the same page

PixelJanitor: Yep! 👍

PR #313, opened 2017-12-08
Add transform to prevent repaints on scroll

faun:
The scrolling performance on the builds page is poor due to repaints on scroll. Add transform to promote the element to its own layer.

https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas

Before:

build__before

---

After:

build__after


faun:
This now uses `will-change`: https://caniuse.com/#feat=will-change

cadeParade:
Another thing that causes slow rendering is rendering _lots_ of components. One alternative to components are partials -- they do not cause extra rendering time, but they also do not have their own logic file or context.

Another solution I've seen was lazy rendering -- only render things that are in the viewport. This helped performance of pages with lots of components considerably, so if it is still a problem, we could consider this. 
https://github.com/DockYard/ember-in-viewport

timhaines:
When scrolling, the frame rate stays above 40fps for me in Kaelig's build with this change..  It was getting down to 20fps on master.

PixelJanitor:
I'm not sure why there are repaints on scroll on this page. Are we calculating something onScroll that's controlling layout?

PR #319, opened 2017-12-11
Remove Github Specific Attributes

djones:
Goes hand in hand with API PR284


app/components/user-name.js
```diff
@@ -1,4 +1,3 @@
-import {computed} from '@ember/object';
 import Component from '@ember/component';
 
 export default Component.extend({
```
cadeParade: This whole file can be deleted (you've deleted the whole template below, and in the app there's only the three uses you've deleted already)

djones:
@cadeParade no failing tests... definitely a sign of not enough coverage here.

PR #325, opened 2017-12-13
Allow customized redirect route when closing lock modal

cadeParade:
When I first wrote the redirect logic for closing the lock modal, there were only two instances, the home page instance and the login route instance. Only the login route instance needed custom behavior. 
Now there are four instances of the lock modal on the site --
Home page
login
profile > reset password
profile > sign in with github 

and only one of them needs to redirect. 

So this PR makes the redirect on close null by default so it stays on the same page, and passes in an option to the login instance to redirect to `/`.


tests/unit/mixins/ensure-stateful-login-test.js
```diff
@@ -122,9 +122,10 @@ describe('EnsureStatefulLoginMixin', function() {
     });
 
     it('transitions to root', function() {
-      subject._onLockClosed();
+      const redirectRoute = 'route';
```
fotinakis: Thanks for improving our testing so much to be able to make these kinds of simple test updates!

PR #332, opened 2017-12-18
Updated Approval Header

PixelJanitor:
Spent my flights this morning building out an update approval header. Added hover animations and a loading animation when you click the approve button that hooks into the api call for loading state. 
Would like to pair with someone on the Ember logic since I just plugged in some ugly behavioral js to make it work.

image

See https://percy-io.slack.com/archives/C59CZC734/p1513617208000296 for video of interactions!


app/components/user-avatar.js
```diff
@@ -1,13 +1,10 @@
...
 
 export default Component.extend({
   user: null,
   classes: null,
   linked: false,
-  width: 40,
-  height: readOnly('width'),
 
-  classNames: ['UserAvatar'],
+  classNames: ['current-user br ml-4'],
```
fotinakis: We often use the `user-avatar` component for users who are not the "current" user, so this is a bit too specific of a class to apply here.

app/components/approval-button.js
```diff
@@ -6,27 +6,29 @@ import utils from '../lib/utils';
 export default Component.extend({
   build: null,
   approve: null,
-
   isApproved: alias('build.isApproved'),
+  isLoading: false,
   tagName: 'button',
-  classNames: ['ApprovalButton', 'Button', 'Button--withLeftIcon'],
-  classNameBindings: ['classes', 'isApproved:ApprovalButton--approved'],
+  'data-js-approve-button': true,
```
fotinakis: Drop this and the attribute binding below? They seem unused.

app/templates/components/build-toolbar.hbs
```diff
@@ -1,32 +1,26 @@
...
-        {{trial-bar organization=organization}}
-      </div>
+<header class="main-header main-header-alt flex space-between align-center px-3">
+  <div class="flex align-center">
+    {{#link-to "index"}}
+      <figure class="svg-container mr-3">
+        {{inline-svg "percy-logo-on-light"}}
+      </figure>
+    {{/link-to}}
+    <div class='current-organization br px-2 py-1'>
```
fotinakis: Keep double quotes here for consistency.

app/templates/components/build-width-selector.hbs
```diff
@@ -1,5 +1,5 @@
-<select onChange={{action 'updateSelectedWidth' value='target.value'}}>
-  <option selected={{noWidthSelected}} value="select-width">Select width</option>
+<select class="alt-select form-control m-0 pl-3 pr-5" onChange={{action 'updateSelectedWidth' value='target.value'}}>
+  <option selected>Select width</option>
```
fotinakis: I believe you need to keep `selected={{noWidthSelected}}` here to keep certain UX behaviors working (I believe it's for this flow:
1) Pick a width from the dropdown
2) Choose a different width in a snapshot's responsive nav
3) Notice that the global nav switcher has switched backed to "Select width..."

app/styles/_header.scss
```diff
@@ -121,3 +121,31 @@
...
+  height: var(--header-height);
+  background: var(--dark-purple);
+}
+
+.main-header-alt {
+  background: var(--lightest-gray);
+  border-bottom: 1px solid var(--lighter-gray);
+}
+
+.current-organization {
```
fotinakis: Shouldn't you nest this inside of `main-header`? Seems very context specific to the header and also prone for conflict if global.

app/styles/components/_user_avatar.scss
```diff
@@ -1,21 +1,5 @@
...
-    border-radius: 4px;
-    align-items: center;
-    justify-content: center;
-
-    .fa {
-      font-size: 2rem;
-      color: var(--gray);
-    }
-  }
+.current-user {
```
fotinakis: Same comment as above, we should keep the `user-avatar` component generic for all users, and not specific to the "current" user.

app/templates/components/user-avatar.hbs
```diff
@@ -1,7 +1,5 @@
 {{#if (or showPlaceholder (not user))}}
-  <div class="UserAvatar-placeholder" style="width: {{width}}px; height: {{height}}px">
-    {{fa-icon "user"}}
-  </div>
+  <img src="images/placeholder-avatar.jpg">
 {{else}}
-  <img width="{{width}}" height="{{height}}" src="{{user.avatarUrl}}&s={{width}}">
```
fotinakis: Did you check and find all the places where `user-avatar` is passed width or height, and remove those since we're removing that functionality here?

app/components/approval-button.js
PixelJanitor: Good catch - done.

app/components/user-avatar.js
PixelJanitor: I should have called this user-avatar in the first place since that's the name of the component. Done

app/styles/_header.scss
PixelJanitor: Thanks for holding me to this. This was a great use case of utility classes and shouldn't have added this css at all. Pushed a commit of a waaaaaaaay nicer solution and less css 👍

app/styles/components/_user_avatar.scss
PixelJanitor: Done

app/templates/components/build-toolbar.hbs
PixelJanitor: 👍

app/templates/components/build-width-selector.hbs
PixelJanitor: Ah this is a case I didn't catch. It seemed unused and I couldn't an unintended behavior without it. I added it back in and tried to reproduce with good results 👍

app/templates/components/user-avatar.hbs
PixelJanitor: It was only used in the user settings with a width passed to it of the same size as the header use. So it's all fixed and polished now. Thanks for pointing this out.

app/templates/components/build-toolbar.hbs
```diff
@@ -1,32 +1,26 @@
...
+    <div class=" flex align-center bg-purple-300 f6 rounded px-2 py-1 hide-on-mobile">
+      {{#link-to "organization" organization class="text-mono text-purple truncate truncate-target hide-on-tablet"}}{{organization.name}}{{/link-to}}
+      <span class="text-weight-semibold mx-1 hide-on-tablet">/</span>
+      {{#link-to "organization.project" organization project class="text-mono text-purple truncate truncate-target"}}{{project.name}}{{/link-to}}
     </div>
   </div>
-</div>
+  {{trial-bar organization=organization}}
+  <div class="flex align-center">
+    {{#if true}}
```
fotinakis: This looks like leftover debugging :)

PixelJanitor: Dammit - I've been trying to make sure that those temp statements don't make it in. These have been toggled 97626310823 times lately and this slipped through. Good catch - committing change now.

fotinakis:
LGTM 🌮  Nice work!

PR #340, opened 2018-01-02
Updated build list UI

PixelJanitor:


app/styles/components/_wip.scss
```diff
@@ -0,0 +1,3 @@
+.css-wip {
```
cadeParade: ?

PixelJanitor: A nifty little utility class that I throw on elements that I need to address at some point in the PR and don't want to forget about. This prevents the build from passing if it's in the codebase :)

app/components/build-card.js
```diff
@@ -8,8 +9,30 @@ export default Component.extend({
...
+  attributeBindings: ['style'],
+  style: computed('status', function() {
+    return `--status-color: ${this.get('status.color')};`;
+  }),
+
+  classNames: ['build-card c-pointer'],
+  classNameBindings: ['classes', 'build.isExpired:opacity-50'],
+  status: computed('build.state', function() {
+    if (this.get('build.isPending') || this.get('build.isProcessing')) {
+      return {name: 'Processing', color: 'var(--gray-200)'};
```
cadeParade: A few things about this:
- Separate display logic and business logic
- Make this reusable 
 
1. We should extract these state strings to be constants so the state labels appearing throughout the app are the same. What I mean by this is above the component definition have constant declarations like: 
```
const APPROVED_LABEL = "Approved"
const NEEDS_REVIEW_LABEL = "Needs Review"
etc
```

2. I am still uneasy about linking these pieces of display logic together. For one, it seems unnatural to me to generate literal style strings in the js rather than generating a class name which has a corresponding style. Perhaps a custom helper to do this logic once and a few related helpers would be a good solution. 
Or add a `label` property to the build model...yes I like that direction...I can take a stab at this or we could pair

app/styles/base/_typography.scss
```diff
@@ -1,15 +1,44 @@
...
 kbd,
 pre,
 samp {
   font-family: $monospace-font;
-  font-size: 1.3rem;
+  font-size: var(--f6-size);
+  letter-spacing: 0.5px;
+}
+
+code {
```
cadeParade: Are the changes to the code blocks in this pr also?
In general it would be helpful to split discrete chunks of work into different prs. This PR I thought was about updating the builds page, so it seems out of scope to have the code block formatting in this PR also. 
Also it would keep the PR size down...this PR is 60 files which is a huge amount to review.

app/styles/components/_status_pill.scss
```diff
@@ -0,0 +1,37 @@
...
+  padding: var(--spacer-0) var(--spacer-2);
+  font-size: var(--f6-size);
+  font-weight: var(--semibold);
+  color: var(--text-color);
+  text-align: center;
+  white-space: nowrap;
+  background: var(--background-color);
+  border-radius: 13px;
+
+  &.is-processing {
```
cadeParade: Here are separate classes for every visual state we want to be displayed. Is there a way we can refactor this to use with the little vertical bars on the left so we don't have to make a class for each state and _also_ generate a css string literal?

app/components/build-card.js
PixelJanitor: Yea let's pair on it. Would love to learn how you'd do that.

app/styles/components/_wip.scss
cadeParade: TIL

app/templates/components/build-card.hbs
```diff
@@ -1,14 +1,72 @@
...
+      {{/if}}
+    {{/if}}
+    <strong class="no-wrap mr-2">{{build.commit.authorName}}</strong>
+    <span class="text-light truncate">{{build.commit.message}}</span>
   </div>
-  <div class="two columns">{{build-state build=build classes="u-pull-right"}}</div>
+</div>
+<div>
+  <div class="flex align-center">
+    {{#if (or build.isPending build.isProcessing)}}
```
cadeParade: Let's make a new property on the model, something like `build.isActive` or something, so we don't have to keep checking this `or` condition,

app/templates/components/main-footer.hbs
```diff
@@ -1,31 +1,29 @@
...
-    </a>
-  </div>
-  <div class="grid col-auto gap-2 align-center justify-center main-footer-link-set-right">
-    {{#link-to "security" class="main-footer-item"}}Security{{/link-to}}
-    {{#link-to "terms" class="main-footer-item"}}Terms{{/link-to}}
-    {{#link-to "privacy" class="main-footer-item"}}Privacy{{/link-to}}
-    {{#link-to "team" class="main-footer-item"}}Team{{/link-to}}
-  </div>
-</footer>
+<div class="grid col-auto gap-2 align-center justify-center main-footer-link-set-left">
```
cadeParade: Another candidate to be in a different PR. Having this change in this PR makes the percy snapshots for this pr very noisy and shows many diffs not related to the subject at hand (updating builds page).

app/templates/components/project-container.hbs
```diff
@@ -1,70 +1,79 @@
...
-    <a {{action "refresh"}} href="#" class="Button Button--withLeftIcon u-pull-right">
-      {{#if isRefreshing}}
-        {{fa-icon "refresh" classNames="fa--left fa-spin"}}
-      {{else}}
-        {{fa-icon "refresh" classNames="fa--left"}}
+<header class="sticky bg-white z-top">
+  <div class="flex space-between align-center py-4 px-5">
+    <div class="flex align-center">
+      <h2 class="m-0 mr-3 f2 text-weight-regular">{{project.name}}</h2>
+      {{#if project.isDisabled}}
```
cadeParade: Can you give a summary of the changes on this file? It seems very different but not sure why.

app/templates/organization/index.hbs
```diff
@@ -1,44 +1,36 @@
...
-    {{else}}
-      {{#if model.savedProjects}}
-        {{#link-to "organizations.organization.projects.new" model.slug classNames="Button u-pull-right"}}
-          New Project
+{{!-- Note: use model.projects here to get the promise proxy, but use savedProjects below for filtering. --}}
+{{#if model.projects.isPending}}
+  {{loading-page}}
+{{else}}
+  {{#if model.savedProjects}}
+  <header class="flex space-between align-center py-4 px-5">
```
cadeParade: Indenting is off here

tests/acceptance/organization-test.js
```diff
@@ -20,12 +20,12 @@ describe('Acceptance: Organization', function() {
         expect(currentPath()).to.equal('organization.index');
       });
 
-      click('a.ProjectCard-link');
+      click('a[data-test-project-link]');
```
cadeParade: would `[data-test-projet-link]` and `[data-test-settings-icon]` below be sufficient selectors? I think so. No need to link it to a tag type for ease of future developers.

app/styles/base/_typography.scss
PixelJanitor: Right. These types of UI updates tend to be huge, and share alot of changes too. There are changes to the `code` blocks in this PR that affect these pieces of UI as well as global code blocks.

app/templates/organization/index.hbs
PixelJanitor: ✅

app/templates/components/project-container.hbs
PixelJanitor: - semantic updates
- header sticks to top when scrolling
- project list is now full width to support new full width build cards
- has a table style thead type element in the header to track what column is what while scrolling
- updated build cards per design

app/templates/components/main-footer.hbs
PixelJanitor: You're absolutely right. This was a distraction change that happened when I found it and I should have pulled this out into a new PR. You and I discussed this in Slack and it will stay in this PR this time, but I will be better about it in the future.

app/components/build-card.js
PixelJanitor: PUSHED. So. Much. Better. Thanks for your help!

app/styles/components/_status_pill.scss
PixelJanitor: We now have separate status classes for the build cards :)

app/templates/components/build-card.hbs
PixelJanitor: ✅

tests/acceptance/organization-test.js
PixelJanitor: Ah you're right. Didn't think about it not needing to be scoped to an element 👍

app/templates/components/project-container.hbs
cadeParade: Thanks!

app/styles/components/_status_pill.scss
cadeParade: thanks :D <3

app/templates/components/build-card.hbs
```diff
@@ -1,14 +1,74 @@
-<div class="row">
-  <div class="two columns">
-    {{!-- NOTE: we use the baseBuild.id instead of the object to force the next build to reload so all relationships are available. TODO: better way? --}}
-    <h6>{{#link-to "organization.project.builds.build" build.project.organization.slug build.project.slug build.id}}{{build.buildTitle}}{{/link-to}}</h6>
+<div class="build-card c-pointer {{if build.isExpired 'opacity-50'}} is-{{dasherize build.buildStatusLabel}}" {{action 'navigateToBuild'}}>
```
cadeParade: P4 double quotes in hbs files

tests/acceptance/organization-test.js
```diff
@@ -20,12 +20,12 @@ describe('Acceptance: Organization', function() {
...
       });
 
-      click('a.ProjectCard-link');
+      click('[data-test-project-link]');
       andThen(() => {
         expect(currentPath()).to.equal('organization.project.index');
       });
 
-      click('.ProjectContainer a:contains("Settings")');
+      click('.project-container [data-test-settings-icon]');
```
cadeParade: Does this need to be scoped to `.project-container`? do we need the class `.project-container`?

app/components/main-footer.js
```diff
@@ -2,5 +2,7 @@ import {inject as service} from '@ember/service';
 import Component from '@ember/component';
 
 export default Component.extend({
+  tagName: 'footer',
+  classNames: 'main-footer container-full grid justify-center py-4',
```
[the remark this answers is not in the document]
PixelJanitor: I'm assuming the array version of this would be `['foo', 'bar', 'baz']`? Probably works since i'm just putting all of the classes in one string instead of multiple?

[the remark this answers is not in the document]
PixelJanitor: Just pushed up a commit that pulls the css out into the hbs template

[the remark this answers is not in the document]
cadeParade: Yes, it would be in an array as in your comment above. :)

PR #341, opened 2018-01-03
Improve tabbing behavior for home screen

cadeParade:
A user reported a bug where they couldn't tab navigate to the dashboard or sign in button.

This PR converts three home page buttons to clickable components, adds a tab index, and an action where if the component is focused and the user pushes return key, it will trigger the click action.


faun:
I added another story for the profile and logout menu items: https://app.clubhouse.io/percy/story/1319/add-keyboard-access-to-profile-and-logout-menu-accessed-from-the-user-s-avatar-on-the-top-right-of-the-page

PR #344, opened 2018-01-04
Fake admin feature flag

cadeParade:
Check localStorage to see if correct admin key and value are present when showing stuff related to username/password login.

Places that are checked:
Login modal:
- Show email/password + github if admin
- Show only github if not admin

Connected accounts screen:
- Show button to add username/password account if admin
- Show only github account if not admin


cadeParade:
Decided not to do this

PR #346, opened 2018-01-04
Add polling to builds list page

cadeParade:
See the most boring gif in the world:
Builds page polls for updates every 10 seconds...gif is not to time scale, but you can see it polling and updating!
poll

This PR:
- Makes a mixin to handle polling mechanics (only polls when document is in focus)
- Mixes in to components with polling to remove duplication of polling mechanics
- Converts polling part of `setup/github-app` route to a component so it can use the mixin.


app/components/build-container.js
```diff
@@ -24,34 +19,13 @@ export default Component.extend({
...
-          // ember-concurrency is planning on adding a better way to cancel tasks during testing
-          // https://ember-concurrency.com/#/docs/testing-debugging
-          if (Ember.testing) {
-            this.get('runningTask').cancel();
-          }
-        });
-      yield timeout(POLLING_INTERVAL_SECONDS * 1000);
-    }
-  }).drop(),
+  pollFunction() {
```
djones: Perhaps `pollRefresh` instead?

app/components/build-container.js
```diff
@@ -1,13 +1,8 @@
-import {on} from '@ember/object/evented';
 import {max, oneWay, or} from '@ember/object/computed';
 import Component from '@ember/component';
-import Ember from 'ember';
-import {task, timeout} from 'ember-concurrency';
+import PollingMixin from 'percy-web/mixins/polling';
```
djones: This mixin really tidies things up. Nice 👍

app/templates/setup/github-app.hbs
```diff
@@ -1,14 +1,4 @@
...
-    <div class="Card Card--extraPadded u-text-center">
-      <h3>GitHub App processing...</h3>
-      <p>
-        {{fa-icon "fa-spinner fa-pulse"}}
-        Waiting for GitHub to complete the installation.
-      </p>
-    </div>
-  </div>
-</div>
+  {{setup-github-app installationId=installationId transitionToOrgSettings=(route-action "transitionToOrgSettings")}}
```
fotinakis: How about naming that `afterAppInstalled`? Like `afterAppInstalled=(route-action "transitionToOrgSettings")`

djones:
It'll be interesting to see how this does with `master` now that we have the project list on the left and there are status icons that will change.

We'd want to refresh that list too on each poll update.

cadeParade:
@djones I've checked this with the new sidebar, and the project list status bubble _does_ update when the builds are refreshed. Yay Ember.

djones:
@cadeParade Just Works™

app/components/build-container.js
fotinakis: Yes this is soooo 💯

fotinakis:
Have QA'd this — it's awesome! Ship it! LGTM

cadeParade:
Ok, I've renamed the two suggested things, will merge after the tests pass!

PR #347, opened 2018-01-05
Do not allow github integration without github identity

cadeParade:
The main part of this PR is wrapping the `github-settings` panel in an `if` statement and checks whether or not the user has a github identity.
View pr without whitespace diffs here: https://github.com/percy/percy-web/pull/347/files?w=1

Secondary changes are:
- converting the user model properties re: identities to be proper promises
- Installing ember-promise-helpers because waiting for promises in the template sucks :(

settings_-_percy___continuous_visual_reviews_for_web_apps


app/components/organizations/github-settings.js
```diff
@@ -34,6 +34,8 @@ export default Component.extend({
     },
   ),
 
+  isEligibleForGithubIntegration: alias('currentUser.hasGithubIdentity'),
```
djones: Nice naming here 👍

app/models/user.js
```diff
@@ -10,19 +10,26 @@ export default DS.Model.extend({
 
   identities: DS.hasMany('identities'),
 
+  hasGithubIdentity: computed('identities.@each.provider', function() {
+    return DS.PromiseObject.create({
```
djones: Could possibly have `hasGithubIdentity` and `hasEmailPasswordIdentity` call a single private function that checks for an identity with a provider name.

Just looking to reduce the duplication here, since these two function are very similar.

PR #352, opened 2018-01-09
Remove optional Github setup step when creating an organization

djones:
Bonus flash message change to stay on screen for 10 seconds.


fotinakis:
Nice!

I think if we're going to officially move this step, we need to make it more clear that after adding the integration you need to go link repos to projects as well for it to work. Maybe just a flash message that says "Remember: to complete GitHub app setup, link your project to a GitHub repo in Percy's project settings."

config/environment.js
```diff
@@ -22,7 +22,7 @@ module.exports = function(environment) {
     },
 
     flashMessageDefaults: {
-      timeout: 5500,
+      timeout: 10000,
```
cadeParade: We want every flash message to display for 10 seconds by default?

djones: @fotinakis preferred it

PR #353, opened 2018-01-09
Redirect to /api/auth/logout after session invalidation

faun:
https://app.clubhouse.io/percy/story/1371/signing-out-of-percy-does-not-delete-http-only-cookies


app/services/session.js
```diff
@@ -35,6 +35,14 @@ export default SessionService.extend({
     }
   },
 
+  invalidateAndLogout() {
+    this.get('session')
+      .invalidate()
+      .then(() => {
+        window.location.replace('http://dev.percy.local:4200/api/auth/logout');
```
cadeParade: This needs to be replaced with the original `/api/auth/logout`

cadeParade:
Also there are some tests failing which I can help you with

PR #354, opened 2018-01-10
Update Auth0 client ID and domain in development

faun:
Configuration changes to fix issues with Auth0 configuration in development


cadeParade:
I think this is fine

PR #356, opened 2018-01-11
Display a friendly message after password change

djones:
Currently we just send users to the login page with no message after their password is updated. This PR sends them to a friendly page to confirm password change instead.

Once this is merged the Auth0 password updated email redirect will need to be set to:

http://dev.percy.local:4200/auth/password-updated (for dev)
https://percy.io/auth/password-updated (for prod)

screen shot 2018-01-11 at 7 52 18 am


tests/acceptance/password-updated-test.js
```diff
@@ -0,0 +1,23 @@
+import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
+import stubLockModal from 'percy-web/tests/helpers/stub-lock-modal';
+
+describe('Acceptance: PasswordUpdated', function() {
+  setupAcceptance({authenticate: false});
```
cadeParade: This is interesting. This route does not have a js file, so I guess you can get to it authenticated or unauthenticated.
That seems fine? But almost certainly when a user sees this page they will be logged in, yes? 
If the user is authenticated, I don't think this test would need `stubLockModal`.

Anyway, I think this is fine as is, just spewing some pre-coffee thoughts 👍

tests/acceptance/password-updated-test.js
```diff
@@ -0,0 +1,23 @@
...
+
+  it('shows page with password update message', function() {
+    stubLockModal(this.application);
+
+    visit('/auth/password-updated');
+    percySnapshot(this.test.fullTitle());
+
+    click('.test-password-updated-signin-link');
+    andThen(() => {
+      expect(currentURL()).to.equal('/login');
```
cadeParade: I think most users will be logged in at the time they reach this page so it might be more "golden path" to test the signed in case.

PR #357, opened 2018-01-11
profile -> settings.profile on return to profile link

djones:


fotinakis:
Thanks! Sorry I missed that.

cadeParade:
I also missed it.
But David caught it.
Yay David 👍

PR #367, opened 2018-01-13
Refactor panels into components

cadeParade:
Take these things:

And make them a single component.


faun:
Q: why are your diffs[erent]?

cadeParade:
@faun they are only _slightly_ different which I am gong to call close enough to be the "same" in this case

cadeParade:
@PixelJanitor Just an FYI, in case you need to use this style of panel again, it is now in a component! Also in case we want to change how these panels look, it is in a component!

PixelJanitor:
🎉

PR #368, opened 2018-01-15
Upgrade packages

timhaines:
Updated a bunch of packages.  Read their changelogs aside from the linting ones.


app/templates/application.hbs
```diff
@@ -1,3 +1,4 @@
+{{head-layout}}
```
timhaines: This was added for the [4.0 release](https://github.com/tim-evans/ember-page-title/releases/tag/4.0.0) of ember-page-title.

PR #370, opened 2018-01-15
Updated build status for Receiving Build

PixelJanitor:


PixelJanitor:
@timhaines and @fotinakis Here is the text change you mentioned for when we're receiving builds.

PixelJanitor:
That is correct. I haven't worked on that header yet, but will continue to mirror this verbiage based on the status over there too. Merging!

PR #371, opened 2018-01-16
Password resets via API

djones:
Removes the need for us to use the Lock.js dialog to reset a password. Looks like this:

cf7qqnk2bb

Will see about getting a primary button loading state from @PixelJanitor as there is a small delay when you click the button while it's doing the XHR request.

Sister PR: https://github.com/percy/percy-api/pull/336


PixelJanitor:
I made a slight primary button color adjustment and added an `is-loading` class that's available to primary buttons now for you to use here. Let me know if you need anything else!

djones:
Thanks @PixelJanitor. I've got that integrated now. It looks good in Chrome but in Safari after the loading state has been applied and removed there's a bar that stays on the button... looks like this:

screen shot 2018-01-16 at 11 13 14 am

PixelJanitor:
Hmmm. I didn't see that happening. I'll try to reproduce and get you a fix.

PixelJanitor:
Ok, I pushed a fix. When an element like this changes in safari, apparently it needs some sort of repaint to re-draw the original gradient. I'm not happy about this fix, but it's the ugly truth.

app/mixins/ensure-stateful-login.js
```diff
@@ -66,16 +66,6 @@ var EnsureStatefulLogin = Mixin.create({
     this.get('flashMessages').success("We've sent an email to the address you've entered.");
   },
 
-  showResetPasswordModal() {
```
cadeParade: Yay

app/models/user.js
```diff
@@ -15,6 +15,10 @@ export default DS.Model.extend({
   }),
 
   hasEmailPasswordIdentity: computed('identities.@each.provider', function() {
+    return this.get('emailPasswordIdentity');
+  }),
+
+  emailPasswordIdentity: computed('identities.@each.provider', function() {
```
cadeParade: Thanks for fixing this. My bad.

app/styles/_buttons.scss
```diff
@@ -234,16 +234,23 @@ button,
...
     background: var(--purple-gradient);
     transform: translate3d(0, -2px, 0);
   }
+
+  &.shadow-on-light {
+    box-shadow: 0 2px 8px 0 rgba(46, 0, 62, 0.26), 0 8px 20px 0 rgba(39, 0, 52, 0.08);
+  }
+
+  &.is-loading {
+    display: table; /* safari needs a forced repaint to draw the gradient again after loading is finished */
```
cadeParade: ¯\_(ツ)_/¯

tests/acceptance/password-change-request-test.js
```diff
@@ -0,0 +1,35 @@
+import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
+import PasswordChangeRequest from '../pages/components/password-change-request';
+
+describe('Acceptance: PasswordChangeRequest', function() {
+  setupAcceptance({authenticate: true});
+
+  setupSession(function(server) {
+    const user = server.create('user');
+    this.user = user;
+    server.create('identity', 'auth0Identity', {user});
```
cadeParade: Not required, but you could add a trait for `user` similar to github identity here: https://github.com/percy/percy-web/blob/master/mirage/factories/user.js#L18

tests/acceptance/password-change-request-test.js
```diff
@@ -0,0 +1,35 @@
...
+      expect(PasswordChangeRequest.isPasswordChangeButtonPresent).to.equal(true);
+    });
+
+    percySnapshot(this.test.fullTitle());
+  });
+
+  it('shows page with change password sent message', function() {
+    PasswordChangeRequest.visitUserSettingsPage();
+    andThen(() => {
+      PasswordChangeRequest.submitRequest();
```
cadeParade: Not required --  I haven't used this feature much, but some of the ember page object methods are chainable, so I _think_ you could do 
```
PasswordChangeRequest
  .visitUserSettingsPage()
  .submitRequest()

andThen(() => {whatever})
```

djones: I wasn't able to get this working... I tried though!

tests/acceptance/password-change-request-test.js
djones: Fixed with 6fe43de7f0f7edb96a9f518b6174d53bbb550434

djones:
Holding the merge of this one until https://github.com/percy/percy-api/pull/336 has shipped.

PR #372, opened 2018-01-16
Keratin - Percy's design system

PixelJanitor:


fotinakis:
@PixelJanitor can you add a bit of a description of what is changing here? Needs more context! :)

PixelJanitor:
Absolutely:

I added a `Keratin` style directory into the styles directory that houses all of our base styles, colors, typography, etc (and eventually the rest of the variables). I've moved our utility classes to this new style system and added a bunch of nifty new ones. The spacing strategy is broader now so required some class name changes to keep our same layout so you'll see those here too. The utility class cleanup fixed some layout spacing inconsistencies in the build cards so you'll find some slight updates in the Percy snapshots too.

The next step will be documenting this in an online style guide to reference!

PR #373, opened 2018-01-16
Cadeparade/ch1398/password reset always reports success even

cadeParade:
password-reset-status

- Convert password reset templating to component for easier testing (also integration tests run way faster than acceptance tests because it doesn't have to boot the whole app, so more integration tests is better)
- Alternate thumbs up and big red x depending on which query parameter is in the password reset route.


djones:
Note this PR has a sister PR: https://github.com/percy/percy-api/pull/333

app/templates/components/password-updated-status-panel.hbs
```diff
@@ -0,0 +1,29 @@
+{{#if (eq success 'true')}}
+
+  {{#status-panel
+    inlineSvgName="thumbs-up-large-icon"
+    mainText="Success!"
+    secondaryText="Your password has been updated"
+  }}
+    {{#if currentUser}}
```
djones: Optionally DRY up this Continue to Profile/Sign In block.

cadeParade:
@djones there are page objects now 👍

PR #375, opened 2018-01-18
New Project Page

PixelJanitor:


PixelJanitor:
We added access to a bold font weight on Typekit, and the side effect of that change in some places is showing up in this PR. Just FYI.

PR #381, opened 2018-01-22
Conditional commit info

PixelJanitor:
When I run tests locally it doesn't leave a commit message to show and makes the build card detail broken. This fixes that.

image

becomes this:

image


PixelJanitor:
Apparently, we need a Percy test setup here because it should have thrown a diff.

cadeParade:
these screenshots look the same?

PixelJanitor:
The icon under the branch name that has no commit info is gone in the second

PR #382, opened 2018-01-23
Snapshot viewer (mostly header) tests

cadeParade:
In preparation for refactoring the snapshot viewer header, I've written tests.


tests/acceptance/build-test.js
```diff
@@ -2,6 +2,7 @@ import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
 import freezeMoment from '../helpers/freeze-moment';
 import moment from 'moment';
 
+// TODO convert this file to use page objects
```
djones: 👍

app/templates/components/snapshot-viewer-full.hbs
```diff
@@ -21,11 +22,12 @@
           buildWidths=buildWidths
           snapshotSelectedWidth=snapshotSelectedWidth
           updateSelectedWidth=(action 'updateSelectedWidth')
+          data-test-SnapshotViewer-widthSwitcher=true
         }}
       </div>
       <div class="btn-group" role="group" aria-label="Second Group">
         <button type="button" class="ToggleFullViewButton btn btn-square minimize-snapshot"
-          {{action (route-action 'closeSnapshotFullModal' build.id snapshotId)}}>
+          {{action closeSnapshotFullModal build.id snapshotId}} data-test-SnapshotViewer-toggleFullScreen>
```
djones: interesting... how does one avoid `route-action` here?

cadeParade: It is now in the route template. Doing some further reading on route-actions, it should not be used in components like this, but should be passed down from routes.
It also makes integration tests fail if they are used in components (but they work in the app).

cadeParade: Hmm, I guess it was always there, and I just removed this unnecessary re-casting of this action.
https://github.com/percy/percy-web/blob/master/app/templates/organization/project/builds/build/snapshot.hbs#L9

PR #383, opened 2018-01-23
Create New Account without Lock.js

djones:
Currently when a user is logged in and they want to add an email/pass identity, the user has to fill out a Lock.js dialog like this:

screen shot 2018-01-16 at 3 29 05 pm

This works but isn't ideal because they have to enter their name and email address in when we already know what those are. There's also opportunity for them to enter in a different email address to the one stored on their profile.

This PR is the `percy-web` portion of fixing that. This PR introduces a new form which allows you to simply set a password and be done. No lock.js dialog required. It looks like this:

screen shot 2018-01-23 at 10 40 06 am

Sister PR here https://github.com/percy/percy-api/pull/341


cadeParade:
Just thinking .... What happens when a user requests a password change, doesn't complete it, changes their email, and completes that verification. Does this scenario work?

djones:
@cadeParade good thought but that's somewhat outside the scope of this PR. This PR is solely focused on creating a new account where there is no verification needed.

app/components/forms/account-new.js
```diff
@@ -0,0 +1,65 @@
...
+          type: 'POST',
+          url: utils.buildApiUrl('identities'),
+          data: {data: {attributes: {password: desiredPassword}}},
+        })
+          .done(() => {
+            this.set('isSaveSuccessful', true);
+            this.sendAction('saveSuccess');
+          })
+          .fail(response => {
+            let errorData = response.responseJSON['errors'];
```
djones: I might add a check here to see if there are errors first as that may not always be the case in the response.

app/validations/account-new.js
```diff
@@ -0,0 +1,19 @@
+import {validateLength, validateFormat} from 'ember-changeset-validations/validators';
+
+export default {
+  password: [
```
djones: This matches our settings in Auth0

app/templates/settings/connected-accounts.hbs
```diff
@@ -1,10 +1 @@
...
-  <h2 class="f1 text-weight-regular">Connected Accounts</h2>
-
-  {{connected-accounts-panel
-    identities=model
-    deleteIdentity=(route-action "deleteIdentity")
-    addIdentity=(route-action "addIdentity")
-  }}
-
-</div>
+{{connected-accounts-list model=model}}
```
cadeParade: I would recommend renaming `model` to something more descriptive as soon as possible (which is here).
Previously it was `identities` so I am guessing it is identities here?
so something like 
`{{connected-accounts-list identities=model}}`.
Then in the `connected-accounts-list` component, rename `model` to `identities` and everyone knows what we're talking about 👍

tests/acceptance/connected-accounts-test.js
```diff
@@ -0,0 +1,68 @@
+import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
```
cadeParade: Optionally convert to path beginning with `percy-web`

tests/acceptance/profile-test.js
```diff
@@ -1,5 +1,5 @@
 import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
-import UserSettingsPageObject from '../pages/user-settings';
+import UserSettingsPageObject from '../pages/user-settings-page';
```
cadeParade: thanks for fixing this 😬

app/components/connected-accounts-list.js
```diff
@@ -0,0 +1,55 @@
...
+  store: service(),
+  showCreateAccount: false,
+  actions: {
+    accountCreated() {
+      this.set('showCreateAccount', false);
+
+      this.get('flashMessages').success('You may now login using this account.', {
+        title: 'Email/password account created!',
+      });
+      this.get('model').reload();
```
cadeParade: Renaming this to `identities` would make it clearer what this is reloading. Model is so vague 😢

cadeParade: (related comment below)

app/components/forms/account-new.js
cadeParade: Yes that's a great idea

app/components/connected-accounts-list.js
djones: Fixed with de08fabf2a713f637a63bf6fefeb4dd96dcfc67f

app/components/forms/account-new.js
djones: Fixed and improved testing with 1363818314fd583bc276df39a1ac0686dc79a7c4

tests/acceptance/connected-accounts-test.js
djones: Fixed with db0c815a70fe51a7862be7b087ae795833571ff5

PR #384, opened 2018-01-23
Move to Lock 11

djones:
This fixes 1 of 2 deprecations we face with Auth0. Once this is deployed we can disable "Legacy Lock API" in Auth0.

I've tested a bunch of flows manually and found no issues 🎉


djones:
@faun at least it's locked to a specific commit 😎 

For reference if this brings us any troubles... this is exactly what has changed: https://github.com/auth0-community/ember-simple-auth-auth0/compare/b53ab33176d9b59086a667bd4b55ed95430044b0...813fb5b85bba79240621dbf38fe95f33c028ec3d

PR #385, opened 2018-01-24
Consolidate snapshot viewer header

cadeParade:
The headers for `snapshot-viewer` and `snapshot-viewer-header` are almost the same so I've made them one component.
It was much more complicated than I originally expected, but I think it's in a good state now. And I learned a lot about the snapshot viewer!

There was a bunch of code left from old versions of Ember that are now unnecessary which I've deleted also.

There are no tests because I wrote all the tests in this PR https://github.com/percy/percy-web/pull/382 in preparation for this PR.
The tests caught a bug!


app/components/snapshot-viewer.js
```diff
@@ -87,9 +87,5 @@ export default Component.extend({
       this.set('snapshotSelectedWidth', value);
       this.get('snapshotWidthChangeTriggered')();
     },
-
-    showSnapshotFullModalTriggered(snapshotId, snapshotSelectedWidth) {
```
cadeParade: Not necessary with closure actions

app/templates/components/comparison-switcher.hbs
```diff
@@ -4,7 +4,7 @@
       selectedWidth=selectedWidth
       width=width
       comparisons=comparisons
-      updateSelectedWidth=(action 'updateSelectedWidth')
```
cadeParade: This is now already a closure action

app/components/comparison-switcher.js
```diff
@@ -1,29 +1,16 @@
...
 
 export default Component.extend({
-  snapshotSelectedWidth: null,
+  selectedWidth: null,
   comparisons: [],
   buildWidths: [],
 
-  updateSelectedWidth() {},
-
-  selectedWidth: computed('snapshotSelectedWidth', {
```
cadeParade: Not sure why is like this?
Some performance gains maybe?
But it seems pretty fast without it to me...

fotinakis: Can't remember exactly — I think this might have to do with statefulness of the full screen mode. ie. you should be able to: 1) select mobile diff, 2) click full screen mode, 3) refresh the page and see the same thing.

fotinakis: Or this might just be cruft.

cadeParade: afaict, refreshing the page and seeing the same thing is related to the query parameters on the full screen route, not sure how this would affect that?
But let's QA this.

PR #388, opened 2018-01-25
Approval carryforward FE

cadeParade:
per-snapshot-approval


app/components/snapshot-viewer.js
```diff
@@ -1,11 +1,13 @@
 import {not, alias, notEmpty} from '@ember/object/computed';
 import {computed} from '@ember/object';
 import Component from '@ember/component';
+import {inject as service} from '@ember/service';
 
 export default Component.extend({
+  snapshot: null,
+  store: service(),
```
cadeParade: do we need this?

app/components/approval-button.js
```diff
@@ -11,24 +9,12 @@ export default Component.extend({
...
-          this.set('isLoading', false);
-        });
 
-      this.get('approve')();
-    },
+  click() {
+    this.set('isLoading', true);
+    const snapshotIds = this.get('build.snapshots').mapBy('id');
+    this.createReview(this.get('build.id'), snapshotIds).finally(() => {
+      this.set('isLoading', false);
```
cadeParade: Put `reloadAll` back in here

app/routes/organization/project/builds/build.js
```diff
@@ -68,5 +68,22 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
+        return this.get('store').peekRecord('snapshot', snapshotId);
+      });
+
+      const review = this.get('store').createRecord('review', {
+        build: build,
+        snapshots: snapshots,
+      });
+      return review.save().then(() => {
+        const build = this.modelFor(this.routeName);
+        build.get('comparisons').reload();
```
cadeParade: will be snapshots after refactor

cadeParade: For my future reference: This is the line that triggered the snapshots endpoint and refactoring the models to depend on snapshots rather than comparisons

app/components/snapshot-list.js
```diff
@@ -87,12 +87,22 @@ export default Component.extend({
     'hideNoDiffs',
     'snapshotsWithDiffs.[]',
     'snapshotsWithoutDiffs.[]',
+    // This computed property is purposefully not depending on snapshot.isApproved.
```
cadeParade: This strategy worked when we were not reloading comparisons after the review object and setting the property locally. Now that we make an actual server request to the backend, it gets updated comparisons into the store and reruns all dependent computed properties. Unclear whether that will happen when reloading snapshots instead of comparisons here: https://github.com/percy/percy-web/pull/388/files#diff-0d8081bc82d75c66beb696b6a8819d64R84

cadeParade: It does not work when reloading snapshots instead of comparisons, so we need another strategy.

PixelJanitor:
- [ ] Expand all snapshots when entire build is approved

tests/integration/components/snapshot-list-test.js
```diff
@@ -0,0 +1,68 @@
...
+    expect(titlesBeforeApproval[0]).to.equal(unapprovedSnapshotTitle);
+    expect(titlesBeforeApproval[1]).to.equal(approvedSnapshotTitle);
+
+    unapprovedSnapshot.set('isApproved', true);
+
+    return wait().then(() => {
+      const titlesAfterApproval = SnapshotListPageObject.snapshotTitles;
+      expect(titlesAfterApproval).to.eql(titlesBeforeApproval);
+    });
+  });
```
cadeParade: Write a test for the thing `hasHideNoDiffsChanged` does

tests/integration/components/snapshot-viewer-full-test.js
```diff
@@ -150,3 +156,74 @@ describe('Integration: SnapshotViewerFull', function() {
     });
   });
 });
+
+describe('Integration: SnapshotViewerFull with per snapshot approval', function() {
```
cadeParade: Copy-pasta of whole test setup so we can test in admin-mode. Will be removed when feature is no longer behind flag.

tests/integration/components/snapshot-viewer-test.js
```diff
@@ -136,4 +142,91 @@ describe('Integration: SnapshotViewer', function() {
...
+      this.set('isDefaultExpanded', false);
+
+      return wait(() => {
+        expect(SnapshotViewerPO.isExpanded).to.equal(false);
+      });
+    });
+  });
+});
+
+describe('Integration: SnapshotViewer with per snapshot approval', function() {
```
cadeParade: Copy-pasta of whole test setup so we can test in admin-mode

tests/integration/components/snapshot-list-test.js
cadeParade: Done

app/components/snapshot-approval-button.js
```diff
@@ -0,0 +1,22 @@
...
+  isApproved: alias('snapshot.isApproved'),
+  isLoading: false,
+  tagName: '',
+
+  actions: {
+    approveSnapshot() {
+      this.set('isLoading', true);
+      this.createReview([this.get('snapshot.id')])
+        .then(() => {
+          this.set('snapshot.reviewState', 'approved');
```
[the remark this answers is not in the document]
fotinakis: Optionally, I think if you're doing this for speed reasons it should always go 1) set temp state 2) make call to backend (instead of having it chained on the promise at all). So should be above the `this.createReview` call, since you may not want it to wait on the save() promise resolving first. There might even be a race condition there where ember-data state is updated from the API to be something else, and then these get called and override the state improperly.

app/components/snapshot-list.js
```diff
@@ -83,16 +86,43 @@ export default Component.extend({
       return total + snapshot.get('comparisons.length');
     }, 0);
   }),
+
+  // This is a bit jank but works for now...
```
fotinakis: Can you clean up this comment a bit? Partially it's just the formatting (ie. full sentences + make the lines up to 100 characters long), partially it's very confusingly worded of WHY this is doing what it's doing. Maybe this belongs in the service itself?

app/components/snapshot-list.js
```diff
@@ -83,16 +86,43 @@ export default Component.extend({
...
-    'snapshotsWithoutDiffs.[]',
+    'snapshotsWithDiffs.@each.isApproved',
+    'snapshotsWithoutDiffs.@each.isApproved',
     function() {
-      if (this.get('hideNoDiffs')) {
-        return this.get('snapshotsWithDiffs');
-      } else {
-        return [].concat(this.get('snapshotsWithDiffs'), this.get('snapshotsWithoutDiffs'));
+      if (
+        this.get('cachedSnapshotOrder.orderedSnapshots') &&
```
fotinakis: Maybe have a method on the service that wraps this condition to answer this question?

app/components/snapshot-list.js
```diff
@@ -83,16 +86,43 @@ export default Component.extend({
...
+        !this.get('cachedSnapshotOrder.hasHideNoDiffsChanged')
+      ) {
+        return this.get('cachedSnapshotOrder.orderedSnapshots');
       }
+
+      let snapshots = this.get('hideNoDiffs')
+        ? this.get('snapshotsWithDiffs')
+        : [].concat(this.get('snapshotsWithDiffs'), this.get('snapshotsWithoutDiffs'));
+      const approvedSnapshots = [];
```
fotinakis: (nitpick) Mixing `let` and `const` is confusing a bit, though I know they're slightly different.

app/components/snapshot-list.js
```diff
@@ -83,16 +86,43 @@ export default Component.extend({
...
+      const unapprovedSnapshots = [];
+      snapshots.forEach(snapshot => {
+        snapshot.get('isApproved')
```
fotinakis: Can you get this on one line? If a ternary spans multiple lines then I think we should just write it as an if statement for clarity — this spread across multiple lines is definitely not easier to read than an if/else.

app/components/snapshot-list.js
```diff
@@ -206,6 +236,7 @@ export default Component.extend({
     },
     toggleNoDiffSnapshots() {
       this.toggleProperty('hideNoDiffs');
+      this.set('cachedSnapshotOrder.hasHideNoDiffsChanged', true);
```
fotinakis: Shouldn't this be a service method, instead of reaching into the service and modifying internal data? It's not just a data struct, right?

app/models/snapshot.js
```diff
@@ -1,6 +1,9 @@
 import DS from 'ember-data';
 import {equal} from '@ember/object/computed';
 
+export const SNAPSHOT_APPROVED_STATE = 'approved';
```
fotinakis: Maybe rename these to:

```
SNAPSHOT_REVIEW_STATE_APPROVED = 'approved'
SNAPSHOT_REVIEW_STATE_UNREVIEWED = 'unreviewed'
```

Two changes there: calling it "review state" instead of just "state" (since we use that elsewhere and it's different), and fixing typo "unapproved" --> "unreviewed".

app/routes/organization/project/builds/build.js
```diff
@@ -68,5 +73,22 @@ export default Route.extend(AuthenticatedRouteMixin, {
         queryParams: {activeSnapshotId: snapshotId},
       });
     },
+
+    createReview(buildId, snapshotIds) {
```
fotinakis: Thought for the future: I think we should should put business logic like this in services, and then call it from this route action. Would be much easier to test in isolation, and routes can be much smaller and targeted.

fotinakis: Two changes here I'd like to request here:

1) I think this should also accept `action` as the first param, to match the backend method and be future proof.

```js
createReview(action, build, snapshots)
```

2) I think we should prefer passing around full objects instead of Ids, 99% of the time (this is mostly true across Percy right now, except in some serialization cases). This would also help you avoid all the peekRecord calls inside this method.

app/templates/components/snapshot-approval-button.hbs
```diff
@@ -0,0 +1,5 @@
+{{#if (eq snapshot.reviewState "approved")}}
```
fotinakis: Just use snapshot.isApproved?

app/templates/components/snapshot-viewer-header.hbs
```diff
@@ -56,6 +56,16 @@
           </figure>
         </button>
       </div>
+
+      {{#if (is-admin)}}
```
fotinakis: Nice! 💯 :)

tests/integration/components/snapshot-approval-button-test.js
```diff
@@ -0,0 +1,68 @@
...
+/* eslint-disable no-unused-expressions */
+import {setupComponentTest} from 'ember-mocha';
+import {expect} from 'chai';
+import {it, describe, beforeEach} from 'mocha';
+import {percySnapshot} from 'ember-percy';
+import hbs from 'htmlbars-inline-precompile';
+import {make, manualSetup} from 'ember-data-factory-guy';
+import sinon from 'sinon';
+import {resolve, defer} from 'rsvp';
+import SABPO from 'percy-web/tests/pages/components/snapshot-approval-button';
```
fotinakis: Style: prefer non-acronyms, almost always. How about `snapshotApproveButton` or just `pageObject`?

tests/integration/components/snapshot-viewer-full-test.js
fotinakis: Optionally, I don't believe the snapshot component does anything different based on admin mode right now so if we could drop the copy/paste here and just set admin-mode with a TODO in the normal setup that'd be better.

app/models/snapshot.js
cadeParade: I'm ok with updating the variables to have `STATE` in them, but I don't really know that changing from `SNAPSHOT_REVIEW_STATE_UNAPPROVED` to SNAPSHOT_REVIEW_STATE_UNREVIEWED` is necessary. 
Using these constants it's not necessary for the user to know the string behind it, and it makes more sense to use the pair `APPROVED`/`UNAPPROVED` than the pair `APPROVED`/`UNREVIEWED`.

app/routes/organization/project/builds/build.js
cadeParade: I like it, although untraditional for Ember apps afaik

tests/integration/components/snapshot-approval-button-test.js
cadeParade: ok

app/components/snapshot-approval-button.js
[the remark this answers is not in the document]
cadeParade: Yes, you're right, upon further reflection I think this is leftover from a previous iteration. I will remove.

app/templates/components/snapshot-approval-button.hbs
cadeParade: 💯

tests/integration/components/snapshot-viewer-full-test.js
cadeParade: I mostly think that's right, but since we're planning on releasing this before the feature flag is off, we need both states to work. And we need the old state to work more than the new state, since that's what customers are using. So this is a super defensive test, yes.

fotinakis: 👍

app/models/snapshot.js
fotinakis: Yeah definitely agree that constants don't have to match and can be much more semantic, but counter argument is that we should balance with consistency so that when we say one thing somewhere it means the same thing other places. Specifically, "unapproved" is not technically a review state we have right now but "unreviewed" is, so adding _UNAPPROVED adds confusion rather than clarity IMO. But only minor minor confusion and I 💟  this change anyway so feel free to ignore this! :)

PR #390, opened 2018-01-26
Convert snapshots to `hasMany` relationship on build

cadeParade:
With the new `snapshots` api, we can load snapshots for a build and single snapshots rather than loading comparisons first and deriving snapshots. This PR converts the app from a `comparisons`-first model to a `snapshots`-first model. 

Comparisons on the build model is converted to a synchronous relationships which means that the app will never ask for anything at the comparisons endpoint, instead assuming that any comparisons we're looking for will have already been loaded. This is the case, as the snapshots api does sideload all related comparison data.
 
~The tests for this PR are failing since the api endpoint has changed, but the fixtures have not. The PR for fixing the tests is tests https://github.com/percy/percy-web/pull/403 and should be merged into this PR before this merges to master.~


app/models/build.js
```diff
@@ -97,12 +97,12 @@ export default DS.Model.extend({
...
   commit: DS.belongsTo('commit', {async: false}), // Might be null.
   baseBuild: DS.belongsTo('build', {async: false, inverse: null}),
-  comparisons: DS.hasMany('comparison', {async: true}),
+  snapshots: DS.hasMany('snapshot', {async: true}),
 
-  snapshots: computed('comparisons', function() {
-    let comparisons = this.get('comparisons');
-    let snapshots = comparisons.map(comparison => comparison.get('headSnapshot')).filter(x => x);
-    return [...new Set(snapshots)];
+  comparisons: computed('snapshots', function() {
```
fotinakis: We should be able to entirely let ember-data handle this from the sideloading, right? Just:

`comparison: DS.hasMany('comparison', {async: false})`

app/components/snapshot-list.js
```diff
@@ -16,9 +16,10 @@ export default Component.extend({
   snapshotComponents: null,
   updateActiveSnapshotId: null,
   updateSelectedWidth: null,
+  snapshots: null,
 
   sortedSnapshots: computed('snapshots.[]', 'buildContainerSelectedWidth', function() {
-    let snapshots = this.get('snapshots');
+    let snapshots = this.get('snapshots').toArray();
```
fotinakis: Why?

timhaines:
Have read through this.  Nothing to add to Mike's comments.

Noticed it needs to be rebased for more up-to-date Percy snapshots, those 8 diffs don't look right.  (probably from what was merged last night..)

timhaines:
Oh - I should mention it's a pretty exciting clean-up though.  Logically it makes much more sense and is easier to grok now. :)

app/models/build.js
cadeParade: Well, that ends up not working -- 
With this update we don't really want `build` to know about its comparisons, we just want it to know about the snapshots which can then handle their own comparisons.
But the build still has to know about comparisons _only_ for `comparisonWidths` which is _only_ used in the `build-width-selector`, which we are [getting rid of](https://github.com/percy/percy-web/pull/408/files#diff-2e4ba0bacc6232c65a0c6c6de84fb212L108)!

But until that is gone, we need to collect all the comparisons for the build to calculate comparison widths 😢

app/components/snapshot-list.js
cadeParade: `snapshots` is a `PromiseManyArray` which doesn't have the `sort` method called on it below.

cadeParade: And before since it was a computed property based on the `comparisons` `PromiseManyArray`, it was...something else that did have a sort method.

app/models/build.js
fotinakis: 👍 👍 👍

PR #396, opened 2018-01-31
Copy snapshot url link in snapshot header

PixelJanitor:


app/components/snapshot-viewer.js
```diff
@@ -73,6 +72,7 @@ export default Component.extend({
     selectChild() {
       this.get('setAsSelected').call(this);
       this.get('selectChild')(this);
+      this.get('flashMessages').success('Snapshot URL was copied to your clipboard');
```
cadeParade: Get rid of this line to make the tests pass

app/components/snapshot-viewer.js
```diff
@@ -3,9 +3,8 @@ import {computed} from '@ember/object';
 import Component from '@ember/component';
 
 export default Component.extend({
-  classNames: ['SnapshotViewer'],
-
   snapshot: null,
+  classNames: ['SnapshotViewer'],
```
cadeParade: P4: I usually put the class things above all the other things in a component

app/helpers/generate-fullscreen-snapshot-url.js
```diff
@@ -0,0 +1,13 @@
+import {helper} from '@ember/component/helper';
+
+export default helper(([snapshotId, selectedWidth, comparisonMode, isFullscreen]) => {
+  let location = window.location;
+  comparisonMode = comparisonMode || 'diff';
+  if (isFullscreen) {
+    return location.href;
+  } else {
+    return `${location.origin}${
```
cadeParade: lol ok prettier

app/components/snapshot-viewer.js
cadeParade: Oh, also this is what is causing the flash message to appear whenever you click anywhere on the snapshot.

cadeParade:
Ok, so I was looking into this...the flash message is not appearing on the full screen view because of Z-INDEX.
Full screen view is in a modal, and that modal is on top of the flash message div. If you watch the "elements" panel in chrome inspector you can see that a flash message container does appear when you click the button, it is just hidden.

PixelJanitor:
@cadeParade I made a test adjustment, but there are some additional failures I think you should look at. Otherwise, this is good to go! Let me know if I can help with anything else here.

app/components/snapshot-viewer.js
cadeParade: @PixelJanitor this is the line that was causing the tests to fail. I've removed it and now the tests should pass.

PR #403, opened 2018-02-03
Refactor acceptance tests for build page

cadeParade:
The build page tests were old and cryptic. Additionally, updating the model loading from comparison-first to snapshot-first made a bunch of the build page tests fail. 

I've updated the tests by:
- Making mirage factories for snapshots, comparisons, screenshots, and images that all work together with traits, so they can be built together and easily edited.
- Making page objects for the pieces of the build page and project page
- Updating build-test file to 1)test only things related to the build page, 2)move tests related to other things (Ex: tests for the projects page are now in `project-test`, a few acceptance tests are now in component integration tests because they didn't need the whole app to run to test, etc), 
- Make only test-models that are necessary (e.g. no more longer/shorter diffs)


tests/pages/components/snapshot-viewer-full.js
```diff
@@ -1,18 +1,48 @@
...
+const SELECTORS = {
+  SNAPSHOT_VIEWER_FULL: '[data-test-snapshotViewerFull]',
+  HEADER: '[data-test-SnapshotViewer-header]',
+  COMPARISON: '[data-test-SnapshotViewer-comparison]',
+  NO_COMPARISON: '[data-test-SnapshotViewer-noComparison]',
   COMPARISON_VIEWER: '[data-test-SnapshotViewerFull-comparison-viewer]',
-  COMPARISON_IMAGE: '[data-test-SnapshotViewerFull-comparison-viewer] [data-test-simple-image]',
+  COMPARISON_IMAGE: '[data-test-SnapshotViewerFull-comparison-viewer] img',
+  COMPARISON_MODE_SWITCHER: '[data-test-comparison-mode-switcher]',
+  DIFF_IMAGE: '.pdiffImageOverlay img',
```
cadeParade: change this to data-test selector

tests/pages/build-page.js
```diff
@@ -0,0 +1,77 @@
...
+  BUILD_LIST: '[data-test-project-container-build-list]',
+  NO_DIFFS_PANEL: '[data-test-SnapshotListToggleNoDiffs]',
+  SNAPSHOT_LIST: '[data-test-snapshotList]',
+  BUILD_INFO_DROPDOWN_TOGGLE: '[data-test-build-info-dropdown-toggle]',
+  SHOW_SUPPORT_LINK: '[data-test-build-overview-show-support]',
+};
+
+// TODO snapshot-list page object?
+const BuildPage = {
+  visitProject: visitable('/:orgSlug/:projectSlug'),
```
cadeParade: Remove this line

mirage/factories/image.js
```diff
@@ -0,0 +1,93 @@
+import {Factory, trait} from 'ember-cli-mirage';
```
fotinakis: Biggest comment here is that everywhere else in Percy, the Image model is designed to be super generic and not know how or where it is used —  it only holds very generic properties, like width, height, and mimetype. It feels like most everything below belongs at a bunch higher level, either at the Screenshot or Comparison level. ie. a "mobile image" is way too specific for percy's current knowledge of an image, but a "mobile screenshot" or a "mobile comparison" might make sense.

mirage/factories/screenshot.js
```diff
@@ -1,22 +1,50 @@
-import {Factory} from 'ember-cli-mirage';
+import {Factory, trait} from 'ember-cli-mirage';
```
fotinakis: Same sort of comment here — Screenshots are currently designed to be only 1 step more specific than Image. So a Screenshot has things like image, lossy_image, etc. but doesn't actually know how or where it is used otherwise, so we probably shouldn't see things like "head" or "base" at this level because those always mean that something is being used as the head or base of a comparison.

It WOULD make sense to turn these into specific types of screenshots, like a `screenshot` with a `longDesktop` or `longMobile` or `shortMobile` traits, that specify something about what the screenshot looks like.

app/components/snapshot-viewer.js
```diff
@@ -18,7 +21,7 @@ export default Component.extend({
       return value;
     },
   }),
-  selectedComparison: computed('snapshot', 'snapshotSelectedWidth', function() {
+  selectedComparison: computed('snapshot.@each.comparisons', 'snapshotSelectedWidth', function() {
```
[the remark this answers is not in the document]
cadeParade: yes

app/models/build.js
```diff
@@ -112,7 +112,7 @@ export default DS.Model.extend({
     }, []);
   }),
 
-  comparisonWidths: computed('comparisons', function() {
+  comparisonWidths: computed('comparisons.@each.width', function() {
```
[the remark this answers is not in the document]
cadeParade: Why would it _not_ depend on comparison widths? The property is getting each comparison's width?

app/templates/components/build-card.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="build-card c-pointer {{if build.isExpired 'opacity-50'}} is-{{dasherize build.state}} is-{{if build.isFinished (dasherize build.reviewState)}}" {{action "navigateToBuild"}}>
+<div class="build-card c-pointer {{if build.isExpired 'opacity-50'}} is-{{dasherize build.state}} is-{{if build.isFinished (dasherize build.reviewState)}}" {{action "navigateToBuild"}} data-test-build-card>
```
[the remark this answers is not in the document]
cadeParade: This component has `tagName: ''`
But I have a better idea for how to handle `data-test` attributes in the future, I just had to draw the line for this PR somewhere.

mirage/factories/image.js
cadeParade: @fotinakis check out https://github.com/percy/percy-web/pull/403/commits/9a0d0a0281a7f7f06ebf8c3971a924c6fd6ce78f
I've made an attempt to separate things here....lmk what you think

tests/integration/components/snapshot-viewer-full-test.js
```diff
@@ -1,13 +1,13 @@
...
 import {setupComponentTest} from 'ember-mocha';
 import {expect} from 'chai';
 import {it, describe, beforeEach} from 'mocha';
 import {percySnapshot} from 'ember-percy';
 import hbs from 'htmlbars-inline-precompile';
-import {make, makeList, manualSetup} from 'ember-data-factory-guy';
+import {make, manualSetup} from 'ember-data-factory-guy';
 import sinon from 'sinon';
-import SnapshotViewerPO from 'percy-web/tests/pages/components/snapshot-viewer';
+import SnapshotViewerFullPO from 'percy-web/tests/pages/components/snapshot-viewer-full';
```
fotinakis: Per discussion, drop the `PO` or `PageObject` naming here and everywhere.

mirage/factories/screenshot.js
```diff
@@ -1,22 +1,43 @@
...
-        width: 900,
-        height: 1044,
-      });
+      let lossyImage = server.create('image', 'lossy');
       screenshot.update({lossyImage});
     }
   },
+
+  base: trait({
+    afterCreate(screenshot, server) {
```
fotinakis: I think you can just override the associations here, you don't have to have an `afterCreate`.

mirage/factories/image.js
fotinakis: Per discussion, we could move all this stuff up into screenshot factory because the screenshots should have the responsibility to know what context images are used in (and things like diff_image make way more sense there).

mirage/factories/image.js
```diff
@@ -0,0 +1,39 @@
...
+  DEFAULT_HEIGHT: 1485,
+  DEFAULT_WIDTH: 1280,
+  LOSSY_HEIGHT: 1044,
+  LOSSY_WIDTH: 900,
+  MOBILE_HEIGHT: 1598,
+  MOBILE_WIDTH: 320,
+};
+
+export const TEST_IMAGE_URLS = {
+  DEFAULT_HEAD_URL: '/images/test/bs-head.png',
```
fotinakis: Per discussion:

`base` --> `v1`
`head` --> `v2`

for traits and image files.

mirage/factories/screenshot.js
fotinakis: Per discussion, mirage models/traits be wonky with multiple inverse associations. We'll figure this out later.

app/components/snapshot-viewer.js
```diff
@@ -37,17 +37,22 @@ export default Component.extend({
...
-    let comparison = comparisons.findBy('width', parseInt(width, 10));
-
-    if (!comparison) {
-      comparison = this.get('comparisonsSortedByWidth').get('lastObject');
-    }
-
-    return comparison;
-  }),
+  selectedComparison: computed(
+    'comparisons.@each.width',
```
cadeParade: Because prettier

PR #405, opened 2018-02-06
Admin mode as util function instead of Ember service

cadeParade:
The `admin-mode` interface was a bit confusing, and also didn't really take advantage of Ember service properties since `localStorage` doesn't really integrate with Ember well. 
It also uses the `localStorageProxy` which is a safer interface to `localStorage` although theoretically percy employees are the only ones who should be using this)

This PR takes the `admin-mode` functionality and converts it to a POJO with some methods that can be called if we need to check something related to admin `localStorage`.


timhaines:
LGTM too.  :)

PR #406, opened 2018-02-06
Snapshot width selector with dropdown

PixelJanitor:


PixelJanitor:
- [ ] No snapshot taken at px.
- [ ] Cannot select different widths
- [ ] Largest width selected by default if there are no diffs

cadeParade:
@PixelJanitor I've made some updates to the logic, I think things work now. 
I'll need your help outlining the tests, I'm not sure all the things the dropdown should and shouldn't do

PixelJanitor:
Ok awesome! Let’s chat about it sometime tomorrow when you’re available.

timhaines:
image

@PixelJanitor Do you think we should hide the dropdown for the previous snapshot:
* When a second snapshot's dropdown is opened
* When the user scrolls or stops scrolling (maybe?)

timhaines:
There's some unexpected dropshadow thinkness changes.  For example, in the 'Acceptance: Pending Build shows as pending on the project page' screenshot.

timhaines:
@PixelJanitor The Percy snapshot titled `Integration: SnapshotViewer with per snapshot approval approve snapshot button sends createReview with correct arguments when approve button is clicked` appears broken.

timhaines:
I’ve been playing around with this locally. Feels pretty good and ready to ship from a UI perspective.  I left one comment above about a broken snapshot.  I *haven’t* read through the code, but +1 from a manual QA perspective.

PixelJanitor:
@timhaines I also saw that yesterday and asked @cadeParade if this is the expected snapshot based on all of the new tests that were added with this PR. She confirmed that this test is primarily testing functionality and the snapshot taken in this test is the one she expected to see.

timhaines:
@PixelJanitor @cadeParade I don't think we should include a snapshot that has broken text (no number before px in `no snapshot taken at px`) and is missing width selectors.  Gives the impression to anyone looking that something is broken for this scenario.

PixelJanitor:
I'm also +1 for removing the snapshot part of this test

app/templates/components/snapshot-viewer-header.hbs
```diff
@@ -22,21 +21,43 @@
...
-          title="Copy snapshot URL to clipboard"
-          success=(action "onCopySnapshotUrlToClipboard")
-        }}
-          {{inline-svg "link-icon" class="mr-sm"}}
-        {{/copy-button}}
+        <div class="dropdown {{if dropdownVisible 'is-visible'}}" {{action 'closeDropdown' bubbles=false}}>
+          <button type="button" class="btn btn-square dropdown-toggle" {{action 'toggleDropdownVisibility' bubbles=false}} data-test-snapshot-header-dropdown-toggle>
+            {{inline-svg "blue-dropdown-icon"}}
+          </button>
+          <div class="dropdown-menu" data-test-snapshot-header-dropdown-pane>
```
cadeParade: This dropdown menu should become a component sometime

PR #408, opened 2018-02-08
Remove tendrils of build width selector, PLUS snapshot sort algo tests

cadeParade:
This PR removes the remaining tendrils of the build width selector, slightly refactors the snapshot sort method, and finally adds some TESTS for the snapshot sort method, which is very exciting.
It is also on top of the comparison -> snapshot model refactor, so some nice things could happen with the snapshot model.

This PR should be the death of the `comparisonWidths` property on the build model which will simplify a bunch of stuff going forward.

### DIFF
--- a/app/lib/snapshot-sort.js
+++ b/app/lib/snapshot-sort.js
@@ -1,0 +1,59 @@
+export default function(snapshots) {
+  let width = _maxWidthForSnapshots(snapshots);
+
+  function comparisonAtCurrentWidth(snapshot) {
+    return snapshot.get('comparisons').findBy('width', width);
+  }
+
+  return snapshots.sort(function(a, b) {
+    // Prioritize snapshots with diffs at any widths over snapshots with no diffs at any widths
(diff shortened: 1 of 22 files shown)


app/models/build.js
```diff
@@ -110,16 +110,6 @@ export default DS.Model.extend({
     }, []);
   }),
 
-  comparisonWidths: computed('comparisons.@each.width', function() {
```
cadeParade: GOODBYE

app/lib/snapshot-sort.js
```diff
@@ -0,0 +1,55 @@
+export default function(snapshots) {
```
djones: 👍 for moving this fairly complex sorting algorithm out into another file

tests/unit/lib/snapshot-sort-test.js
```diff
@@ -0,0 +1,67 @@
+import {expect} from 'chai';
```
djones: Thanks for adding tests on this!

PR #410, opened 2018-02-10
Fix 2 bugs with per-snapshot approval

cadeParade:
1. Do not allow a snapshot to be approved when the build is not finished. This is accomplished by hiding the button until the build is finished. This will prevent this error (I think): https://sentry.io/percy/percy-prod/issues/458599740/ 

2. Do not access the cached snapshot order until the build is finished. This fixes the first item in https://app.clubhouse.io/percy/story/1574/snapshot-approval-qa

PLUS TESTS!!


app/templates/components/snapshot-viewer-header.hbs
```diff
@@ -57,8 +57,8 @@
         </button>
       </div>
 
-      {{#if (is-admin)}}
-        <div class="btn-group">
+      {{#if (and (is-admin) isBuildFinished)}}
```
timhaines: are the () still needed for is-admin because it’s hyphenated?

tests/integration/components/snapshot-list-test.js
```diff
@@ -8,42 +8,39 @@ import {make, makeList, manualSetup} from 'ember-data-factory-guy';
...
-  const unapprovedSnapshotTitle = 'Unapproved snapshot!!';
-  let approvedSnapshot;
-  let unapprovedSnapshot;
-
   beforeEach(function() {
     manualSetup(this.container);
     SnapshotListPageObject.setContext(this);
+  });
 
+  it('displays snapshots in the correct order, before and after approval when build is finished', function() { // eslint-disable-line
```
timhaines: eslint disabled due to line length?

app/templates/components/snapshot-viewer-header.hbs
cadeParade: It's because `is-admin` is a template helper, it's not a component property (as is `isBuildFinished`).
Template helpers are called with surrounding parenthesis in handlebars.

tests/integration/components/snapshot-list-test.js
cadeParade: yes, due to line length.

app/templates/components/snapshot-viewer-header.hbs
cadeParade: Also, you can't really have a hyphenated component property. It will definitely error. And template helpers are always hyphenated. So that's one way to tell the difference.

PR #414, opened 2018-02-12
Adjust keyboard scrolling on snapshot list

PixelJanitor:
Fixes https://app.clubhouse.io/percy/story/1572/snapshot-arrow-key-scroll-queuing-is-super-slow


cadeParade:
Can you explain a little how this change fixes the problem in the ticket?

PixelJanitor:
The debounce didn’t work, but bringing the animated scroll speed down to 100 made it much less likely to get hung up.
		
		

		Get Outlook for iOS
	




On Mon, Feb 12, 2018 at 8:30 PM -0500, "Lindsay" <notifications@github.com> wrote:












Can you explain a little how this change fixes the problem in the ticket?

PixelJanitor:
switching this to `window.scrollTo()` instead of `Element.scrollTop()` since its 93% faster in performance: https://jsperf.com/window-scrollto-vs-jquery-scrolltop.

I've put it back to a non-scrollling animation as well. This should be good to go

PR #415, opened 2018-02-12
"No changes" label

PixelJanitor:
No changes label for snapshots/builds that are auto approved

Build page:
image

Snapshot list:
image


app/templates/components/snapshot-approval-button.hbs
```diff
@@ -1,5 +1,11 @@
 {{#if snapshot.isApproved}}
-  <div class="status-pill is-approved" data-test-snapshot-approved-pill>Approved</div>
+  <div class="status-pill is-approved" data-test-snapshot-approved-pill>
+    {{#if (eq snapshot.reviewStateReason 'no_diffs')}}
+      No Changes
```
fotinakis: @PixelJanitor shouldn't you just use buildStatusLabel here instead of duplicating this logic?

PixelJanitor: @fotinakis Because we're on the snapshot level and the buildStatusLabel is a status for the entire build.

Build status label returns this:
image

PR #417, opened 2018-02-13
Add basic GitHub enterprise integration UI

faun:
Spike to get familiar with and surface GitHub enterprise integration status.

https://app.clubhouse.io/percy/story/1492/surface-ghe-integration-details-on-the-organization-settings-page-once-the-integration-is-manually-installed

settings_-_percy___continuous_visual_reviews_for_web_apps


app/components/organizations/github-enterprise-settings.js
```diff
@@ -0,0 +1,24 @@
...
+import Component from '@ember/component';
+
+export default Component.extend({
+  organization: null,
+
+  store: service(),
+  session: service(),
+  currentUser: alias('session.currentUser'),
+  versionControlIntegrations: alias('organization.versionControlIntegrations'),
+  githubEnterpriseIntegration: computed('organization', function() {
```
cadeParade: This should depend on `organization.githubEnterpriseIntegration` since that's the thing you're using in the computed property.

But also, couldn't it be
`githubEnterpriseIntegration: alias('organization.githubEnterpriseIntegration')`?

app/components/organizations/github-enterprise-settings.js
```diff
@@ -0,0 +1,24 @@
...
+    let organization = this.get('organization');
+    if (organization) {
+      let githubEnterpriseIntegration = organization.get('githubEnterpriseIntegration');
+      return githubEnterpriseIntegration;
+    }
+  }),
+
+  githubEnterpriseIntegrationInstalled: computed('organization', function() {
```
cadeParade: This could be `alias('organization.isGithubEnterpriseIntegrated')`?

Also, it looks like this property returns a boolean which we usually name with `is` in front.

app/models/organization.js
```diff
@@ -32,6 +41,14 @@ export default DS.Model.extend({
   isGithubIntegrated: computed('githubAuthMechanism', function() {
     return this.get('githubAuthMechanism') !== 'no-access';
   }),
+
+  isGithubEntrerpriseIntegrated: computed('githubEnterpriseIntegration', function() {
+    if (this.get('githubEnterpriseIntegration')) {
```
cadeParade: `return !!this.get('githubEnterpriseIntegration');` ?

app/models/version-control-integration.js
```diff
@@ -1,13 +1,24 @@
...
   createdAt: DS.attr('date'),
   updatedAt: DS.attr('date'),
 
-  isGithubIntegration: notEmpty('githubInstallationId'),
+  isGithubIntegration: computed('integrationType', function() {
+    return this.get('integrationType') === 'github';
+  }),
+
+  isGithubEnterpriseIntegration: computed('integrationType', function() {
+    return this.get('integrationType') == 'github_enterprise';
```
cadeParade: double equals?

Also, I propose making `'github'` and `'github_enterprise'` strings constants at the top of the file in case we need to use these types anywhere else.

app/templates/components/organizations/github-enterprise-settings.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if (not githubEnterpriseIntegrationInstalled)}}
+  <div class="mt-4">
+    To install the GitHub Enterprise integration, please contact sales.
```
cadeParade: Probs we should massage the wording in this file.

tests/factories/organization.js
```diff
@@ -13,11 +13,22 @@ FactoryGuy.define('organization', {
...
     withGithubIntegration: {
       versionControlIntegrations: () => {
-        // TODO: This will clobber other versionControlIntegrations
-        // When more integrations are added here, fix this to support them
         return [make('version-control-integration', 'github')];
       },
     },
+    withGithubEnterpriseIntegration: {
+      versionControlIntegrations: () => {
+        return [make('version-control-integration', 'githubEnterprise')];
```
cadeParade: There is a `makeList` method on FactoryGuy which may be useful here 👍

app/models/organization.js
```diff
@@ -32,6 +41,11 @@ export default DS.Model.extend({
   isGithubIntegrated: computed('githubAuthMechanism', function() {
     return this.get('githubAuthMechanism') !== 'no-access';
   }),
+
+  isGithubEnterpriseIntegrated: computed('githubEnterpriseIntegration', function() {
+    return !!this.get('githubEnterpriseIntegration');
```
cadeParade: Would `isGithubEnterpriseIntegrated: bool('githubEnterpriseIntegration')` would work here?

https://www.emberjs.com/api/ember/2.16/classes/@ember%2Fobject%2Fcomputed/methods/bool?anchor=bool

The `computed` section of this page is super useful to know what all the computed macros are: https://www.emberjs.com/api/ember/2.16/modules/@ember%2Fobject
I have it saved to my bookmarks bar because it is hard to find organically 🙄

app/models/version-control-integration.js
```diff
@@ -1,13 +1,27 @@
+import {computed} from '@ember/object';
 import DS from 'ember-data';
-import {notEmpty} from '@ember/object/computed';
+
+const githubEnterpriseIntegrationType = 'github_enterprise';
```
cadeParade: This type of string constant is usually in ALL_CAPS to notate that it is special and should never change.

app/models/version-control-integration.js
```diff
@@ -1,13 +1,27 @@
...
+  githubEnterpriseHost: DS.attr(),
+  githubEnterpriseInstallationId: DS.attr(),
+  githubEnterpriseIntegrationId: DS.attr(),
+
   createdAt: DS.attr('date'),
   updatedAt: DS.attr('date'),
 
-  isGithubIntegration: notEmpty('githubInstallationId'),
+  isGithubIntegration: computed('integrationType', function() {
+    return this.get('integrationType') === githubIntegrationType;
```
cadeParade: This and the one below could use `equal` computed macro:
`isGithubIntegration: equal('integrationType', githubIntegrationType)`

One tricky thing about the `equal` macro is that the second argument must be literal rather than a reference to a property on the current object context. I think a variable should work here, but try it and if it doesn't, then we know the second argument must be really, really literal and not even a reference to a variable will work. Thanks Ember.

tests/integration/components/organizations/github-enterprise-settings-test.js
```diff
@@ -0,0 +1,59 @@
...
+  beforeEach(function() {
+    manualSetup(this.container);
+    GithubEnterpriseSettings.setContext(this);
+  });
+
+  describe('with a github enterprise integration', function() {
+    beforeEach(function() {
+      const user = make('user');
+      const organization = make('organization', 'withGithubEnterpriseIntegration');
+      user.organization = organization;
```
cadeParade: The user model shows the organization relationship being a `hasMany` so this should be `user.organizations = [organization]`;

tests/integration/components/organizations/github-enterprise-settings-test.js
```diff
@@ -0,0 +1,59 @@
...
+      this.set('user', user);
```
cadeParade: For future reference, to save typing, when you want to set multiple properties, you can use:
`this.setProperties({key1: value1, key2: value2})`

Additionally, if you want to super save typing, you can use the fancy es6 syntax for [property definitions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer) (grep for "Shorthand property names") which would end up like this:
`this.setProperties({user, organization})` 

chef kissy

tests/integration/components/organizations/github-enterprise-settings-test.js
```diff
@@ -0,0 +1,59 @@
...
+    it('shows that the integration is installed', function() {
+      const isIntegrated = this.get('organization.isGithubEnterpriseIntegrated');
+      expect(isIntegrated).to.equal(true);
+      this.render(hbs`{{
+        organizations/github-enterprise-settings
+        currentUser=user
+        organization=organization
+      }}`);
+      expect(GithubEnterpriseSettings.integrationMessage).to.include('is installed');
+      percySnapshot(this.test.fullTitle(), this.$());
```
cadeParade: The `this.$()` as the second argument is not really necessary... what will it do?

tests/pages/components/github-enterprise-settings.js
```diff
@@ -0,0 +1,11 @@
+import {create, text} from 'ember-cli-page-object';
+
+const SELECTORS = {
+  INTEGRATION_STATUS_MESSAGE: '[data-test-github-enterprise-settings]',
+};
+
+export const GithubEnterpriseSettings = {
```
cadeParade: Page objects should have a scope.

app/models/organization.js
faun: computed. 9b0dc87c.

tests/pages/components/github-enterprise-settings.js
faun: Fixed in 60345131

tests/integration/components/organizations/github-enterprise-settings-test.js
faun: Neat! c80aa621

tests/integration/components/organizations/github-enterprise-settings-test.js
faun: I fix in 6992b53e

faun: I fix in 6992b53e

tests/integration/components/organizations/github-enterprise-settings-test.js
faun: Per discussion IRL, it scopes the selector for the snapshot. It is unnecessary and has been removed in c50d7899

app/models/version-control-integration.js
faun: ¯\\\_(ツ)\_/¯ Seems to work fine in 8e07b9d3

app/components/organizations/github-enterprise-settings.js
faun: Fixed in af46bc9f1b6afd5a0292a408cd528737dc4ca0e7

app/models/organization.js
faun: Ultimately fixed in 9b0dc87c4a237664bbbd3e8ab1e0ebc2ff3f792c

app/templates/components/organizations/github-enterprise-settings.hbs
faun: I reached out to @timhaines about this.

app/models/version-control-integration.js
faun: Fixed in 8e07b9d38c7269cd8dfdb3518e4da3356e32f131

faun:
This is ready for the @PixelJanitor golden touch.

cadeParade:
@PixelJanitor 's golden touch

faun:
🤣

faun:
This is essentially feature flagged at this point, since it shows nothing if you don't have the integration installed.

PR #421, opened 2018-02-14
Build header update

PixelJanitor:


PixelJanitor:
- [ ] Pending states
- [ ] Comparison Icon component that uses build state

app/models/build.js
```diff
@@ -68,6 +68,10 @@ export default DS.Model.extend({
   // Each comparison represents a single individual rendering at a width and browser.
   totalSnapshots: DS.attr('number'),
   totalSnapshotsUnreviewed: DS.attr('number'),
+  totalSnapshotsApprovedByUser: computed('totalSnapshots', 'totalSnapshotsUnreviewed', function() {
```
fotinakis: Unused, drop this?

app/templates/components/build-overview.hbs
```diff
@@ -1,68 +1,40 @@
...
+    {{#if build.isPending}}
+      Receiving build
+    {{else if build.isProcessing}}
+      Processing visual differences
+    {{else if build.isFinished}}
       {{#if build.isApproved}}
-        were <strong>approved</strong>
-        {{#if build.approvedBy}}
-          <span class="ApprovalSection-user">by {{build.approvedBy.name}}</span>
+        {{#if build.hasNoDiffs}}
```
fotinakis: Can't use `hasDiffs` or `hasNoDiffs` logically here — "hasDiffs" will still be true for a build, even if all of them are approved. We treat "having diffs" and "having unreviewed snapshots" as different things.

If you ever check `build.isApproved` and then need to switch on something else, likely you should be checking `build.reviewStateReason` next and doing:

```hbs
{{#if (eq build.reviewStateReason 'all_snapshots_approved')}}
{{else if (eq build.reviewStateReason 'all_snapshots_approved_previously')}}
{{else if (eq build.reviewStateReason 'no_diffs')}}
{{/if}}
```

app/models/build.js
PixelJanitor: Nice catch. This was here and previously used as a placeholder.

timhaines:
Take a look inside the build info.  Is all in a big font.

PixelJanitor:
Font sizing fixed!

fotinakis:
@PixelJanitor added a commit here, just want to enumerate all the review_state_reasons because we will likely expand them over time. https://github.com/percy/percy-web/pull/421/commits/ec94ac197fd937616afc6e12403abb3206403970

PR #432, opened 2018-02-19
Upgrade ember-cli-mocha to 0.15.0 which breaks acceptance tests

cadeParade:
# Do not merge

These test failures are legitimate and the upgrade breaks all of our tests that require logging in. It breaks between ember-cli-mocha v0.14.5 and v0.15.0-beta.1
https://github.com/ember-cli/ember-cli-mocha/blob/master/CHANGELOG.md

The symptom is that whenever it runs a test requiring authentication, it redirects the test to the homepage (dev.percy.local:422) which breaks the test flow so the test will time out.

The `authenticateSession` method in `setup-acceptance` gets called but does not correctly authenticate the session, causing the `else` block in `loadCurrentUser` method in our session service to be hit, which redirects to root.

The `ember-cli-mocha` functions used in our tests are from `ember-simple-auth-auth0` which still uses an old version (0.14.4) of `ember-cli-mocha`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "ember-cli-htmlbars-inline-precompile": "^1.0.2",
     "ember-cli-inject-live-reload": "^1.7.0",
     "ember-cli-mirage": "0.3.3",
-    "ember-cli-mocha": "^0.14.4",
+    "ember-cli-mocha": "^0.15.0",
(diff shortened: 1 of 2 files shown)


cadeParade:
Closing for a REAL FIX COMING SOON

PR #435, opened 2018-02-19
Remove snapshot id param (aka: goodbye auto-scrolling behavior)

cadeParade:
Removing the snapshot id query parameter solves both scroll-related issues in clubhouse.
https://app.clubhouse.io/percy/story/1594/auto-scrolling-is-broken-in-builds-with-a-mix-of-approved-unapproved-snapshots
https://app.clubhouse.io/percy/story/1599/if-click-to-approve-a-snapshot-then-scroll-down-to-the-next-one-without-waiting-for-the-approval-once-the-first-one-is-approved

Removing this means that clicking in a snapshot will not generate a unique URL, and navigating to a URL with a snapshot id query param will not scroll to that snapshot anymore. But we can still get a unique URL from a snapshot by either using the nice copy-link button or copying the full-screen page url.

The keyboard nav for hopping up and down snapshots still works, including for builds that have more than 150 diffs (where they are all collapsed). Hopping up and down via keyboard still scrolls the page to the correct snapshot.

However, to make the keyboard nav work, there still needs to be some scroll behavior. I've rewritten the scroll behavior to be less code and make more sense (IMO).

This *does not* fix the bug currently in production where, when closing a full screen snapshot, it does _not_ scroll the the correct position for that snapshot.

## QA suggestions:
Click on some snapshots. Does what you expect happen? Click approve, click some more snapshots, scroll and click some snapshots, use the arrow keys to navigate, open the no-diff batch of snapshots and click/keyboard nav around there.


app/components/snapshot-list.js
```diff
@@ -108,97 +110,21 @@ export default Component.extend({
...
-      let selectedComponent = snapshotComponents.objectAt(selectedIndex);
-      let lastIndex = this.get('lastSnapshotIndex');
-
-      this.get('updateActiveSnapshotId')(selectedComponent.get('snapshot.id'));
-
-      // Expand the selected component.
-      selectedComponent.get('setAsSelected').call(selectedComponent);
+  _snapshotIds: mapBy('computedSnapshots', 'id'),
+  _numSnapshots: alias('computedSnapshots.length'),
+  _calculateNewActiveSnapshotId({isNext = true} = {}) {
```
fotinakis: you'll have to explain this syntax to me sometime ;)

app/components/snapshot-viewer.js
```diff
@@ -31,57 +29,41 @@ export default Component.extend({
...
-  isDefaultExpanded: true,
-  isFocus: false,
-  isExpanded: computed('isDefaultExpanded', 'snapshot.isApproved', 'build.isApproved', function() {
-    if (this.get('build.isApproved')) {
-      return true;
-    } else if (this.get('snapshot.isApproved')) {
-      return false;
-    } else {
-      return this.get('isDefaultExpanded');
+  isActiveSnapshot: computed('activeSnapshotId', 'snapshot.id', function() {
```
fotinakis: So every snapshot-viewer computes if it is the current active snapshot? I guess that's probably fine, but probably not super performant in some situations.

tests/integration/components/snapshot-list-test.js
```diff
@@ -115,4 +113,57 @@ describe('Integration: SnapshotList', function() {
     expect(SnapshotList.isNoDiffsBatchVisible).to.equal(false);
     expect(SnapshotList.snapshots().count).to.equal(numSnapshots);
   });
+
+  describe('when there are more than 150 snapshots with diffs', function() {
```
fotinakis: Yay thank you! 🎉

tests/integration/components/snapshot-list-test.js
```diff
@@ -115,4 +113,57 @@ describe('Integration: SnapshotList', function() {
...
+      });
+      percySnapshot(this.test);
+    });
+
+    it('allows keyboard nav with up and down arrows', function() {
+      SnapshotList.typeDownArrow();
+      expect(SnapshotList.snapshots(0).isExpanded).to.equal(true);
+      expect(SnapshotList.snapshots(0).isFocused).to.equal(true);
+      expect(SnapshotList.snapshots(1).isExpanded).to.equal(false);
+      expect(SnapshotList.snapshots(1).isFocused).to.equal(false);
```
fotinakis: Take a screenshot to see the expanded one, even though the scroll position will be wrong?

tests/pages/build-page.js
```diff
@@ -51,12 +48,8 @@ const BuildPage = {
     return `/${build.project.fullSlug}/builds/${build.id}?snapshot=${snapshot.id}`;
   },
 
-  typeDownArrow: triggerable('keydown', SELECTORS.SNAPSHOT_LIST, {
-    eventProperties: {keyCode: DOWN_ARROW_KEY},
-  }),
-  typeUpArrow: triggerable('keydown', SELECTORS.SNAPSHOT_LIST, {
-    eventProperties: {keyCode: UP_ARROW_KEY},
-  }),
+  typeDownArrow: alias('snapshotList.typeDownArrow'),
```
fotinakis: I learn something new Ember from you every time I read your code ⚡️  this is a nice way to keep things DRY.

app/components/snapshot-list.js
cadeParade: The syntax is super weird. But this is es6's attempt at named arguments with defaults: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment and cmd+f for "Setting a function parameter's default value"

tests/pages/build-page.js
cadeParade: Anna A contributed this property to ember page objects :)

djones:
Should we remove the subtle border darken when you select a snapshot?

screen shot 2018-02-26 at 5 30 36 pm

timhaines:
I think the only thing the border is helpful for now, is knowing which snapshot you have selected before pressing an up/down key.  It doesn't impact other interactions/decisions?

cadeParade:
That's right @timhaines it no longer does anything except mark which snapshot your keyboard nav will start from.

PR #443, opened 2018-02-22
Don't poll for snapshots when we don't need them. Don't show snapshots on a failed build.

cadeParade:
This PR accomplishes 2 things:
1) It stops polling for snapshots every 3 seconds on the build page while the build is processing or pending. It still polls for the build to check the build status, but since we are no longer displaying snapshots while the build is pending/processing, we don't need to continue refetching them.

2) It fixes the bug we saw in design review where snapshots would show up on a failed build page. No more snapshots on failed or expired build pages.


cadeParade:
@fotinakis Can you take a look at these percy snapshots? I think they are right now, but let's just confirm with more than one person.

app/components/build-container.js
```diff
@@ -14,16 +14,16 @@ export default Component.extend(PollingMixin, {
...
   shouldPollForUpdates: or('build.isPending', 'build.isProcessing'),
 
   pollRefresh() {
     this.get('build')
       .reload()
       .then(build => {
-        build.get('snapshots').reload();
+        if (build.get('isFinished')) {
+          build.get('snapshots').reload();
+          this.get('pollingTask').cancel();
```
fotinakis: Can you explain why `shouldPollForUpdates` above doesn't do this, and we have to manually cancel the task?

app/models/build.js
```diff
@@ -44,7 +44,8 @@ export default DS.Model.extend({
       return 'Timed out';
     }
   }),
-  isRunning: or('isFinished', 'isFailed', 'isExpired'),
+  isRunning: or('isPending', 'isProcessing'),
+  isFinishedRunning: not('isRunning'),
```
fotinakis: This is unused? Same with isRunning now too I think.

cadeParade:
@fotinakis Ok so your percy snapshot comments bring up some interesting questions:

Should we set up the entire application environment whenever we take a percy snapshot? These tests are specifically titled to emphasize the showing of snapshot list or not, and these snapshots show _that particular aspect_ correctly. But since these tests are visual, we notice when data unnecessary for a very specific test is not present.

I normally don't like to set up tests with more data than what is needed to test it. In this case, the build failure reason, snapshot count, and project do not directly relate to snapshot list showing up or not. If all of that information is in the test, I would argue that a future developer would have a harder time figuring out what the test was doing and what parts were important due to all the extraneous data.

But this makes percy snapshots weird, since percy cannot tell which part to focus on.
However, I would argue they are not _too_ weird, as we'll still get signal if the snapshot changes.

Also re: expired should not show "The project is not enabled" -- that is a legitimate code path, it happens in the `else` for `isExpired` state in `build-header` when the project is not expired (in this case it is because there is no project)

cadeParade:
@fotinakis I don't think we show total snapshots in the header while snapshots are pending or processing

fotinakis:
@cadeParade `total_snapshots` will be set for processing builds, but not for pending ones.

cadeParade:
@fotinakis yes it is set, but we don't currently surface that number in the UI for processing builds, so the test also won't show it.

app/components/build-container.js
cadeParade: nope, you're right, we don't need this,

app/models/build.js
cadeParade: Yes, good catch. `isFinishedRunning` is not used anymore. But `isRunning` is still being used in the `build-container` template.

PR #444, opened 2018-02-23
Ability to download the HTML behind a given snapshot

djones:


app/components/snapshot-viewer-header.js
```diff
@@ -51,5 +52,11 @@ export default Component.extend({
     closeDropdown() {
       this.set('dropdownVisible', false);
     },
+
+    downloadHTML(type, snapshot) {
+      const url = utils.buildApiUrl(`${type}Asset`, snapshot.get('id'));
+      window.location.replace(url);
```
cadeParade: Not a blocker: Should we replace the URL? maybe we could open it in a new window/tab?

tests/pages/components/snapshot-viewer-header.js
```diff
@@ -64,9 +65,8 @@ export const SnapshotViewerHeader = {
   isDropdownToggleVisible: isVisible(SELECTORS.DROPDOWN_TOGGLE),
   isDropdownPaneVisible: isVisible(SELECTORS.DROPDOWN_PANE),
   dropdownOptions: collection({
-    itemScope: SELECTORS.DROPDOWN_PANE,
+    itemScope: SELECTORS.DROPDOWN_PANE_ITEMS,
```
cadeParade: Thanks for fixing this 😬

PR #449, opened 2018-02-26
Refactor snapshot list, plus snapshot order caching

cadeParade:
This PR makes snapshots stay in the same order, during the same page session, even if you approve a snapshot.

There are two cases when snapshots need to be ordered:
1. When you load a finished build page 
2. When you are on a build page where the build is processing, and then it polls and gets the finished build.
This is accomplished by creating a route action (that calls a controller action) that is accessible to both the build route and the build polling component.

Before this, when you approved a snapshot, that snapshot would immediately move to the bottom of the snapshot list leaving the user to wonder where it went. Now, the snapshot collapses in place, so the user can see they just approved a snapshot.

To accomplish this, we added percy-web's first controller, and further gutted `snapshot-list` component. Additionally, there are tests that cover both ordering scenarios above.

TODO:
- [x] too many snapshots message shows while snapshots are loading.
- [x] snapshots all show as collapsed while loading, then expand. looks bad.


tests/acceptance/build-test.js
```diff
@@ -159,6 +163,75 @@ describe('Acceptance: Build', function() {
...
+
+      const firstSnapshotExpectedName = defaultSnapshot.name;
+      const secondSnapshotExpectedName = twoWidthsSnapshot.name;
+
+      BuildPage.visitBuild(urlParams);
+
+      andThen(() => {
+        expect(BuildPage.snapshots(0).name).to.equal(firstSnapshotExpectedName);
+        expect(BuildPage.snapshots(1).name).to.equal(secondSnapshotExpectedName);
+        BuildPage.snapshots(0).clickApprove();
```
fotinakis: I believe all of these are async helpers? So this could optionally be pulled out of the `andThen`. Can't wait until we can flatten all of these with async/await anyway so feel free to ignore.

tests/acceptance/build-test.js
```diff
@@ -159,6 +163,75 @@ describe('Acceptance: Build', function() {
...
+
+      BuildPage.visitBuild(urlParams);
+
+      andThen(() => {
+        // We approved the snapshot that would normally be seen as first (default snapshot).
+        // So the normal second snapshot (twoWidthsSnapshot) will now be first, and defaultSnapshot
+        // will be second.
+        expect(BuildPage.snapshotList.lastSnapshot.name).to.equal(defaultSnapshot.name);
+        expect(BuildPage.snapshots(0).name).to.equal(twoWidthsSnapshot.name);
+        percySnapshot(this.test);
```
fotinakis: Same with this, this is an async helper and could be outside of andThen blocks.

timhaines:
I've spent some time manually QA'ing (not code reviewing) and it's looking good to me.

tests/acceptance/build-test.js
cadeParade: These suggestions are correct 👍 👍

PR #450, opened 2018-02-27
Wording updates on snapshot download: HTML → source

djones:


PixelJanitor:
image

This should be in some sort of submenu, but I added some icons to unblock you for now. This is G2G!

PR #451, opened 2018-02-28
Give admins access to snapshot source downloads

djones:
Introduce a new `includePercyMode` option for `buildApiUrl()`. It adds a query param of `percy-mode=admin` to the resulting URL if the user is in admin mode. This option should be enabled when the `X-Percy-Mode` header cannot be set but we need to tell the backend we want superuser mode.

This is particularly needed when you direct the browser to a URL that doesn't run via the `JSONAPIAdapter`.

Sister PR: https://github.com/percy/percy-api/pull/383


app/lib/utils.js
```diff
@@ -35,9 +48,16 @@ export default {
         path = path.replace('%@', arg);
       });
     }
+
     return window.location.origin + path + queryParams;
   },
 
+  adminMode() {
```
cadeParade: This seems a bit redundant.

djones:
@cadeParade with https://github.com/percy/percy-web/pull/451/commits/79b8daa38872ddc302657786e601f2b779923c73#diff-41903b950cc3372ce37a8036b82c1292R27 an undefined param is no longer sent. Thanks for finding that bug.

Yes, the backend has this completely locked down. All the frontend is doing in this case is requesting admin mode be considered, but the backend is doing the verification. The security model is no different to the existing one where you can currently go set the `X-Percy-Mode` header yourself on the frontend.

PR #468, opened 2018-03-16
Build column info

PixelJanitor:


app/templates/components/build-header.hbs
```diff
@@ -1,9 +1,46 @@
...
+        {{build.baseBuild.branch}}
+      </code>
+    </div>
+    <div class="f6">
+      Build {{#link-to "organization.project.builds.build" build.project.organization.slug build.project.slug build.baseBuild.id class="text-mono"}}{{build.baseBuild.buildNumber}}{{/link-to}}
+      -
+      {{moment-from-now build.baseBuild.createdAt interval=60000}}
+    </div>
+  </div>
+  <div style="margin-left: -8px">
```
cadeParade: Why this inline style?

app/templates/components/build-header.hbs
```diff
@@ -1,9 +1,46 @@
...
+    <figure class="svg-container">
+      {{inline-svg "build-column-separation-icon"}}
+    </figure>
+  </div>
+  <div class="w-50 text-light text-weight-semibold p-2">
+    <div class="flex align-center no-wrap">
+      Changes from
+      {{#if build.commit}}
```
cadeParade: When would there not be a commit?

app/styles/keratin/base/typography.scss
```diff
@@ -68,6 +68,7 @@ samp {
   margin-bottom: 0;
   font-family: var(--keratin-monospace-font);
   font-size: calc(var(--keratin-body-font-size) - 4px);
+  font-weight: 400 !important;
```
cadeParade: Important?

app/templates/components/build-header.hbs
PixelJanitor: It's a one off case for just this icon. No reason to create a class for this one liner

app/templates/components/build-header.hbs
PixelJanitor: /shrug This is logic that was brought from everywhere else that uses a PR or commit. I assumed it was the safest route by reflecting those other implementations

app/styles/keratin/base/typography.scss
PixelJanitor: These types of elements can be inside another element that might have a utility class on it. Utility classes utilize `!important`s, but we ALWAYS want our monospaced font elements to have a font weight of 400. `important` on this ensures that the surrounding utility classes don't change that.

app/templates/components/build-header.hbs
cadeParade: In 8th grade I learned never to use inline styles....😱

PixelJanitor: Oh for sure. It's a long standing, unspoken rule that is misunderstood. If something only happened in one place for one element and is never shared, it's nbd to style it there with it's markup.

PR #472, opened 2018-03-19
First build state for build column header

PixelJanitor:


app/templates/components/build-header.hbs
```diff
@@ -4,19 +4,28 @@
     {{build-overview build=build}}
   </div>
 </div>
-<div class="flex">
+<div class="flex align-center" style="box-shadow: 0 1px 0 var(--gray-100)">
```
PixelJanitor: Used a box shadow here instead of a border utility class so that the snapshot headers don't double border that space, but yet we still get the bottom line on the column header for when a snapshot header isn't there. Without this, it sits on empty space and the center divider ends at nothing.

PR #473, opened 2018-03-19
Show the snapshot names when a render_timeout occurs

timhaines:
**IMPORTANT: Deploy API before deploying this**

Addresses: https://app.clubhouse.io/percy/story/1678/if-a-build-fails-due-to-1-snapshot-show-the-snapshot-name-in-the-ui-for-the-build


app/templates/components/build-header.hbs
```diff
@@ -62,8 +62,13 @@
   {{/if}}
   {{#if (eq build.failureReason 'render_timeout')}}
     <div class="p-2 m-2 bg-gray-000 rounded border border-gray-100">
-      <div class="text-weight-semibold mb-1">Some snapshots in this build took too long to render even after multiple retries. This can happen in rare cases when a page loads an asset that takes too long and hangs the Percy renderer.</div>
-      We have automatically recorded this error and will look into it, but feel free to <a href="#" {{action showSupport}} data-test-build-overview-show-support>reach out</a> and we can help get this fixed.
+      {{#with build.failureDetails as |d|}}
```
cadeParade: We should name this variable something better than `d`. `failureDetail` maybe? Or `snapshotName` describes what it actually is.
I realize I started this :P

tests/factories/build.js
```diff
@@ -43,7 +44,10 @@ FactoryGuy.define('build', {
     },
     missingResources: {failureReason: 'missing_resources'},
     noSnapshots: {failureReason: 'no_snapshots'},
-    renderTimeout: {failureReason: 'render_timeout'},
+    renderTimeout: {
+      failureReason: 'render_timeout',
+      failureDetails: {failed_snapshots: 'The snapshot that failed'},
```
cadeParade: Isn't the value of `failed_snapshots` an array of strings?

app/models/build.js
```diff
@@ -52,6 +52,7 @@ export default DS.Model.extend({
       return 'Timed out';
     }
   }),
+  failureDetails: DS.attr(),
```
cadeParade: Can we add a comment above the new model property that says it will be a POJO of form `{something: [strings, strings, etc]}`and therefore not be able to trigger computed property recomputes?

timhaines:
@cadeParade Fixed the broken things.  I agree re snapshot name - have made it bigger (in h2 tag).  I haven't reformatted them moving them to new lines though - at the moment after one times out, we cancel the rest of the snapshots, so there's only going to be one.  Want to keep this change timeboxed as it's an infrequently used error page.

fotinakis:
IMO, I love the data this presents, but think it looks odd design wise. The title is super long and is 2X larger than any other text on the page.

Proposal (use a "This build as failed" title the same as above and styled the same—or drop it entirely—and use the normal font-size list for the user-provided snapshot title content):

screen shot 2018-03-19 at 10 45 13 pm

timhaines:
@fotinakis Okay, I've switched the h2 out for div, making the text small again.

The aesthetics of the error message now match what was there before.  It's not pretty, but it conveys the information needed.  

I'm reluctant to spend any more time on this.  Missing info is now on the screen as per the story.  If we want to make error messages prettier, we can create a new story for it and prioritize it accordingly.

timhaines:
Thanks y'all.  I agree, better formatting would be nicer.  2 snapshots failing would be incredibly rare though - not going to invest time on it now.

PR #474, opened 2018-03-21
Force reload orgs on user relationships.

cadeParade:
The user <-> organization relationship strikes again. For these two instances, it was not always re-fetching organizations so the organization list would not exist. Force reloading causes these two lists to display correctly.


app/templates/components/organizations/jump-to-billing.hbs
```diff
@@ -2,16 +2,18 @@
   Choose your organization:
 </header>
 <div class="OrganizationsJumpToBilling-pageClickOverlay" {{action "hide"}}></div>
-{{#power-select
+{{#with (await orgs) as |organizations|}}
```
timhaines: very interesting!

app/templates/components/organizations/jump-to-billing.hbs
```diff
@@ -2,16 +2,18 @@
...
-    options=currentUser.organizations
+    options=organizations
     onchange=(action "chooseOrganization")
     as |organization|
     }}
     {{organization.name}}
-{{/power-select}}
\ No newline at end of file
+  {{/power-select}}
+{{/with}}
```
timhaines: Would you be open to adjusting your editor to always add newlines to the end of files?  It's probably a setting.  Prettier would do this automatically for JS files I think.

app/templates/components/organizations/jump-to-billing.hbs
cadeParade: This is a helper Stanley wrote, and it's helpful in some circumstances although it does not come without its caveats :)

https://github.com/fivetanley/ember-promise-helpers

app/templates/components/organizations/jump-to-billing.hbs
cadeParade: Great idea

PR #475, opened 2018-03-21
Seed faker when setting up FactoryGuy

cadeParade:
To prevent flaky builds like this : https://percy.io/percy/percy-web/builds/597277

Seed factoryguy for integration tests as well as acceptance tests.


tests/integration/components/build-state-badge-test.js
```diff
@@ -3,8 +3,9 @@ import {setupComponentTest} from 'ember-mocha';
 import {beforeEach, it, describe} from 'mocha';
 import {percySnapshot} from 'ember-percy';
 import hbs from 'htmlbars-inline-precompile';
-import {make, manualSetup} from 'ember-data-factory-guy';
+import {make} from 'ember-data-factory-guy';
```
timhaines: Is this needed?

cadeParade: `make` is used in all these files, it's just not in the area that github is showing you. Also, if it wasn't used, the linter would throw an error.

timhaines: @cadeParade Ha yes!  Was totally there all along.  Sorry, missed that!

PR #476, opened 2018-03-21
Fix data loading state

cadeParade:
This fixes two bugs with the UI related to loading states:
One where only a couple projects would show on the project page before the rest would load, the ~other where the "too many snapshots to view" notice would show up on the build page before all the snapshots were loaded.~


app/components/build-container.js
```diff
@@ -13,6 +13,12 @@ export default Component.extend(PollingMixin, {
   snapshotsChanged: null,
   snapshotsUnchanged: null,
 
+  areSnapshotsPending: or(
+    'snapshotsChanged.isPending',
+    'snapshostUnchanged.isPending',
+    'isSnapshotsLoading',
```
fotinakis: This doesn't exist?

cadeParade: Fair question, there is a property `isSnapshotsLoading` which we manually handle in the route and controller that is set to false after the snapshots are sorted. It wasn't working the way it was intended for a couple of weeks, but I just pulled master and can't reproduce this particular problem on master or prod any longer...
So I'll remove this change for now....

My guess for why this wasn't working was that there was some sort of race condition OR the app was asking for snapshots twice, so even though we loaded them initially, maybe it was asking for them again and waiting on that.

PR #478, opened 2018-03-22
Show branch on build page when missing commit data

timhaines:
Updates:
image

To:
image


app/templates/components/build-header.hbs
```diff
@@ -39,7 +39,7 @@
       {{else}}
         Changes from
       {{/if}}
-      {{#if build.commit}}
+      {{#if build.branch}}
```
fotinakis: branch is always required, we could drop this.

timhaines: Great point

PR #480, opened 2018-03-22
Build progress bars

PixelJanitor:


PixelJanitor:
Going to need @cadeParade or @fotinakis look at the test on line 74 of `build-container-test.js` that keeps failing. Not sure if this is a result of removing the loader or something. This is ready for review after that's fixed!

PixelJanitor:
I see the behavior that's happening with showing while loading, and it's not happening on `master`. I rebased from `master` and it's still happening.

app/components/build-overview.js
```diff
@@ -4,6 +4,13 @@ export default Component.extend({
   build: null,
   classes: null,
   tagName: '',
+  buildCompletionProgress: 0,
 
   showSupport() {},
+
+  actions: {
+    updateBuildCompletionProgress() {
```
cadeParade: Do we need this action anymore? Wasn't this only being used by the countup component?

cadeParade: Same with the buildCompletionProgress property above.

app/models/build.js
```diff
@@ -83,8 +83,18 @@ export default DS.Model.extend({
...
   totalComparisonsFinished: DS.attr('number'),
   totalComparisonsDiff: DS.attr('number'),
+  buildCompletionPercent: computed('totalComparisons', 'totalComparisonsFinished', function() {
+    if (!this.get('isProcessing')) {
+      return 0;
+    }
+
+    const totalComparisons = this.get('totalComparisons') || 0;
+    const totalComparisonsFinished = this.get('totalComparisonsFinished') || 0;
+    return totalComparisonsFinished / totalComparisons * 100;
```
cadeParade: I am not sure what to do with this but if `totalComparisons` is 0, we will end up dividing by 0 which returns NaN I think. Which will throw off further calculations....
Maybe in the `if` statement above, we can add `or totalComparisons = 0` and it will return early as 0 rather than trying to divide by 0?

app/models/build.js
```diff
@@ -83,8 +83,18 @@ export default DS.Model.extend({
   totalSnapshots: DS.attr('number'),
   totalSnapshotsUnreviewed: DS.attr('number'),
 
+  totalComparisons: DS.attr('number'),
   totalComparisonsFinished: DS.attr('number'),
   totalComparisonsDiff: DS.attr('number'),
+  buildCompletionPercent: computed('totalComparisons', 'totalComparisonsFinished', function() {
```
cadeParade: We need this to depend on `isProcessing` also, since we are checking `isProcessing` in the computed property.

PR #483, opened 2018-03-26
[WIP] Allow users to click "Back" from full screen snapshot

cadeParade:
This problem was because there were nested routes rendering into the same template. See issue here for more information: https://github.com/emberjs/ember.js/issues/2186

The solution is instead of using `builds/build` route, use `builds/build/index` route. Then when you transition between `builds/build/index` and `builds/build/snapshot` route, you'll transitioning between sibling routes rather than a child and a parent.

This refactor also allows the deletion of a bunch of modal management code. Going to the `builds/build/snapshot` route just shows the modal. There's no controlling the modal state, only controlling the route.

Fixes https://app.clubhouse.io/percy/story/1506/navigating-to-a-fullscreen-view-on-the-builds-page-and-back-results-in-a-blank-page

### DIFF
--- a/app/components/build-container.js
+++ b/app/components/build-container.js
@@ -6,10 +6,6 @@ import {inject as service} from '@ember/service';
 export default Component.extend(PollingMixin, {
   build: null,
   classNames: ['BuildContainer'],
-  classNameBindings: [
-    'classes',
-    'isShowingModal:BuildContainer--snapshotModalOpen:BuildContainer--snapshotModalClosed',
-  ],
 
   snapshotQuery: service(),
   snapshotsChanged: null,
--- a/app/components/snapshot-list.js
+++ b/app/components/snapshot-list.js
@@ -43,14 +43,12 @@ export default Component.extend({
     $(document).bind(
       'keydown.snapshots',
       function(e) {
-        if (!this.get('isShowingModal')) {
-          if (e.keyCode === 40) {
-            // down arrow
-            this.set('activeSnapshotId', this._calculateNewActiveSnapshotId({isNext: true}));
-          } else if (e.keyCode === 38) {
-            // up arrow
-            this.set('activeSnapshotId', this._calculateNewActiveSnapshotId({isNext: false}));
-          }
+        if (e.keyCode === 40) {
+          // down arrow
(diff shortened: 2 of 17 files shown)


fotinakis:
Unfortunately I think we may not be able to do this. :( We currently use the transition from a parent to child route `builds/build` --> `builds/build/snapshot` in order to preserve state on the parent route.

You can see this by doing something like:
1) click a responsive nav item, or toggle off diff mode
2) click full screen mode
3) press ESC
4) notice that the state is still the same (except, right now scroll position is broken)

In this PR, this becomes:
1) click a responsive nav item, or toggle off diff mode
2) click full screen mode
3) press ESC
4) notice that the page state is fully reset to its beginning state, as if you had just loaded the page.

There are definitely currently two bugs that made there way in to prod:
1) from full screen mode, scroll position is not retained after pressing ESC
2) from full screen mode, pressing back button causes blank screen (I think this is because the .BuildContainer is still "display: none").

So fundamentally, I think making fullscreen mode a sibling route will break any ability to retain state handling.

cadeParade:
Ok, I see these bugs, but here are some devil's advocate questions:
1. Has any user ever complained about these features being gone? The back button has been broken for a long time and afaik no users have mentioned it? Which means that (a) not that many users use that flow and (b) even if they do use that flow, not that many users notice/care that the state is not saved.
2. The application has a new route defined for the fullscreen modal. But in fact, we do not want any of the things related to routes to happen when transitioning back and forth between the fullscreen/build route. We do not want model hooks to run, or any transitional things to happen when we transition between those two routes, it seems like the only thing we are using the route for is query parameters. Additionally, the fact that this works at all seems like taking advantage of a loophole in ember, see discussion here: https://github.com/emberjs/ember.js/issues/2186
It seems like a new route is actually not what we want here, maybe we just want query parameters on the build route. If that was the case, I believe all the things we want (saving state, scroll position) would work automatically and we wouldn't be working against Ember.

Interested to hear your thoughts!

cadeParade:
Closing because this is fixed with https://github.com/percy/percy-web/pull/497

But I still think we should consider alternatives to this routing structure.

PR #484, opened 2018-03-28
Filter query for snapshots

cadeParade:
Run with https://github.com/percy/percy-api/pull/413 for nice results.


app/services/snapshot-query.js
```diff
@@ -0,0 +1,22 @@
+import Service, {inject as service} from '@ember/service';
+
+export default Service.extend({
```
fotinakis: I love that you wrapped this into a query service 👍we should really do this way more.

cadeParade: It makes stubbing it easier too, but we don't really need to stub queries in components that much...queries are usually in routes and not really reused. And also, the way Ember is intended to be used is to load all the data in a route. But it's not really possible in practice to load aaallll the data in the routes ...

app/services/snapshot-query.js
```diff
@@ -0,0 +1,23 @@
+import Service, {inject as service} from '@ember/service';
+import {SNAPSHOT_REVIEW_STATE_REASONS, DIFF_REVIEW_STATE_REASONS} from 'percy-web/models/snapshot';
+
+export default Service.extend({
+  store: service(),
+  getUnchangedSnapshots(buildId) {
```
fotinakis: Convention in our codebases is to pass objects around instead of IDs whenever possible, for future flexibility and easier debugging. Would prefer that here and then do `build.get('id')` in the place below.

Here and in `getChangedSnapshots`.

mirage/config.js
```diff
@@ -241,8 +241,20 @@ export default function() {
     return invite;
   });
 
+  this.get('/builds/:build_id/snapshots', function(schema, request) {
+    const build = schema.builds.findBy({id: request.params.build_id});
+    const queryParams = request.queryParams;
+    if (queryParams['filter[review-state-reason]']) {
+      const reasons = queryParams['filter[review-state-reason]'].split(',');
```
fotinakis: Very nice!

mirage/factories/build.js
```diff
@@ -15,8 +17,54 @@ export default Factory.extend({
...
+    reviewStateReason: 'all_snapshots_approved',
+  }),
+
+  approvedPreviously: trait({
+    state: BUILD_STATES.FINISHED,
+    reviewState: 'approved',
+    reviewStateReason: 'all_snapshots_approved_previously',
+  }),
+
+  no_diffs: trait({
```
fotinakis: --> noDiffs
(or probably `approvedWithNoDiffs` is better?)

mirage/factories/build.js
```diff
@@ -15,8 +17,54 @@ export default Factory.extend({
...
+
+  expired: trait({
+    state: BUILD_STATES.EXIPRED,
+  }),
+
+  failed: trait({
+    state: BUILD_STATES.FAILED,
+  }),
+
+  renderTimeout: trait({
```
fotinakis: I think all 3 of these should have `failedWith` prepended, otherwise `create('build', 'noSnapshots')` doesn't look like a failure state.

- failedWithRenderTimeout
- failedWithNoSnapshots
- failedWithMissingResources

Alternatively, we could require that people always use `failed` trait + another one, but that seems like something easy to miss.

tests/acceptance/project-test.js
```diff
@@ -120,74 +120,35 @@ describe('Acceptance: Project', function() {
...
-      });
-      server.create('build', {
-        project,
-        createdAt: moment().subtract(25, 'minutes'),
-        state: 'failed',
-        failureReason: 'render_timeout',
-      });
-      server.create('build', {
+      server.create('build', 'failed', {project, createdAt: _timeAgo(3, 'hours')});
+      server.create('build', 'renderTimeout', {project, createdAt: _timeAgo(25)});
```
fotinakis: `_timeAgo(25)` causes question marks in my brain, no idea how long ago this is. :) Can we always include the timeUnit?

tests/acceptance/build-test.js
```diff
@@ -6,109 +6,14 @@ import BuildPage from 'percy-web/tests/pages/build-page';
 import {TEST_IMAGE_URLS} from 'percy-web/mirage/factories/screenshot';
 import {SNAPSHOT_APPROVED_STATE, SNAPSHOT_REVIEW_STATE_REASONS} from 'percy-web/models/snapshot';
 import {BUILD_STATES} from 'percy-web/models/build';
-
-describe('Acceptance: Pending Build', function() {
```
cadeParade: These tests just went to the bottom since I never really want to look at these first when I open this file.

config/environment.js
```diff
@@ -46,6 +46,7 @@ module.exports = function(environment) {
         emailVerifications: '/api/v1/email-verifications/%@',
         baseAsset: '/api/v1/snapshots/%@/assets/base.html',
         headAsset: '/api/v1/snapshots/%@/assets/head.html',
+        organizationSnapshots: '/api/v1/builds/%@/snapshots',
```
fotinakis: `organizationSnapshots` --> `buildSnapshots`

cadeParade: yuss

PR #485, opened 2018-03-29
Universal diff toggle

PixelJanitor:


cadeParade:
Here's an ember tooltip library that looks reasonable on first glance: https://github.com/sir-dunxalot/ember-tooltips

cadeParade:
Also, it seems like we need to pay for hint.css? See bottom of page here: https://kushagragour.in/lab/hint/

PR #486, opened 2018-04-01
[WIP] Run tests in Chrome.

fotinakis:
Walk through `ember init` diffs and get things as close as we can.

Also, because phantomjs compatibility was removed in Ember 3, move our tests to run in Chrome. (I am unsure why they were passing before in the ember-3 branch).


fotinakis:
This is weirdly corrupt, ditching.

PR #496, opened 2018-04-06
Use rel=noopener in links with target=_blank

timhaines:
It's good security practice to use rel=noopener when using target=_blank.

https://mathiasbynens.github.io/rel-noopener/


PixelJanitor:
This is great. Thanks, Tim!

PR #497, opened 2018-04-06
Full-screen snapshot adjustments

cadeParade:
Fixes two problems:

1. After we loaded a partial set of snapshots for a build, the way the full-screen snapshot route was set up would _still_ load all snapshots for a build before loading which made it very slow. This refactors these routes to only load necessary snapshots for both build route and full-screen snapshot route.

2. Blank screen when clicking browser `back` button out of full-screen snapshot mode.


app/routes/organization/project/builds/build.js
```diff
@@ -69,12 +78,13 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
-    closeSnapshotFullModal(buildId) {
-      this.send('updateModalState', false);
-      this.transitionTo('organization.project.builds.build', buildId);
+    closeSnapshotFullModal() {
+      this.send('updateIsHidingBuildContainer', false);
+      this.transitionTo('organization.project.builds.build', this.get('build.id'));
     },
 
-    createReview(action, build, snapshots) {
+    createReview(action, snapshots) {
```
fotinakis: Is it weird that there are now three different `createReview` interfaces across the code base? (snapshots), (action, snapshots), and (action, build, snapshots).

Anything we can do there to make that clearer/more-explicit/easier to maintain and understand context?

app/templates/organization/project/builds/build/snapshot.hbs
```diff
@@ -1,12 +1,12 @@
 {{#modal-dialog wrapperClass="SnapshotViewerFullModalWrapper"}}
   {{snapshot-viewer-full
-    build=build
-    snapshotId=snapshotId
+    snapshot=model
+    isBuildFinished=build.isFinished
```
fotinakis: This is a big ol' code smell 👃 — ie. why pull out this specific property and decouple it from the rest of the system, rather than pass through the build object? What happens when we need some other property from the build?

Would highly prefer that we pass through the `build` object if we can, but I'm sure there was motivation for doing this this way.

cadeParade: Mostly that the component we're passing this into is encompassing a snapshot, so it shouldn't need to know about the build.
What about passing something like `isSnapshotApprovable=build.isFinished` which reflects what this property is used for in the end?

fotinakis: Per discussion, can fix the snapshots <> build link and then do an alias for this.

app/routes/organization/project/builds/build.js
fotinakis: Per discussion, we can clear this up by moving the action to the service and having the interface be like `reviewService.approve(snapshots)`.

app/components/snapshot-approval-button.js
```diff
@@ -9,7 +9,7 @@ export default Component.extend({
   actions: {
     approveSnapshot() {
       this.set('isLoading', true);
-      this.createReview([this.get('snapshot')])
+      this.get('createReview')([this.get('snapshot')])
```
fotinakis: How about we semantically name this something else so it isn't as confusing and doesn't muddy the interface of createReview itself.

How about `approveSnapshots([...])` which flows up to the top-level reviews service approval?

PR #498, opened 2018-04-08
Pricing page visual update

PixelJanitor:
This PR: 
Updates the pricing page formatting, especially:
- Header with three tiers of pricing
- Features table
- FAQ section

The sections above are also now driven by contentful.
This PR also deletes several components used on the old pricing page which are no longer used.

Check out [this snapshot to see what it will look like]( https://percy.io/percy/percy-web/builds/885325/view/51841079/1280?browser=chrome&mode=head)

The acceptance tests for the pricing page are skipped for now due to waiting on contentful support. But once the tests are unskipped, it would look like this: https://percy.io/percy/percy-web/builds/885325/view/51841079/1280?browser=chrome&mode=head.
We're not using the current implementation for now because the tests are getting the live contenful data. That means if anyone changed anything in that contenful data, it would show up in every Percy run for every branch. Instead we'll make a test environment for Contenful and get fake (but correctly structured) content from there.


app/routes/pricing.js
```diff
@@ -1,7 +1,26 @@
...
+    });
+  },
+
+  setupController(controller, resolvedModel) {
+    controller.setProperties({
+      tableData: resolvedModel.tableData,
+      smallCard: resolvedModel.cardData.smallCard,
+      mediumCard: resolvedModel.cardData.mediumCard,
+      largeCard: resolvedModel.cardData.largeCard,
+      faqs: resolvedModel.faqData,
```
maprules1000: ❤️

app/templates/pricing.hbs
```diff
@@ -1,15 +1,181 @@
...
-    {{pricing-section showSupport="showSupport" redirectToDefaultOrganization="redirectToDefaultOrganization"}}
-  </div>
-</div>
-<div class="container container--light">
-  <div class="container container--fixedWidth container--forContent">
-    {{pricing-questions}}
-  </div>
-</div>
+<header class="bg-gradient-pattern border-b border-purple-300 pb-9" style="margin-bottom: -144px">
+  {{fixed-top-header class="mb-9" transparent-header=true}}
```
maprules1000: `transparent-header` should be camelCase. Hyphens should only be used for components.

it should probably be `isTransparentHeader` too since it's only a boolean value.

app/templates/components/fixed-top-header.hbs
```diff
@@ -27,8 +27,8 @@
   {{trial-bar organization=currentOrganization type="is-on-dark"}}
 
   {{#if session.isAuthenticated}}
-    {{nav-menu user=session.currentUser currentOrganization=currentOrganization}}
+    {{nav-menu user=session.currentUser currentOrganization=currentOrganization transparent-header=transparent-header}}
   {{else}}
-    {{nav-menu}}
+    {{nav-menu transparent-header=transparent-header}}
```
maprules1000: all these `transparent-header`s should be camel cased and be prefixed with `is`

app/routes/pricing.js
```diff
@@ -10,4 +29,33 @@ export default Route.extend(ResetScrollMixin, {
       this.analytics.track('Pricing Viewed');
     },
   },
+
+  _getPricingCardData() {
+    return this.store.findAll('pricing-plan-card').then(cards => {
+      const smallCard = cards.findBy('id', '4x8tkdOmKsEQ2eKWc68CyS');
+      const mediumCard = cards.findBy('id', '3o7cIf5i1aQoMuIkUGwO8K');
+      const largeCard = cards.findBy('id', 'bvzQmZjBCwIC4Awa6gwc2');
```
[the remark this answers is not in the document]
cadeParade: Yes. I chose to use the `id`s rather than any of the content fields (like `name` for example) because the content fields could change.

app/templates/pricing.hbs
cadeParade: Excellent point.

PR #502, opened 2018-04-15
Bump `ember-cli-sass` version

twokul:


timhaines:
Building the app (like with `yarn test` fails after updating this package.  (cc: @PixelJanitor )

```
Error: Invalid CSS after "...var(--gray-100)": expected "}", was "!important;"
        on line 2 of tmp/sass_compiler-input_base_path-IUpSUkpV.tmp/app/styles/keratin/utilities/status.scss
        from line 27 of tmp/sass_compiler-input_base_path-IUpSUkpV.tmp/app/styles/keratin/keratin.scss
        from line 3 of tmp/sass_compiler-input_base_path-IUpSUkpV.tmp/app/styles/app.scss
>>   --status-color: var(--gray-100) !important;
```

PixelJanitor:
Interesting. Will look into ASAP!

twokul:
This is quite interesting. I was getting the exact same error when running `master` of `ember-cli` against `master` of `percy-web`; bumping `ember-cli-sass` helped remedy that issue. I'm not too familiar with Sass syntax but from what I can tell it's a valid syntax.

timhaines:
@twokul I think if you'd done an `npm i` rather than installing with yarn it would have installed the more recent version of the package (ignoring our lock file) breaking things for you.  I'm not sure what else could have happened there.

twokul:
@timhaines I used `yarn`, but retrying it now to see if I can reproduce it locally again

twokul:
```sh
rm -rf node_modules tmp dist && yarn && ember t --reporter dot
...
649 tests complete (30613 ms)
```
🤔

timhaines:
@twokul you're right.  I can reproduce the results you're reporting.

timhaines:
@twokul I'd tried upgrading ember-cli-sass just by running `yarn upgrade ember-cli-sass` from master, and had seen the breakage I'd reported.  This command results in a bunch of other packages being updated too though.  

Updating package.json with the new ember-cli-sass version specification, and then running `yarn`, does a cleaner package update, and results in only the changes shown in your PR.  Which don't break.  TIL.

timhaines:
@PixelJanitor Disregard my earlier comment.  The problem lies elsewhere.

PixelJanitor:
👍🏻

PR #509, opened 2018-04-25
For large builds, only load images if in the viewport.

fotinakis:
https://app.clubhouse.io/percy/story/1829/improve-experience-for-large-builds-with-many-changes

This was mostly an experiment based on today's discussion, but seems to work really well. It is also only enabled for builds with > 150 changed snapshots, otherwise behavior is unchanged.

Questions: 
Why is deferredImageLoading only passed to some image-spacer components?
What is the deal with _isExpanded property?
What is the correct number for when deferred loading should kick in ?

### DIFF
--- a/app/components/loading-page.js
+++ b/app/components/loading-page.js
@@ -2,4 +2,6 @@ import Component from '@ember/component';
 
 export default Component.extend({
   classNames: ['LoadingPage'],
+  attributeBindings: ['dataTestId:data-test-id'],
+  dataTestId: null,
 });
--- a/app/templates/components/image-spacer.hbs
+++ b/app/templates/components/image-spacer.hbs
@@ -1,5 +1,5 @@
 {{#if shouldLoadImage}}
   {{simple-image image=image classes=imageClass}}
-{{else}}
+{{else if (not deferredImageLoading)}}
   {{loading-page}}
 {{/if}}
--- a/app/templates/components/image-spacer.hbs
+++ b/app/templates/components/image-spacer.hbs
@@ -1,5 +1,5 @@
 {{#if shouldLoadImage}}
   {{simple-image image=image classes=imageClass}}
 {{else if (not deferredImageLoading)}}
-  {{loading-page}}
+  {{loading-page dataTestId=dataTestId}}
 {{/if}}
--- a/app/templates/components/image-spacer.hbs
+++ b/app/templates/components/image-spacer.hbs
@@ -1,5 +1,6 @@
 {{#if shouldLoadImage}}
   {{simple-image image=image classes=imageClass}}
-{{else if (not deferredImageLoading)}}
+{{else if isHidingSpinner}}
+{{else}}
   {{loading-page dataTestId=dataTestId}}
 {{/if}}
--- a/app/templates/components/image-spacer.hbs
+++ b/app/templates/components/image-spacer.hbs
@@ -1,3 +1,5 @@
 {{#if shouldLoadImage}}
   {{simple-image image=image classes=imageClass}}
+{{else}}
+  {{loading-page}}
 {{/if}}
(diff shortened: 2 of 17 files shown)


cadeParade:
Wondering why put the lazy loading in the image-spacer component rather than block the whole snapshot-viewer component from loading? One of the biggest hits to performance is loading multiple nested components, so preventing the whole snapshot-viewer component from loading when not in viewport would improve performance even more?
Could have scrolling implications though....

app/components/loading-page.js
```diff
@@ -2,4 +2,6 @@ import Component from '@ember/component';
 
 export default Component.extend({
   classNames: ['LoadingPage'],
+  attributeBindings: ['dataTestId:data-test-id'],
+  dataTestId: null,
```
maprules1000: This is setup so that attribute bindings can be dynamically passed into this component since it gets instantiated multiple times in a single page object context.

This was the most ideal pattern I came up with, but I would really like to know what you think @cadeParade!

tests/pages/components/snapshot-viewer.js
```diff
@@ -1,13 +1,22 @@
...
+} from 'ember-cli-page-object';
 
 const SELECTORS = {
   SNAPSHOT_VIEWER: '[data-test-snapshot-viewer]',
   DIFF_IMAGE: '[data-test-comparison-viewer-full-diff-image-overlay] img',
   DIFF_IMAGE_BOX: '[data-test-comparison-viewer-diff-image-container] img',
   NO_DIFF_BOX: '[data-test-comparison-viewer-unchanged]',
   SHOW_UNCHANGED_COMPARISONS: '[data-test-comaprison-viewer-show-unchanged-comparisons]',
+  BASE_SNAPSHOT_LOADING_SPINNER: '[data-test-id=comparison-viewer-base-image-loading]',
+  HEAD_SNAPSHOT_LOADING_SPINNER: '[data-test-id=comparison-viewer-head-image-loading]',
```
maprules1000: These are the two loading spinners seen in the snapshot viewer. The test selectors are a little different than others

app/templates/components/image-spacer.hbs
```diff
@@ -1 +1,6 @@
-{{simple-image image=image classes=imageClass}}
+{{#if shouldLoadImage}}
+  {{simple-image image=image classes=imageClass}}
+{{else if isHidingSpinner}}
```
maprules1000: This isn't an ideal way to hide the loading spinner, but I think any other way would require a refactor that is greater than the scope of this PR. Also, I hear tales that this section of the app is going to get a big refactor in the future so that'll be an ideal time to optimize this.

maprules1000:
Btw most of the files changed (and all of the ones added) are for tests.

tests/factories/comparison.js
```diff
@@ -8,35 +9,30 @@ FactoryGuy.define('comparison', {
...
-      };
+    baseScreenshot: () => {
+      return FactoryGuy.make('screenshot', 'base');
     },
-    baseScreenshot: f => {
-      // TODO: make the screenshot and image a real FactoryGuy model instead of POJO
-      return {
-        id: f.id,
-        image: {id: f.id, url: '/images/test/v1.png', width: 375, height: 500},
-        lossyImage: {id: f.id, url: '/images/test/v1-lossy.png', width: 375, height: 500},
```
maprules1000: This is weird. @cadeParade it looks like the lossy images were `png`s originally, but the only images in the test assets are `jpg`s. Could this be why there are visual changes on other tests?

cadeParade:
Closing this in favor of https://github.com/percy/percy-web/pull/568.

The test updates here are still valuable and we can use this PR for a reference when we need to make these changes.

PR #510, opened 2018-04-25
Usage graph to billing page

PixelJanitor:


app/components/organizations/billing-section.js
```diff
@@ -17,14 +17,24 @@ export default Component.extend({
...
+
   actions: {
     changingSubscription(savingPromise) {
       this.set('isSaveSuccessful', null);
       this.set('isSaving', true);
       savingPromise.then(
         () => {
-          this.set('isSaving', false);
-          this.set('isSaveSuccessful', true);
+          self.set('isSaving', false);
```
fotinakis: There is no `self` in this context? This should be `this`?

PixelJanitor: I brought this over from what Benjamin had contributed. So maybe? :)

fotinakis:
Have synced with @cadeParade on the remaining API piece to get this finished out. Thanks for patience here.

PR #512, opened 2018-04-27
Allow connecting a GitLab version control integration to an Organization

faun:
This PR adds support for authorizing the Percy OAuth app to access the gitlab.com API on the behalf of a user.

Required backend companion changes: https://github.com/percy/percy-api/pull/437


app/components/organizations/gitlab-settings.js
```diff
@@ -0,0 +1,19 @@
...
+import Component from '@ember/component';
+
+export default Component.extend({
+  organization: null,
+
+  store: service(),
+  session: service(),
+  currentUser: alias('session.currentUser'),
+  gitlabIntegration: alias('organization.gitlabIntegration'),
+  isGitlabIntegration: alias('organization.isGitlabIntegration'),
```
cadeParade: Could line 11 and 12 be simplified by `isGitlabIntegration: bool('organization.gitlabIntegration')` as it is in the organization model?

Even better, if organization is guaranteed to be present, you could use `organization.isGitlabIntegrated` right?

app/models/organization.js
```diff
@@ -48,10 +52,16 @@ export default DS.Model.extend({
   }),
 
   isGithubEnterpriseIntegrated: bool('githubEnterpriseIntegration'),
+  isGitlabIntegrated: alias('isGitlabIntegration'),
```
cadeParade: These two properties are named incorrectly I believe -- `is` would imply they are booleans, where actually they are returning integration objects.

It makes more sense to me to have the properties `isGitxxbIntegrated` be `bool(gitxxbintegration)`. 

And then `isGitxxbIntegration`is also misleading since it makes it seem like the object you're testing would be an integration, but actual it is an org. Why do we need both `isGitxxbIntegrated` and `isGitxxbIntegration`?

tests/integration/components/organizations/github-enterprise-settings-test.js
```diff
@@ -46,6 +46,8 @@ describe('Integration: Github Enterprise Settings', function() {
     });
 
     it('shows nothing', function() {
+      const isIntegrated = this.get('organization.isGithubEnterpriseIntegrated');
+      expect(isIntegrated).to.equal(false);
```
cadeParade: This is testing whether the org in the test context (i.e. the one on line 43) does not have a property `isGithubEnterpriseIntegrated`. Is this what you're going for?

tests/pages/components/gitlab-settings.js
```diff
@@ -0,0 +1,20 @@
+import {create, text, isHidden, isVisible} from 'ember-cli-page-object';
```
cadeParade: Yaaaaaay! :D 🎉

tests/pages/components/gitlab-settings.js
```diff
@@ -0,0 +1,20 @@
...
+
+const SELECTORS = {
+  GITLAB_INTEGRATION: '.gitlab-settings',
```
cadeParade: Let's make selectors we use only for tests `data-test` attributes. So this would be `data-test-gitlab-settings`.

tests/integration/components/organizations/github-enterprise-settings-test.js
faun: Nope, this test is gabrage. Pthhff.

app/components/organizations/gitlab-settings.js
```diff
@@ -0,0 +1,19 @@
...
+  actions: {
+    redirectToGitlabAuthorization(organization) {
+      let slug = organization.get('slug');
+      window.location.replace(`/api/auth/gitlab/redirect/${slug}`);
```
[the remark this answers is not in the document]
faun: Fixed in 0083ef19

tests/integration/components/organizations/github-enterprise-settings-test.js
faun: Fixed this garbage in ad314cc9

app/models/organization.js
faun: Also fixed in dbcc9e3d

app/components/organizations/gitlab-settings.js
faun: Good point. Fixed in dbcc9e3d

tests/pages/components/gitlab-settings.js
faun: Fixed in 9dae423a

faun:
Back to you, @cadeParade.

app/components/organizations/gitlab-settings.js
```diff
@@ -0,0 +1,22 @@
...
+import utils from 'percy-web/lib/utils';
+import Component from '@ember/component';
+
+export default Component.extend({
+  organization: null,
+
+  store: service(),
+  session: service(),
+  currentUser: alias('session.currentUser'),
+  isGitlabIntegrated: bool('organization.isGitlabIntegrated'),
```
cadeParade: Low priority: 
I know I suggested this originally, but since `organization.isGitlabIntegrated` is already a boolean, this property can just be an `alias` rather that reconverting a bool into a bool. This implementation should work though.

app/models/version-control-integration.js
```diff
@@ -3,19 +3,24 @@ import DS from 'ember-data';
...
   organization: DS.belongsTo('organization'),
+  gitLabBotUser: DS.belongsTo('user'),
   githubInstallationId: DS.attr(),
   githubAccountAvatarUrl: DS.attr(),
   githubHtmlUrl: DS.attr(),
   integrationType: DS.attr(),
   githubEnterpriseHost: DS.attr(),
   githubEnterpriseInstallationId: DS.attr(),
   githubEnterpriseIntegrationId: DS.attr(),
+  gitlabIntegrationId: DS.attr(),
```
cadeParade: Not for now but...I wonder if there is an alternative way of structuring things that doesn't require a new property for every integration host...something to think about...

tests/integration/components/organizations/github-enterprise-settings-test.js
cadeParade: Hmm, this is still testing that the organization on the test context has an integration. 
I think the expectation on line 56 is sufficient to show that the organization doesn't have an integration. Integration tests are how you test that the UI correctly reflects the state of the component. If you want to set an organization on a component and test some resulting property on the component that _doesn't_ have an affect on the UI, you could use a unit test. (For example, in [this](https://github.com/percy/percy-web/blob/master/tests/unit/routes/application-test.js#L48) test, we're using a unit test to set a property on a route, call some action, then make sure the right thing happens, not involving the UI).

tests/integration/components/organizations/gitlab-settings-test.js
```diff
@@ -0,0 +1,83 @@
...
+    beforeEach(function() {
+      const user = make('user');
+      const organization = make('organization', 'withGitlabIntegration');
+      user.set('organizations', [organization]);
+      this.setProperties({user, organization});
+    });
+
+    it('shows that the integration is installed', function() {
+      const isGitlabIntegrated = this.get('organization.isGitlabIntegrated');
+      expect(isGitlabIntegrated).to.equal(true);
```
cadeParade: Same deal as mentioned elsewhere -- this test is testing that the org on the test context is github integrated. Which sort of tests whether the test is set up correctly, which is useful to know. But again, I think the UI test suffices.

tests/factories/repo.js
```diff
@@ -25,5 +25,12 @@ FactoryGuy.define('repo', {
       source: 'github_enterprise',
       hostname: () => 'enterprise-host.com',
     },
+    gitlab: {
+      name: () => faker.lorem.slug(20),
```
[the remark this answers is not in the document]
faun: Fixed in fd4a9936

app/components/organizations/gitlab-settings.js
faun: Fixed in e900324a

tests/integration/components/organizations/gitlab-settings-test.js
faun: Fixed in 0632aaa5

tests/integration/components/organizations/github-enterprise-settings-test.js
faun: Fixed in 0632aaa5

app/models/version-control-integration.js
faun: 🤔

PR #519, opened 2018-05-03
Add browser models

cadeParade:
In preparation for allowing users to see and interact with browsers for a build, we need to have models that represent them.

Requires https://github.com/percy/percy-api/pull/439#issuecomment-386397462 to be merged.


app/models/browser.js
```diff
@@ -0,0 +1,9 @@
+import DS from 'ember-data';
+import {alias} from '@ember/object/computed';
+
+export default DS.Model.extend({
+  builds: DS.hasMany('build', {inverse: null}),
```
fotinakis: Drop this? We should never do `browser.builds`, the arrow (data relationship) should always go build --> browsers.

mirage/factories/browser.js
```diff
@@ -0,0 +1,18 @@
+import {Factory, trait} from 'ember-cli-mirage';
+
+export default Factory.extend({
+  id(i) {
+    return `${i}`;
+  },
+
+  version: 'abc.123',
+  afterCreate(browser, server) {
```
fotinakis: You should be able to drop this `afterCreate` and just do:

```
browserFamily: association(),
```

and it should pick up the ember-data model association.

mirage/factories/build.js
```diff
@@ -18,6 +18,11 @@ export default Factory.extend({
     return i + 1;
   },
 
+  afterCreate(build, server) {
```
fotinakis: Same here, hopefully? `browsers: association(),` not sure about hasMany's though.

mirage/factories/browser.js
cadeParade: This is awesome

mirage/factories/build.js
cadeParade: I get this error when I try it: `Error: You're using the `association` factory helper on the 'browsers' attribute of your build factory, but that attribute is not a `belongsTo` association. Read the Factories docs for more information: http://www.ember-cli-mirage.com/docs/v0.3.x/factories/#factories-and-relationships`

Looks like it doesn't work with hasMany :(

cadeParade: It's a really friendly error though!

PR #527, opened 2018-05-08
Replace Stripe Checkout with Stripe Elements for CC Update

fotinakis:


maprules1000:
What's Left
---

- [ ] Styling and positioning (@PixelJanitor)

Description
---
This replaces Stripe Checkout with a much more streamlined and better integrated Stripe Element. The stripe popup with a weird "Update Credit Card" plan name no longer pops up. Instead the user sees a very clean and friendly.

Pictures
---
Coming Soon™

Reviewing this PR
---
This feature is tricky to test because Stripe Elements is securely loaded in an iFrame by Stripe. As a result, there are no automated tests. 

Pull this branch down locally an try to update the credit card using one of Stripe's [Test Cards](https://stripe.com/docs/testing#cards). There should be a successful network call to Stripe's server to get the token, and then another successful call to the Percy server to process the token.

app/templates/components/organizations/billing-card-updater.hbs
```diff
@@ -0,0 +1,21 @@
...
+  {{#stripe-card
+    options=options
+    change=(action 'checkCard')
+    as |stripeElement stripeError|}}
+    {{#if stripeError}}
+      <p class="error">{{stripeError.message}}</p>
+    {{/if}}
+
+    {{#if shouldShowSubmit}}
+      <button class="Button" onclick={{perform updateCreditCard stripeElement}}>Submit</button>
```
cadeParade: Do we need to use the `onclick` handler here (rather than `{{action ...}}`)? It is better to not mix ember actions and browser events. Here is a talk which goes into more detail than you could ever want to know about this topic: https://www.youtube.com/watch?v=G9hXjjHFJVs and the accompanying blog post: https://medium.com/square-corner-blog/deep-dive-on-ember-events-cf684fd3b808

app/components/organizations/billing-card-updater.js
```diff
@@ -0,0 +1,70 @@
...
+
+    // Get or create the plan record with the right ID.
+    let plan = this.get('store').peekRecord('plan', planId);
+    plan = plan || this.get('store').createRecord('plan', {id: planId});
+
+    let savingPromise = subscriptionService.changeSubscription(organization, plan, token);
+    this.get('_updateSubscriptionSavingStatus').perform(savingPromise);
+  },
+
+  _updateSubscriptionSavingStatus: task(function*(savingPromise) {
```
cadeParade: Let's talk about this pattern!

PixelJanitor:
I'm finished up with the styling here @maprules1000

app/components/organizations/billing-card-updater.js
```diff
@@ -0,0 +1,70 @@
...
+    const planId = this.get('planId');
+    const response = yield this.get('stripeService').createToken(stripeElement);
+    this._changeSubscription(planId, response.token);
+  }),
+
+  _changeSubscription(planId, token) {
+    const organization = this.get('organization');
+    const subscriptionService = this.get('subscriptionService');
+
+    // Get or create the plan record with the right ID.
```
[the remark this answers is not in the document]
maprules1000: It's cool because this component won't actually show up in the interface unless they already have a non-free plan
https://github.com/percy/percy-web/pull/527/files#diff-46261f75b2baabbaf2a7da5b3c28b58dR173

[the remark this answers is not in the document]
cadeParade: Then why would you have to potentially create a plan?

app/components/organizations/billing-card-updater.js
```diff
@@ -0,0 +1,77 @@
...
+      yield savingPromise;
+      this.get('flashMessages').success('Your card was updated successfully!');
+      this.set('isSaveSuccessful', true);
+      this.setProperties({
+        isSaving: false,
+        isUpdatingCard: false,
+        isCardComplete: false,
+      });
+    } catch (e) {
+      this.get('flashMessages').danger(
```
cadeParade: awwright :)

app/components/organizations/billing-card-updater.js
```diff
@@ -0,0 +1,77 @@
...
+        'There was a problem updating your card.' +
+          ' Please try again or contact customer support.',
+      );
+      this.set('isSaveSuccessful', false);
```
cadeParade: Was testing this out locally and looks like we need `this.set('isSaving', false')` here also, or the spinner will appear forever if it fails.

Not necessary for now:
Speaking of `isSaving` property, if we keep using ember concurrency (which I am for), we should get familiar with all the stuff it can do, for example, each `task` has an `isRunning` property which we could use to get rid of the manual `isSaving` management.
http://ember-concurrency.com/docs/derived-state

maprules1000: I'll fix this and also look into using the derived state in the template. That would be a great code cleanup

app/components/organizations/billing-card-updater.js
maprules1000: 🎉

app/components/organizations/billing-card-updater.js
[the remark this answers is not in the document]
maprules1000: That is a very valid point. Will look into

PR #528, opened 2018-05-09
Add maprules1000 to dependenciesbot settings

cadeParade:
Also partially update to v2 settings syntax


cadeParade:
FYI @maprules1000

maprules1000:
giphy

PR #529, opened 2018-05-10
[Multiple browser support Part 1] Make builds and comparisons browser-aware (No user-facing changes)

cadeParade:
This PR updates the existing app to use the idea of `filteredComparisons` instead of comparisons. A `filteredComparison` is the comparison for a width and a certain browser.  Because we only ever want to see a comparison for a single browser, we always need to filter a snapshot's comparisons by browser.

This PR converts the app to invisibly use `filteredComparisons`. It should introduce no user-facing changes. This PR sets up the framework to allow users to switch browsers, since snapshots and comparisons now have knowledge of the active browser.

This happens with the addition of the `filtered-comparisons` object. This object takes many of the repeated properties from `snapshot-viewer`, `snapshot-viewer-header`, and `snapshot-viewer-full` and combines them into one object. The object is sort of like an intermediate model for `browserComparisons`. 
[Several alternatives approaches to the mixin were explored](https://docs.google.com/document/d/1lRueqe-pJGHN5HM-u9yT8Kp5_eY4BPxwH7HkouvHsBM/edit#heading=h.ncallmtvetbx), but this solution seemed like it was the easiest to maintain and had the fewest and least serious cons.


tests/integration/components/snapshot-list-test.js
```diff
@@ -96,11 +96,10 @@ describe('Integration: SnapshotList', function() {
     percySnapshot(this.test);
   });
 
-  describe('when there are more than 150 snapshots with diffs', function() {
-    const numSnapshots = 151;
+  describe('when there are too many snapshots with diffs', function() {
+    const numSnapshots = 10;
```
maprules1000: 🎉 This is great!

tests/factories/browser.js
```diff
@@ -1,10 +1,10 @@
 import FactoryGuy from 'ember-data-factory-guy';
 
 FactoryGuy.define('browser', {
-  sequences: {
-    id: i => i,
-  },
   default: {
+    // Make this always the same number by default so all the default models can pretend
+    // they have the same browser
+    id: 1,
```
maprules1000: Thanks for this comment!

tests/factories/comparison.js
```diff
@@ -9,7 +9,9 @@ FactoryGuy.define('comparison', {
     finishedProcessingAt: () => moment().subtract(23, 'seconds'),
     diffRatio: 0.23,
     browser: () => {
-      return FactoryGuy.make('browser');
+      // Make this always the same number by default so all the default models can pretend
+      // they have the same browser
+      return FactoryGuy.make('browser', {id: 1});
```
maprules1000: Does this need an explicit id here?

app/components/snapshot-viewer-full.js
```diff
@@ -1,36 +1,38 @@
...
   updateComparisonMode: null,
   transitionRouteToWidth: null,
   closeSnapshotFullModal: null,
   createReview: null,
+  snapshot: null,
+  snapshotSelectedWidth: null,
+  activeBrowser: null,
+
+  filteredComparisons: computed('snapshot', 'activeBrowser', 'snapshotSelectedWidth', function() {
+    return filteredComparisons.create({
```
maprules1000: I think using the filtered comparisons is an elegant solution. It abstracts everything out

app/components/snapshot-viewer-full.js
```diff
@@ -1,36 +1,38 @@
...
   comparisonMode: null,
   updateComparisonMode: null,
   transitionRouteToWidth: null,
   closeSnapshotFullModal: null,
   createReview: null,
+  snapshot: null,
+  snapshotSelectedWidth: null,
+  activeBrowser: null,
+
+  filteredComparisons: computed('snapshot', 'activeBrowser', 'snapshotSelectedWidth', function() {
```
maprules1000: should this computed property have a different name than the imported object?

tests/factories/comparison.js
cadeParade: Great question. At some point I had thought so, but just tried it without and all the tests pass. So I'll take it out!

app/components/snapshot-viewer-full.js
cadeParade: This doesn't need to have a different name because it is scoped inside the component. Also I like having the name be the same so all the similar things can be found by the same name. But do you have a name that you think would be better?

maprules1000: Ah ok, I think it's fine then

app/components/snapshot-viewer-full.js
fotinakis: This is so nice! 🥇

fotinakis: Composition over inheritance FTW.

app/models/browser.js
```diff
@@ -3,7 +3,7 @@ import {alias} from '@ember/object/computed';
 
 export default DS.Model.extend({
   browserFamily: DS.belongsTo('browserFamily', {async: false, inverse: null}),
-  name: alias('browserFamily.name'),
-  slug: alias('browserFamily.slug'),
+  familyName: alias('browserFamily.name'),
```
fotinakis: Nice

tests/unit/lib/filtered-comparisons-test.js
```diff
@@ -0,0 +1,143 @@
...
+      const snapshotWithChrome = {comparisons: [chromeComparisonWithDiff]};
+      const snapshotWithFirefox = {comparisons: [firefoxComparisonWithDiff]};
+      const snapshotWithBothBrowsersAndDiffs = {
+        comparisons: [chromeComparisonWithDiff, firefoxComparisonWithDiff],
+      };
+      const snapshotWithBothBrowsersAndMixedDiffs = {
+        comparisons: [chromeComparisonNoDiff, firefoxComparisonWithDiff],
+      };
+
+      expect(
```
fotinakis: (optional personal style nitpick preference) The heavy nesting makes this hard to read, using a few variables and multiple lines here would flatten this kind of thing up and make it easier to read.

Ie.

```js
const allSnapshots = [
  snapshotWithChrome,
  snapshotWithFirefox,
  snapshotWithBothBrowsersAndDiffs,
  snapshotWithBothBrowsersAndMixedDiffs,
];
let snapshots = snapshotsWithDiffForBrowser(allSnapshots, chromeBrowser)
expect(snapshots).to.eql([snapshotWithChrome, snapshotWithBothBrowsersAndDiffs]);
```

vs.

```js
expect(
  snapshotsWithDiffForBrowser(
    [
      snapshotWithChrome,
      snapshotWithFirefox,
      snapshotWithBothBrowsersAndDiffs,
      snapshotWithBothBrowsersAndMixedDiffs,
    ],
    chromeBrowser,
  ),
).to.eql([snapshotWithChrome, snapshotWithBothBrowsersAndDiffs]);
```

cadeParade: Fair, fair, updated.

cadeParade:
Yes, I've done quite a bit of testing on this branch to make sure things work correctly. 

The only visible change is an extra query parameter in fullscreen view `browser=whatever`

PR #534, opened 2018-05-10
user facing changes for adding browser support

cadeParade:
This is the second part of the work to integrate multiple browsers into Percy. This PR builds on #529 to add the browser-switcher component in the build header and fullscreen snapshot header.


cadeParade:
Some UX notes mentioned by Tim related to performance: 
- When you switch browsers and there are a lot of snapshots in the browser you're switching to, it takes a long time and the browser switcher buttons remain in their half selected state for a long time.

Other notes:
- Numbers in snapshot count badges look really small
- The selection state of the buttons feels like it’s not quite obvious enough

cc @PixelJanitor

PixelJanitor:
:on-it:

app/components/build-container.js
```diff
@@ -11,10 +13,27 @@ export default Component.extend(PollingMixin, {
...
   snapshotQuery: service(),
-  snapshotsChanged: null,
-  numSnapshotsUnchanged: null,
   snapshotsUnchanged: null,
   allDiffsShown: true,
+  updateActiveBrowser: null,
+  isUnchangedSnapshotsVisible: false,
+
+  snapshotsChanged: computed('allChangedBrowserSnapshotsSorted', 'activeBrowser.id', function() {
+    if (!this.get('allChangedBrowserSnapshotsSorted')) return;
```
fotinakis: The linter should enforce this but it apparently doesn't — we should avoid 1-line if statements for clarity, let's always use if statements with blocks. This follows most common JS and other styleguides.

app/components/build-container.js
```diff
@@ -23,19 +42,47 @@ export default Component.extend(PollingMixin, {
...
+    this.set('isUnchangedSnapshotsVisible', false);
+  },
+
   actions: {
     showSnapshotFullModalTriggered(snapshotId, snapshotSelectedWidth, activeBrowser) {
       this.sendAction('openSnapshotFullModal', snapshotId, snapshotSelectedWidth, activeBrowser);
     },
 
+    updateActiveBrowser(newBrowser) {
+      this.set('chosenBrowser', newBrowser);
```
fotinakis: I think this might be deserving of some analytics tracking (as a followup to this PR) — I think we should come up with a name ('Build Browser Selected'?) to add to the Event Taxonomy spreadsheet and then add tracking to this as a followup.

app/components/build-container.js
```diff
@@ -23,19 +42,47 @@ export default Component.extend(PollingMixin, {
...
+      this._resetUnchangedSnapshots();
+    },
+
+    toggleUnchangedSnapshotsVisible() {
```
fotinakis: Also think we should add tracking for this as well (outside of this PR).

app/controllers/organization/project/builds/build.js
```diff
@@ -1,51 +1,42 @@
...
-
-    const numSnapshotsMissing = this.get('build.totalSnapshots') - snapshots.get('length');
-    this.set('numSnapshotsMissing', numSnapshotsMissing);
-
-    this.set('isSnapshotsLoading', false);
+  // Creates a hash with a keys of each browser id,
+  // and the correctly ordered snapshots as the values and sets it as
+  // allChangedBrowserSnapshotsSorted.
+  initializeSnapshotOrdering() {
+    const orderedBrowserSnapshots = {};
```
fotinakis: This is great! 🎉  Super clear for how complex of things this method is doing.

app/templates/components/browser-switcher.hbs
```diff
@@ -0,0 +1,17 @@
...
+    <button
+      type="button"
+      class="browser-switcher btn btn-square position-relative {{if (eq browser.id activeBrowser.id) "is-browser-active"}} {{concat "data-test-browser-switcher-" browser.familySlug}}"
+      {{action updateActiveBrowser browser}}
+      data-test-browser-switcher-button
+    >
+      {{inline-svg (concat browser.familySlug "-browser-icon") class="opacity-25"}}
+      {{#if build}}
+        <div class="browser-switcher-diff-count position-absolute bg-white font-weight-semibold px-sm lh-1 z-1 {{if (eq browser.id activeBrowser.id) "text-white shadow-purple-lg bg-purple-600" "text-purple"}}" data-test-browser-switcher-diff-count>
+          {{get (get unapprovedSnapshotsWithDiffForBrowsers browser.id) "length"}}
```
fotinakis: Would be much easier to read if this was a computed property

app/components/build-container.js
cadeParade: I'll add clubhouse tickets for these suggestions.

cadeParade: https://app.clubhouse.io/percy/story/1952/add-analytics-for-browser-switching-events

app/templates/components/browser-switcher.hbs
cadeParade: Fair

PR #537, opened 2018-05-15
Add Deprecation Workflow

maprules1000:
Description
---
Adds `ember-cli-deprecation-workflow` and silences existing deprecation by explicitly documenting them in `config/deprecation-workflow.js`. Now, all deprecations are in one place and they don't clutter the browser console.

_Note: You might get some weird test failures locally. If you do, you'll have to restart your browser. Then the errors will disappear._

Reviewing this PR
---
Pull this branch down locally, run `yarn` to install the new addon, and run your tests locally. You should see no explicit deprecations in your console, and all tests should pass.


cadeParade:
Will the "weird test failures locally" keep happening in the future and will it require a browser restart each time? Or only the first run?

maprules1000:
@cadeParade It looks like it only happens once and doesn't come back after a browser reboot, even after switching branches, killing the server, rerunning `yarn`, and restarting the server.

maprules1000:
The error in question is `Cannot read property '__deprecatedInstance__' of null`

cadeParade:
Ok, thanks for the details! I've experienced this before also, so it's unrelated to the package update. Also IIRC, you only need to close and reopen the tab when you see this, not the whole browser.

PR #540, opened 2018-05-16
Add pull request template

maprules1000:
(below is the new template!)

## Description
This PR adds a default template for PRs to make them a little more informative, and to link back to the clubhouse card where it all started.

## Photos 
Not applicable here, but will be the place to highlight sweet interface changes without going through all the diffs in percy.

## Reviewing this PR
I would really like feedback on the usefulness of the 4 sections here, and the language in the actual template that suggests what to put in each section.

- Should we have clubhouse links on this public repository?
- Do we really want/need a photo section? (I think a highlight or two is super useful)
- Anything else?


maprules1000:
Awesome! @cadeParade I'll add a testing call out in the `Reviewing this PR` section and ask that users explain when a PR is not covered by tests.

I've updated this PR description to show what that looks like too.

PR #541, opened 2018-05-16
generate a working package-lock.json based on our yarn.lock file

ballPointPenguin:
I generated the package-lock.json with a combination of https://github.com/imsnif/synp, sed, and prettier.
I nuked my node_modules dir and tested it with node v.8.x and npm v.5.x and the web app runs locally and I can log in without any Auth0 problems.

Do we actually want a package-lock.json *and* a yarn.lock? I'm not sure.


cadeParade:
I'm also not sure what the advantage would be for having both. It seems like more to maintain and yarn works fine for us at the moment. Do you have any advantages in mind for suggesting to add a package-lock.json?

timhaines:
I don’t think we want to have both.  If we decided to switch tooling and use npm instead of yarn, then we should discuss first, and update throughout our CI and deployment stack.  Maintaining both lock file types and keeping them in sync is more of a burden than what it gains us IMO.

timhaines:
@ballpointpenguin not sure if you saw - Lindsay updated the Auth0 package today.

djones:
Agreed with others here. We shouldn't have both and if we change our default tooling we should have a discussion first.

I suggest we close this PR and if there's support for a change in tooling, let's discuss and revisit.

ballPointPenguin:
I also do not think the effort of switching is justified. There is nothing we need that Yarn is not giving us, afaik. Npm continues to improve and npm 6.0 brings some new security and audit features to the table. I am glad to know that if and when we want to switch, the process is relatively painless.

ballPointPenguin:
I'm going to leave the steps to (re)-generate package-lock.json from yarn.lock here for future reference:

```sh
yarn install
npx synp --source-file yarn.lock
npx prettier --write package-lock.json
rm -rf node_modules
npm install
```

PR #547, opened 2018-05-18
Remove some computed style deprecations

cadeParade:
## Description
Our buildkite logs are too big to display all of them at once and that is ridiculous. In an effort to make this reasonable, let's remove some of the chatter in the logs. One noisy test is the teams page, it is unhappy about assigning the style attribute in that dynamic way so I've moved the random calculation into the JS.

## Pictures
:(
image

Covered by a Percy test

This should not affect how users see the Teams page (although the random sort may be different initially in the Percy test)

### Update:
Woohoo, this is enough to push our test output to fit on one page (76 warnings)...but let's keep getting rid of these notices :D


cadeParade:
cc @maprules1000

PR #549, opened 2018-05-21
revert dependencies.yml to v1

cadeParade:
## Description
I tried to update dependencies.yml to use v2 syntax, but it makes bad PRs now, so I reverted it back to v1 (plus @maprules1000 which was the original goal)(which is still supported on the site).


cadeParade:
omg i love this photo please keep going

PR #553, opened 2018-05-22
Remove redundant flashMessage

maprules1000:
## Description
Remove redundant stripe failure message. Frankly, the existing one is better.


cadeParade:
Which "existing one" are you talking about?

PR #554, opened 2018-05-22
Do not display collapsed message when build is approved and therefore snapshots are expanded

cadeParade:
## Description
When a build is approved, we automatically expand all snapshots with diffs. However, when there's "too many" snapshots on a page (over 150) we always display the "Whew! There are too many snapshots to show all at once, so they are collapsed below." message. This message is wrong when there's > 150 of snapshots on a page where the build is approved.
This PR removes this message for builds that are approved.

#### Before
image

#### After
image

## Reviewing this PR

Does that behavior make sense?


tests/integration/components/snapshot-list-test.js
```diff
@@ -84,6 +87,16 @@ describe('Integration: SnapshotList', function() {
       expect(SnapshotList.snapshots(1).isExpanded).to.equal(false);
       expect(SnapshotList.snapshots(1).isFocused).to.equal(false);
     });
+
+    it('expands all snapshots if build is approved', function() {
```
maprules1000: Hurray for tests! 🎉

PR #555, opened 2018-05-23
Browser switcher finished state

PixelJanitor:
Adds clarity to when browser comparisons are approved:

image


app/templates/components/browser-switcher-badge.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="browser-switcher-diff-count text-center position-absolute bg-white font-weight-semibold px-sm lh-1 z-1 {{if isActive "text-white shadow-purple-lg bg-purple-600" "text-purple"}} {{if (and (eq unnapprovedSnapshotCount 0) isActive) "bg-green-500 shadow-green-lg"}}" data-test-browser-switcher-diff-count>
+  {{#if unnapprovedSnapshotCount}}
+    {{unnapprovedSnapshotCount}}
+  {{else}}
+  <div data-test-browser-switcher-diff-count-all-approved>
```
maprules1000: I'm really not a fan of adding empty markup like this just for testing, but I couldn't think of any better way to make this element available to a page object.

fotinakis: Seems good to me, these data attributes get stripped in production so there are no worries here about clutter or performance outside of just test maintainability and convenience.

maprules1000: but without the data attribute the empty div tag will still be there  😕

PixelJanitor: Was thinking the same. It's not a huge issue, just an un-needed div. How about passing the attribute to the check-icon instead? Will that work for your test or to test attributes need to be on a markup element instead of a component itself?

PixelJanitor: It takes classes, but you're right about no data attrs. I'd like to create a new svg component to remove all the console warnings altogether :)

PR #557, opened 2018-05-24
Slight wait for loading builds with no diffs

cadeParade:
## Description

To reproduce this problem:
- Go to projects page.
- Refresh the page.
- Navigate to a build that has no diffs.
- Sometimes the snapshots will not display.

For some reason, if there are no snapshots, even though the browsers are in the store and associated with the build, the browsers do not get to a state where they are recognized before this method ends. 
One stupid, naive way to solve this is wait a tick when displaying builds with no diffs. This fixes the problem, but I can't say why.

## Reviewing this PR
_**Is Covered by Automated Tests?: No**_
I tried writing a test to replicate this problem but the test scenarios all passed before this update. I'm not sure how to reproduce this in a test since I can't properly explain the problem. :(

The diff is confusing -- No logic inside `initializeSnapshotOrder` changed, it was just renamed to `_initializeSnapshotOrder` and turned into a task. It also has an added `if/else` in the middle to decide whether to use the delay.


app/controllers/organization/project/builds/build.js
```diff
@@ -18,17 +19,25 @@ export default Controller.extend({
...
+    const buildSnapshotsWithDiffs =
       this.get('build')
         .hasMany('snapshots')
         .value() || [];
 
+    // If there are no snapshots with changes, this method will complete too fast for
+    // the loading state to display properly. This gives a small delay so we can associate
+    // all the data together and display the build page correctly.
+    if (!buildSnapshotsWithDiffs.get('length')) {
+      yield timeout(300);
```
maprules1000: This works, but long term I feel like we might want to come up with a better solution to show the loading state, possibly with some sort of in-and-out transition animation? 

Or even a abstract this timeout to the loading component so it'll always display for at lease a half second or something.

app/controllers/organization/project/builds/build.js
```diff
@@ -18,17 +19,25 @@ export default Controller.extend({
   // Creates a hash with a keys of each browser id,
   // and the correctly ordered snapshots as the values and sets it as
   // allChangedBrowserSnapshotsSorted.
-  initializeSnapshotOrdering() {
+  _initializeSnapshotOrdering: task(function*() {
```
maprules1000: concurrency all the things!

app/controllers/organization/project/builds/build.js
cadeParade: I agree. I don't love this solution either. I would like to get to the bottom of this issue first, to make an informed decision on the correct solution.

app/controllers/organization/project/builds/build.js
cadeParade: I am really liking it so far!

PR #558, opened 2018-05-24
Add polling for organization repo freshness

faun:
repo refresh

##  [Clubhouse Story](https://app.clubhouse.io/percy/story/1873/poll-to-refresh-repos-list-on-the-project-settings-page)

[Related backend PR](https://github.com/percy/percy-api/pull/457)

## Description

On the project settings page, poll for updates in the background for all of the organization’s recently added or removed repositories. We’re updating all of the repos on the backend and checking for freshness on the frontend and retrying until they’re fresh.

The repos are fetched using the PollRefresh mixin and are polled until the lastUpdatedAt timestamp is current and the isSyncing prepoerty is false.

This feature exists for GitHub enterprise users using personal access tokens, but also will prove useful for GitLab  and Bitbucket integrations. The polling happens on page load until the organization is no longer flagged as polling on the backend and the frontend determines the repositories are no monger stale. Additionally, when you open the power select to select a repo, we fire off another call to refresh repos.

It’s worth mentioning also that the backend has rate limiting built in so that you can only refresh repos once during a set period. It’s for this reason that I’ve decided to exclude any manual control over repo syncing and only provide feedback on the freshness of the repositories.

## Reviewing this PR

The functionality has been manually QA and tested. I’m looking for a review of completeness of this PR and opportunities to learn idiomatic Ember. I also am open to suggestions about how to test this better. 

Also, I’d love some design and style input from @PixelJanitor.

**Is Covered by Automated Tests?**: YES, where reasonable. There was significant complexity in the asynchronous nature of polling, so I’ve left [the test](https://github.com/percy/percy-web/pull/558/files#diff-68143e15c50fe304ba3b50253ca61566R160) that covers that as pending for the moment. I’m likely going to delete it unless there are objections.


tests/integration/components/projects/repo-integrator-test.js
```diff
@@ -108,18 +110,64 @@ describe('Integration: RepoIntegratorComponent', function() {
...
+    it('displays correctly in the loading state', function() {
+      this.render(hbs`{{projects/repo-integrator project=project}}`);
+      clickTrigger();
+      expect(RepoIntegrator.dropdown.isSelectorOpen).to.eq(true);
+      expect(RepoIntegrator.repoFreshness.message).to.eq('Checking repo status...');
+      percySnapshot(this.test.fullTitle());
+    });
+
+    it('refreshes the repo collection', function() {
+      this.render(hbs`{{projects/repo-integrator project=project}}`);
```
cadeParade: You have just changed this in the middle of my comment, so this applies to the previous version:

Ok so here are a couple of notes about this test:
The repoRefreshService stub looks good, except it returns an organization and the real one returns repos.
In any case that is not why this test is failing. 
This test is failing because even though the repoRefreshStub is returning an organization with a recent `lastSyncedAt` time, nothing in the test code or otherwise tells ember that it should now be paying attention to `freshOrganization` rather than the original, `organization`, whose `lastSyncedAt` time has not changed.
To get the freshness message to update, the `lastSyncedAt` time on the organization of the project in the component will have to be updated. To do this, the service stub could be changed to be something like this: 

```
export default function mockRepoRefresh(context, repos, newLastSyncedAt) {
  const repoRefreshServiceStub = Service.extend({
    getFreshRepos() {
      context.set('project.organization.lastSyncedAt', newLastSyncedAt);
      return resolve(repos);
    },
  });
  context.register('service:repoRefresh', repoRefreshServiceStub);
  context.inject.service('repoRefresh', {as: 'repoRefresh'});
  return repoRefreshServiceStub;
}
```

Another possible approach (untested) would be to pass a new organization into the stub, and run `context.set('project.organization`, newOrg)` in the stub, but this is a bit misleading since we're not actually getting a compeletly new org, just an updated existing one, right?

Let me know if this makes sense!


For the new version, looks like you have figured out the crux of the problem, but let's try to use `async/await` instead of `Ember.run`.

faun: Fixed in 062715e6

tests/pages/components/repo-integrator.js
```diff
@@ -7,22 +7,30 @@ const SELECTORS = {
   POWER_SELECT_FIRST_GROUP: 'li.ember-power-select-group:first-of-type',
   POWER_SELECT_LAST_GROUP: 'li.ember-power-select-group:last-of-type',
   POWER_SELECT_GROUP_NAME: '.ember-power-select-group-name',
+  REPO_FRESHNESS_CONTAINER: '.ProjectsRepoIntegratorRefreshStatus',
```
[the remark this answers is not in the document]
faun: Fixed in 6c12ab75

app/components/projects/repo-integrator.js
```diff
@@ -10,7 +14,41 @@ export default Component.extend({
...
+    this.triggerRepoRefresh();
+  },
+  triggerRepoRefresh() {
+    let self = this;
+    let organization = this.get('organization');
+    this.set('isRepoRefreshInProgress', true);
+    this.get('repoRefresh')
+      .getFreshRepos(organization)
+      .then(() => {
+        self.set('isRepoRefreshInProgress', false);
```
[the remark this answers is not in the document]
faun: Fixed in 1dfe2518

app/components/projects/repo-integrator.js
```diff
@@ -10,7 +14,41 @@ export default Component.extend({
 
   selectedRepo: alias('project.repo'),
   organization: alias('project.organization'),
+
+  repoRefresh: service(),
+  isRepoRefreshInProgress: false,
+  isSyncing: alias('organization.isSyncing'),
   groupedRepos: alias('organization.groupedRepos'),
+  lastSyncedAt: alias('organization.lastSyncedAt'),
+  lastSyncedAtRelative: computed('lastSyncedAt', function() {
```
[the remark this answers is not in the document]
faun: Fixed in 28fcc254

tests/integration/components/repo-refresh-status-test.js
```diff
@@ -0,0 +1,29 @@
...
+describe('Integration | Component | repo-refresh-status', function() {
+  setupComponentTest('repo-refresh-status', {
+    integration: true,
+  });
+
+  it('renders correctly when the repo is refreshing', function() {
+    this.set('isRepoRefreshInProgress', true);
+
+    this.render(hbs`{{repo-refresh-status isRepoRefreshInProgress=true}}`);
+    expect(this.$().text()).to.include('Checking repo status...');
```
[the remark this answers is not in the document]
faun: Noted in 63ebd6c9

app/components/projects/repo-integrator.js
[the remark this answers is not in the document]
faun: Fixed in 4519a835

app/components/repo-refresh-status.js
```diff
@@ -0,0 +1,8 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+  classNames: ['ProjectsRepoIntegratorRefreshStatus'],
+  classNameBindings: ['classes'],
```
cadeParade: This is unnecessary and also the `classes: null` property in the above file. This is an old pattern before ember combined classes I guess.

app/templates/components/repo-refresh-status.hbs
```diff
@@ -0,0 +1,11 @@
+<span data-test-repo-refresh-status-message>
+{{#if isRepoRefreshInProgress }}
+  Checking repo status...
+{{else}}
+  {{#if lastSyncedAt }}
```
cadeParade: I do not object to this implementation, but fyi, it is possible to use `else if` in hbs: https://guides.emberjs.com/v3.0.0/templates/conditionals/

tests/helpers/mock-repo-refresh-service.js
```diff
@@ -0,0 +1,20 @@
...
+    getFreshRepos() {
+      let organization = context.get('project.organization');
+      if (newLastSyncedAt) {
+        organization.set('lastSyncedAt', newLastSyncedAt);
+      }
+      if (newRepos) {
+        organization.set('repos', newRepos);
+      }
+      return resolve(organization.get('repos'));
+    },
```
cadeParade: 👍

tests/integration/components/projects/repo-integrator-test.js
```diff
@@ -18,59 +20,59 @@ describe('Integration: RepoIntegratorComponent', function() {
   });
 
   describe('with a github integration', function() {
-    beforeEach(function() {
+    beforeEach(async function() {
```
cadeParade: `async` and `await` are not necessary for `beforeEach` renders. Actions in the `beforeEach` block by definition complete before each test is run, so we don't have to worry about this here.

tests/integration/components/projects/repo-integrator-test.js
```diff
@@ -79,49 +81,87 @@ describe('Integration: RepoIntegratorComponent', function() {
...
+      this.setProperties({project});
+    });
+
+    it('shows that the repos are out of date', async function() {
+      oldRepos = organization.get('repos');
+      oldUpdatedAt = organization.get('lastSyncedAt');
+      repoRefreshServiceStub(this, oldRepos, oldUpdatedAt);
+      this.render(hbs`{{projects/repo-integrator project=project}}`);
+      await clickTrigger();
+      await expect(RepoIntegrator.dropdown.isSelectorOpen).to.eq(true);
```
cadeParade: Also, I believe the last 3 awaits in this test (and the ones after the click in the next tests) are not necessary. The only thing that takes up time and makes a promise is the trigger click. The assertions are not async so do not have to be awaited.

tests/integration/components/projects/repo-integrator-test.js
```diff
@@ -108,18 +110,59 @@ describe('Integration: RepoIntegratorComponent', function() {
...
+
+  describe('with stale repos', function() {
+    let organization;
+    let freshOrganization;
+    beforeEach(function() {
+      const project = make('project');
+      organization = make('organization', 'withMultipleIntegrations', 'withStaleRepoData');
+      freshOrganization = make('organization', 'withMultipleIntegrations', 'withFreshRepoData');
+
+      project.set('organization', organization);
```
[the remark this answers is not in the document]
faun: Huh.

tests/integration/components/projects/repo-integrator-test.js
faun: Fixed in c51e3fbf

app/components/repo-refresh-status.js
faun: Fixed in 35e2091e and ba6e88bb

tests/integration/components/projects/repo-integrator-test.js
faun: Fixed in c51e3fbf

PR #562, opened 2018-05-29
Auto-Approve Branches: UI

ballPointPenguin:
## Description
see https://app.clubhouse.io/percy/story/1930/auto-approve-branches
Additional text field in the Project settings form, with a default value and a bit of explanation.

## Reviewing this PR

Covered by snapshot test and simple check within Project Acceptance test.

Review notes:
[x] Look at improving the wording / explanation of the field.
[x] Open to suggestions if you think of a better attribute name than 'autoApproveGlob'
[x] Look at testing strategy (use of before/after hooks, snapshot, and 'find' helper)

### DIFF
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -10,7 +10,7 @@ export default DS.Model.extend({
   isEnabled: DS.attr('boolean'),
   isDisabled: not('isEnabled'),
   diffBase: DS.attr(), // Either "automatic" or "manual".
-  autoApproveGlob: DS.attr('string', {defaultValue: 'master'}),
+  autoApproveBranchFilter: DS.attr('string', {defaultValue: 'master'}),
   createdAt: DS.attr('date'),
   updatedAt: DS.attr('date'),
 
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -10,6 +10,7 @@ export default DS.Model.extend({
   isEnabled: DS.attr('boolean'),
   isDisabled: not('isEnabled'),
   diffBase: DS.attr(), // Either "automatic" or "manual".
+  autoApproveGlob: DS.attr('string', {defaultValue: 'master'}),
   createdAt: DS.attr('date'),
   updatedAt: DS.attr('date'),
 
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -10,7 +10,7 @@ export default DS.Model.extend({
   isEnabled: DS.attr('boolean'),
   isDisabled: not('isEnabled'),
   diffBase: DS.attr(), // Either "automatic" or "manual".
-  autoApproveBranchFilter: DS.attr('string', {defaultValue: 'master'}),
+  autoApproveBranchFilter: DS.attr(),
   createdAt: DS.attr('date'),
   updatedAt: DS.attr('date'),
 
--- a/app/templates/components/forms/project-edit.hbs
+++ b/app/templates/components/forms/project-edit.hbs
@@ -19,7 +19,7 @@
   {{#if is-admin}}
     <h4>Auto-Approve Branches</h4>
     {{form-fields/input
-      property="autoApproveGlob"
+      property="autoApproveBranchFilter"
       title="Branch filter pattern"
       changeset=changeset
     }}
--- a/app/templates/components/forms/project-edit.hbs
+++ b/app/templates/components/forms/project-edit.hbs
@@ -16,16 +16,18 @@
     the <code>PERCY_PROJECT</code> environment variable in your CI service. Note, old build URLs
     will break if the slug is changed.
   </p>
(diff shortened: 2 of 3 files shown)


app/models/project.js
```diff
@@ -10,6 +10,7 @@ export default DS.Model.extend({
   isEnabled: DS.attr('boolean'),
   isDisabled: not('isEnabled'),
   diffBase: DS.attr(), // Either "automatic" or "manual".
+  autoApproveGlob: DS.attr('string', {defaultValue: 'master'}),
```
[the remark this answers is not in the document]
ballPointPenguin: Yes I like that better. I'll change my rails migration too.

app/templates/components/forms/project-edit.hbs
```diff
@@ -12,13 +12,27 @@
...
   </p>
+  {{#if is-admin}}
+    <h4>Auto-Approve Branches</h4>
+    {{form-fields/input
+      property="autoApproveBranchFilter"
+      title="Branch filter pattern"
+      changeset=changeset
+    }}
+    <p>
+      Separate multiple branches with spaces. Use * for wildcard, e.g. <code>master release-*</code> . These
```
djones: I'm thinking something like this for the wording:

```
Builds with branches that match this filter will be automatically approved and won't require manual approval.

<input>

Separate multiple branches with spaces. You can use <code>*<code> as a wildcard match. Example: <code>master staging-* release-*.*.*</code>
```

ballPointPenguin: ok done.

app/models/project.js
```diff
@@ -10,6 +10,7 @@ export default DS.Model.extend({
   isEnabled: DS.attr('boolean'),
   isDisabled: not('isEnabled'),
   diffBase: DS.attr(), // Either "automatic" or "manual".
+  autoApproveBranchFilter: DS.attr('string', {defaultValue: 'master'}),
```
cadeParade: Hmm, why do we need the default value defined? We don't (and should not) give the user an option to update this field in the project creation flow, so when we reach the form to edit the branch glob, the `autoApproveBranchFilter` property will already be set from the backend. We should depend on the backend for the source of truth rather than filling in data from the frontend.

app/templates/components/forms/project-edit.hbs
```diff
@@ -12,13 +12,31 @@
       validateProperty=(action "validateProperty" changeset "slug")
   }}
   <p>
-    The slug is used to identify this project in URLs and the Percy API. If changed, you must update the <code>PERCY_PROJECT</code> environment variable in your CI service. Note, old build URLs will break if the slug is changed.
+    The slug is used to identify this project in URLs and the Percy API. If changed, you must update
+    the <code>PERCY_PROJECT</code> environment variable in your CI service. Note, old build URLs
+    will break if the slug is changed.
   </p>
+  {{#if is-admin}}
```
cadeParade: `is-admin`  is a template helper found [here](https://github.com/percy/percy-web/blob/master/app/helpers/is-admin.js) and as such, the correct syntax would be `{{#if (is-admin}}`. As written, this would never evaluate to `true`. I've pulled the branch and verified that it works with the parenthesis. This will also help with your tests :)

In case you haven't been introduced to how to turn on admin mode in the FE for testing or other uses, lmk in Slack, I'm happy to help.

app/models/project.js
ballPointPenguin: removing default value here in favor of having the API/DB set the default.

app/templates/components/forms/project-edit.hbs
ballPointPenguin: thanks for catching this

tests/acceptance/project-test.js
```diff
@@ -97,6 +109,13 @@ describe('Acceptance: Project', function() {
       await visit(`/${this.enabledProject.fullSlug}/settings`);
       await percySnapshot(this.test);
     });
+
+    it('displays Auto-Approve Branches setting', async function() {
```
cadeParade: The setup is great! But since you're switching on admin-mode for every test, this test is a bit redundant. The updates you've added show up in all three tests in this `describe` block. Two options are:
- Keep `describe` setup as-is  and remove this test, or
- Move the admin setting and clearing to before the `visit` and after the `expect` in this test.

I prefer the second option so we can both make sure the page displays as we want to current users (without the admin flag) and monitor the progress of what is happening behind the admin flag. 
Make sense?

ballPointPenguin: Yes. I like the 2nd option as well. I originally tried to be the beforeEach / afterEach inside the settings 'descrbe' block but it didn't work as I'd hoped. I'll try your suggestion.

PR #566, opened 2018-05-30
TailwindCSS

PixelJanitor:


PixelJanitor:
Yes - I knew that would be the case and empathize with you 100%. I'm relying heavily on Percy here and will continue to pull legacy styles out to replace with tailwind classes as we move forward. This initial step was going to be huge, and at 119 files it sure is true. I'll wait until the majority of the team is online today to merge this and get it on canary.

PR #567, opened 2018-05-30
Improve handling for project to build loading race condition.

fotinakis:
Followup to https://github.com/percy/percy-web/pull/557, attempt a more long-term fix for the project --> build navigation race condition. 

It seems to be about the caching from `findRecord` and the side-loading of the build browser information when navigating quickly from project list --> build. Adding `{reload: true}` to the model call seems to fix the issue and I'm unable to reproduce it with that on.

See the reloading section here:
https://www.emberjs.com/api/ember-data/release/classes/DS.Store/methods/findRecord?anchor=findRecord

Also, add tracking for how often this issue exists in prod if it still does:
https://sentry.io/percy/development-web/issues/568428879/?query=is:unresolved


app/components/project-container.js
```diff
@@ -18,7 +18,9 @@ export default Component.extend(PollingMixin, {
           .get('builds')
           .reload()
           .then(() => {
-            this.set('isRefreshing', false);
+            if (!this.isDestroyed) {
```
fotinakis: Sneaking in a fix for an error.

PR #568, opened 2018-05-31
Lazy render snapshot viewer when there are many on the screen.

cadeParade:
## Description
Rendering performance when switching browsers for builds with many snapshots to display is bad. This PR speeds up perceived performance by rendering on the screen only snapshots that are in the viewport (and some buffer below).

## Reviewing this PR
_**Is Covered by Automated Tests?: YES**ish_

To view the improvements, check out `/percy/percy-web/builds/739227` on prod and on this branch. Switching to Firefox browser is much faster.

This PR introduces `shouldDeferRendering` which is true only when there are more than 75 snapshots on a page. 
When `shouldDeferRendering` is true, snapshot-viewer components only render when they are in the viewport.

This also has some implications for the `isDefaultExpanded` thread of logic -- this was purely to manage the snapshot expansion state when there were "too many" snapshots. It also removes the "too many" snapshots banner which appeared when there were more than 150 snapshots.


app/components/snapshot-list.js
```diff
@@ -32,7 +32,7 @@ export default Component.extend({
   // Set internally by actions
   activeSnapshotId: null,
 
-  isDefaultExpanded: lt('snapshotsChanged.length', 150),
+  shouldDeferRendering: gt('snapshotsChanged.length', 75),
```
maprules1000: 👍

app/components/snapshot-viewer.js
```diff
@@ -65,13 +72,12 @@ export default Component.extend({
     this.set('_shouldScroll', true);
   }),
 
-  isDefaultExpanded: true,
   isFocus: alias('isActiveSnapshot'),
   isExpanded: or('isUserExpanded', '_isDefaultExpanded'),
   isUserExpanded: false,
 
   _isDefaultExpanded: computed(
```
maprules1000: Could this be renamed to `_shouldDeferRendering` too?

tests/integration/components/snapshot-list-test.js
```diff
@@ -37,66 +37,27 @@ describe('Integration: SnapshotList', function() {
...
       expect(SnapshotList.snapshots().count).to.equal(numSnapshots);
       SnapshotList.snapshots().forEach(snapshot => {
-        expect(snapshot.isCollapsed).to.equal(true);
+        expect(snapshot.isLazyRenderHeaderVisible).to.equal(true);
       });
-      expect(SnapshotList.isTooManySnapshotsAlertVisible).to.equal(true);
       percySnapshot(this.test);
     });
-
-    it('allows keyboard nav with up and down arrows', function() {
```
maprules1000: wow this test gets simplified a ton 🎉

app/components/snapshot-viewer.js
cadeParade: This one is a bit different, and it depends on `shouldDeferRendering`.
`isDefaultExpanded` was only allowed to be passed in because the snapshot-list would sometimes have too many to display. 
Now that snapshots have almost the same behavior regardless of how many snapshots are displayed, we don't need to pass that in.
But snapshots still need to know whether they should be expanded or not (ex: if the snapshot is approved but the build is not approved, the snapshot is collapsed), independently of the rendering.

PR #573, opened 2018-06-04
Github Marketplace Purchase

maprules1000:
## Description
This PR dampens some of the fire of purchasing (and correctly activating) a $99 plan from the github marketplace. Now users that have signed up with a user name and password will be forced to connect their github account before creating an organization.

That way we will properly connect the github purchase to the correct organization for billing purposes.

This is also a bandaid until the signup flow is redesigned.

Want to check it out locally? Pull down this branch and use this link: http://dev.percy.local:4200/organizations/new?marketplace_listing_plan_id=9

## Reviewing this PR
_**Is Covered by Automated Tests?: YESish**_

There are tests in place, but manual testing is required to know that all the parts are working together (`percy-web`, `percy-api`, auth0, and github 😬 ).

### Manual Testing Steps

**note: Don't use the production backend for this. Use `percy-api` locally and make sure it's up to date.**

1. Open up a browser you don't use, or an incognito window. You should be logged out of Percy.
2. You'll need a test account for ~~Team Foundation Server~~ Github. A test account is important because we're going to unlink Percy from that account. It'd be a headache to do it from your main account.

3. Use this url to start the process: http://dev.percy.local:4200/organizations/new?marketplace_listing_plan_id=9 This is the same url github provides when a user makes a marketplace purchase. Remember, you should be signed out of Percy.
4. Setup a new account with Percy using an email and password.
5. Verify your email **in the same browser session** (ie in the same browser or incognito browser window).
6. Log in with your newly created user

### The important part of the manual testing
7. You should now see the org creation page with the connect to github section and the create org section below it, greyed out.
8. Connect your github account.
9. You should be redirected back to the org creation page with the github section greyed out and your github user name displayed with a green check.
10. You should be able to create an org now.

### Before
screen shot 2018-06-05 at 2 25 02 pm

### After (GitHub not connected)
screen shot 2018-06-07 at 2 03 42 pm

### After (GitHub Connected)
screen shot 2018-06-07 at 2 03 50 pm

### After (Org Name Entered)
screen shot 2018-06-07 at 2 04 08 pm


cadeParade:
Initial thought from reading the PR description: Can we make an e2e test for this?

app/components/organizations/new-organization.js
```diff
@@ -0,0 +1,40 @@
+import Component from '@ember/component';
+import EnsureStatefulLogin from 'percy-web/mixins/ensure-stateful-login';
+import {computed} from '@ember/object';
+import {inject as service} from '@ember/service';
+
+export default Component.extend(EnsureStatefulLogin, {
+  session: service(),
+
+  githubMarketplacePlanId: null, // required
```
cadeParade: Is the `required` comment correct? Users who didn't come from github marketplace should also be able to create new organizations.

app/components/organizations/new-organization.js
```diff
@@ -0,0 +1,40 @@
...
+
+  currentUser: computed.alias('session.currentUser'),
+  githubIdentity: computed.alias('currentUser.hasGithubIdentity'),
+  hasGithubIdentity: computed.alias('currentUser.hasGithubIdentity.isTruthy'),
+  isGithubPurchase: computed.notEmpty('githubMarketplacePlanId'),
+
+  githubSetupState: computed('isGithubPurchase', 'hasGithubIdentity', function() {
+    if (!this.get('isGithubPurchase')) {
+      return 'is-invisible';
```
cadeParade: Hmm, I don't love mixing these class strings among the JS. Generally we should try to keep the template stuff in the template. Could we make a property that's named appropriately so in the template we could do something like `{{if notGithubPurchase "is-invisible"}}`, etc?

app/components/organizations/new-organization.js
```diff
@@ -0,0 +1,40 @@
...
+    } else if (this.get('isGithubPurchase') && this.get('hasGithubIdentity')) {
+      return 'opacity-25 pointer-events-none';
+    }
+  }),
+
+  needsGithubIdentity: computed('isGithubPurchase', 'hasGithubIdentity', function() {
+    // false if not GH purchase or gh purchase without connected gh account
+    if (this.get('isGithubPurchase') && !this.get('hasGithubIdentity')) {
```
cadeParade: Spitballing --
`this.get('isGithubPurchase') && !this.get('hasGithubIdentity') ? true : false` ?
`return this.get('isGithubPurchase') && !this.get('hasGithubIdentity')` ?

app/components/organizations/new-organization.js
maprules1000: Good catch! I think I threw that in early on. Will remove

app/components/organizations/new-organization.js
maprules1000: I'll get with Derek and discuss. I threw it into js because there's 3 states that can be rendered. We'll either remove a state or I'll figure out a way to put it into the template entirely

maprules1000: Spoke with derek and I'm going to push these states into the template. We actually don't use the `is-invisible` state anymore so it should be pretty simple.

app/components/form-fields/input.js
```diff
@@ -16,4 +16,11 @@ export default Component.extend({
   fieldErrors: computed('changeset.error', function() {
     return get(this.get('changeset.error'), this.get('property'));
   }),
+
+  didUpdateAttrs() {
+    // if autofocus changes and is true then the input needs to be manually focused
+    if (this.get('autofocus')) {
+      document.getElementsByClassName('form-control')[0].focus();
+    }
```
maprules1000: `autofocus=true` only works on initial page render, and does not update focus when `autofocus` subsequently changes. This provides a trigger to focus on the input when `autofocus` changes to `true`.

PR #575, opened 2018-06-06
Rename visual validations to screenshots

timhaines:
Our pricing and billing pages currently include language around visual validations.  We've decided to swap this to screenshots, as it's easier to understand.


timhaines:
Going to hold on this until next week, and chat with Payton and Mike again before shipping.

PR #579, opened 2018-06-08
Quick fix for autofocus issue

maprules1000:
Quick fix for autofocus that gets it working again


maprules1000:
The input box wasn't being autofocused at all

PR #588, opened 2018-06-11
[WIP] Changelog

PixelJanitor:
Current state:

image


app/templates/changelog.hbs
```diff
@@ -0,0 +1,149 @@
+{{title "Team"}}
```
cadeParade: This should be something else.

PixelJanitor: Haha why yes... yes it should be what the page actually is and not what it was duplicated from. Thanks lol

PR #589, opened 2018-06-13
Upgrade ember to 3.1.0

maprules1000:
## Description
Uses `ember-cli-update` to upgrade ember from 3.0.0 to 3.1.0.

## Reviewing this PR

There should be no issues with this upgrade. The version bumps are minimal and are mostly full of bugfixes. However, this should be tested in canary and with end-to-end tests since it does contain an `ember-data` update.

_Update_: Looks good in end-to-end tests.


cadeParade:
I don't think we should disable ember-view div wrapper for the initial push of 3.1. Let's ship 3.1 without that flag on, and let @PixelJanitor work on updating the app to accomodate no component divs as a separate project.

PixelJanitor:
I thought that's what the game plan was all along :)

maprules1000:
Just pushed it up to see if there were any visual changes real quick. It has been removed

PR #595, opened 2018-06-15
Add Gitlab Merge Request Support

faun:
##  [Clubhouse Card](https://app.clubhouse.io/percy/story/2105/add-gitlab-merge-requests-links-and-logo-when-appropriate)

See also: [API #490](https://github.com/percy/percy-api/pull/490)

## Description

In order to fully support GitLab, show Gitlab logo instead of Github logo on the build cards and project container. Links to GitLab merge requests should link to the correct place.

**Note**: This requires https://github.com/percy/percy-api/pull/490 running on the backend for this to work properly.

## Pictures

faun_gitlab-quickstart

## Reviewing this PR

I'm looking for Ember patterns and testing best practices in this PR. Do these integration tests make the most sense here? Would a different level of testing be more appropriate?


app/models/repo.js
```diff
@@ -11,4 +13,18 @@ export default DS.Model.extend({
   description: DS.attr(),
   createdAt: DS.attr('date'),
   updatedAt: DS.attr('date'),
+
+  isGithubRepo: computed('source', function() {
```
cadeParade: There is an ember computed macro for this `isGithubRepo: equal('source', 'github')`. The tricky part about this is that the second argument _must_ be a literal value, but in these cases it is! 
https://www.emberjs.com/api/ember/2.16/functions/@ember%2Fobject%2Fcomputed/equal

app/models/project.js
```diff
@@ -16,6 +16,10 @@ export default DS.Model.extend({
 
   // Repo will be set if this project is linked to a repository.
   repo: DS.belongsTo('repo', {async: false}),
+  isGithubRepo: alias('repo.isGithubRepo'),
```
cadeParade: How is behavior of this feature affected if the project does not have a repo?

tests/integration/components/build-card-test.js
```diff
@@ -0,0 +1,136 @@
+import {setupComponentTest} from 'ember-mocha';
```
cadeParade: Yesss, This file has been a long time coming 🎉

app/templates/components/build-card.hbs
```diff
@@ -8,25 +8,29 @@
     </div>
   </div>
   <div style="min-width: 0">
-    <div class="flex items-center mb-sm pt-sm">
-      {{#if build.isGithubLinked}}
+    <div class="flex items-center mb-sm pt-sm" data-test-build-card-source-code-metadata>
```
cadeParade: Optional: If we think we'll be choosing an icon based on integration source a lot (seems likely) we can consider writing a template helper that takes a build or project and returns an icon name or class.

app/models/repo.js
faun: Sweet!

faun: 🎉

app/models/repo.js
```diff
@@ -11,4 +12,9 @@ export default DS.Model.extend({
   description: DS.attr(),
   createdAt: DS.attr('date'),
   updatedAt: DS.attr('date'),
+
+  isGithubRepo: equal('source', 'github'),
+  isGithubEnterpriseRepo: equal('source', 'github_enterprise'),
+  isGithubRepoFamily: or('isGithubRepo', 'isGithubEnterpriseRepo'),
+  isGitlabRepo: equal('source', 'gitlab'),
```
maprules1000: since the values for the 'source' are strings that come down from the api in a specifically formatted way they should be imported in from the `version-control-integration` model (e.g. https://github.com/percy/percy-web/blob/master/app/routes/organizations/organization/integrations/index.js#L4)

It'll be abstracted out to a better place in the future but for now it'd be good to import it so it's easy to keep track of where we're using those magic strings that must match the api.

tests/integration/components/build-card-test.js
```diff
@@ -0,0 +1,136 @@
...
+import {expect} from 'chai';
+import {it, describe, beforeEach} from 'mocha';
+import {percySnapshot} from 'ember-percy';
+import hbs from 'htmlbars-inline-precompile';
+import {make} from 'ember-data-factory-guy';
+import BuildCard from 'percy-web/tests/pages/components/build-card';
+import setupFactoryGuy from 'percy-web/tests/helpers/setup-factory-guy';
+
+describe('Integration: BuildCard', function() {
```
maprules1000: Yay tests! 🎉

app/models/repo.js
faun: Fixed in 852976b5

faun: I'm really happy with the mechanics of this.

PR #602, opened 2018-06-21
Make integration tests deterministic

faun:
While seeing random company names with new builds is fun, it makes it hard to determine what has changed. This PR seeds faker before running my recent integration tests additions.


cadeParade:
Does this work? `setupFactoryGuy` is also called from your tests and also calls `seedFaker`. Why is your test disobeying `setupFactoryGuy` `seedFaker`?

tests/integration/components/build-card-test.js
```diff
@@ -6,13 +6,15 @@ import hbs from 'htmlbars-inline-precompile';
 import {make} from 'ember-data-factory-guy';
 import BuildCard from 'percy-web/tests/pages/components/build-card';
 import setupFactoryGuy from 'percy-web/tests/helpers/setup-factory-guy';
+import seedFaker from '../../helpers/seed-faker';
```
cadeParade: You can/should use `percy-web/tests/helpers/...` instead of a relative path.

faun:
This PR does not work as expected.

PR #606, opened 2018-06-22
refresh lockfile

kellyselden:
I would like to use your app to test my new packager, but first, you need some updated dependencies of dependencies that I fixed, so I regenerated your lockfile to get latest everything.


fotinakis:
Thanks @kellyselden! There was a broken test due to a prettier update, I've started PR #607 based on this PR and will get that merged in.

PR #607, opened 2018-06-22
Refresh lockfile and upgrade all packages.

fotinakis:
Supersede #606 and fix a style issue.


kellyselden:
This appears to have gone stale.

PR #608, opened 2018-06-23
Add GitLab personal access tokens to integrations page

faun:
Add Gitlab personal access token settings to unlock self-service gitlab installations.

[Clubhouse Story](https://app.clubhouse.io/percy/story/2025/add-gitlab-personal-access-token-to-gitlab-integrations-page)

image


app/components/organizations/integrations/gitlab-settings.js
```diff
@@ -5,11 +5,13 @@ import Component from '@ember/component';
 
 export default Component.extend({
   organization: null,
+  model: alias('gitlabIntegration'),
```
cadeParade: Renaming `gitlabIntegration` to model is a bit unnecessary IMO. Models usually refer to things that come from routes which this is not. I think you are renaming it to `model` to pass into the form in the template, but why not in the template do `model=gitlabIntegration` and then we know what we're dealing with because it is a named thing instead of generic "model".

app/styles/components/forms/_gitlab_settings_edit.scss
```diff
@@ -0,0 +1,36 @@
+.GitlabSettingsEdit {
```
cadeParade: This seems like a lot of custom css for a fairly simple form. @PixelJanitor can you suggest any ways it could be simplified with our new fancy tailwind addon?

app/templates/components/forms/gitlab-settings-edit.hbs
```diff
@@ -0,0 +1,46 @@
+<form action="javascript:;" class="mb-0" data-test-gitlab-settings-edit-form>
+  <div class="flex items-center">
+    <div class="grid col-auto gap-1 text-xl font-semibold mb-2">
+      <div>{{inline-svg "gitlab-icon-lg" class="mr-1"}}</div>
+      <div><h2 class="text-xl font-semibold m-0">GitLab Integration</h2></div>
+    </div>
+  </div>
+  <div class="p-2 bg-gray-000 rounded border mb-2">
+    {{form-fields/input
+        data-test-gitlab-personal-access-token-field
```
cadeParade: p999999 I think this is too many tabs

cadeParade: Does this selector work? Usually when passing into components like this it needs to be `data-test-whatever=true`. But I see it in your tests, so I guess it works...

EDIT: I think this works because of https://github.com/simplabs/ember-test-selectors

app/templates/components/organizations/integrations/github-enterprise-settings.hbs
```diff
@@ -1,7 +1,7 @@
 <section class="px-2 pt-3">
   <div class="w-3/4 pb-3">
     <h1 class="text-xl font-semibold mb-sm">Set up Github Enterprise</h1>
-    <p class="text-gray-400">We’ll need just a bit of information to get Percy set up on your GitHab account. If you need some help, check out our <a href="#" class="font-semibold">Percy + GitHub</a> doc to guide you through the set up process.</p>
+    <p class="text-gray-400">We’ll need just a bit of information to get Percy set up on your GitHub account. If you need some help, check out our <a href="#" class="font-semibold">Percy + GitHub</a> doc to guide you through the set up process.</p>
```
cadeParade: lmao GitHab 😂

app/templates/components/organizations/integrations/gitlab-settings.hbs
```diff
@@ -1,18 +1,20 @@
-<section class="p-2">
-  <div class="bg-white rounded shadow-1 p-2 mb-2" data-test-gitlab-settings>
-    <div class="flex items-center mb-2">
-      {{inline-svg "gitlab-icon-lg" class="mr-1"}}
-      <h2 class="text-xl font-semibold m-0">GitLab Integration</h2>
-    </div>
+<section class="px-2 pt-3">
+  <div class="w-3/4 pb-3">
+    <h1 class="text-xl font-semibold mb-sm">Set up GitLab</h1>
+    <p class="text-gray-400">We’ll need just a bit of information to get Percy set up on your GitLab account. If you need some help, check out our <a href="#" class="font-semibold">Percy + GitLab</a> doc to guide you through the set up process.</p>
```
cadeParade: This link is empty.

tests/acceptance/gitlab-settings-test.js
```diff
@@ -0,0 +1,46 @@
...
+
+describe('Acceptance: Gitlab Settings', function() {
+  setupAcceptance();
+
+  describe('with a standard user', function() {
+    setupSession(function(server) {
+      let organization = server.create('organization', 'withUser', 'withGitlabIntegration');
+      let project = server.create('project', {organization});
+      this.organization = organization;
+      this.project = project;
```
cadeParade: `this.project` is not used in the `it` blocks of either test so it does not need to be assigned to the text context.
Also, a preferred pattern is:

```
let organization;
setupSession(function(server) {
   organization = whatever
});
```

Using this pattern would also require you to use the variables you define because ESLint will fail if you have unused variables.

tests/integration/components/organizations/integrations/gitlab-settings-test.js
```diff
@@ -30,17 +30,32 @@ describe('Integration: GitLab Settings', function() {
...
 
-    it('shows that the integration is installed', function() {
+    it('shows the settings form', function() {
       this.render(hbs`{{
         organizations/integrations/gitlab-settings
         currentUser=user
         organization=organization
       }}`);
-      expect(GitlabSettings.integrationMessage).to.include('is installed');
+      expect(GitlabSettings.integrationSettings.personalAccessTokenField.isVisible).to.eq(true);
```
cadeParade: Just want to make you aware of this property for future reference because chaining these properties 4 levels deep will get hard to maintain -- there is an `alias` property in page objects: http://ember-cli-page-object.js.org/docs/v1.14.x/api/alias (actually Anna Andresian wrote it :)) which allows you to alias other page object properties. It is helpful when you need to reference something deep like this, you can put a property `isPersonalAccessTokenFieldVisible: alias('integrationSettings.personalAccessTokenField.isVisible')` on `gitlabSettings` page object. and have your test read shorter. This also has the advantage that if the structure of the integration settings page changes and therefore the structure of the page object, the test shouldn't have to change, you would just have to re-alias the `isPersonalAccessTokenFieldVisible` property.

app/styles/components/forms/_gitlab_settings_edit.scss
PixelJanitor: Yep! Working on this now

app/components/forms/base.js
```diff
@@ -92,7 +92,16 @@ export default Component.extend({
       }
     },
     delete() {
-      this.get('model').destroyRecord();
+      let store = this.get('store');
+      let model = this.get('model');
+      // Delete the record on the server
+      model.destroyRecord().then(
```
fotinakis: I'd probably write this as:

```
model.destroyRecord().then(() => store.unloadRecord(model));
```

app/components/organizations/integrations/gitlab-settings.js
```diff
@@ -1,22 +1,64 @@
...
+  }),
 
   actions: {
-    redirectToGitlabAuthorization(organization) {
-      let slug = organization.get('slug');
-      if (slug) {
-        utils.setWindowLocation(`/api/auth/gitlab/redirect/${slug}`);
-      }
+    createGitlabIntegration(organization) {
+      let org = this.store.peekRecord('organization', organization.id);
```
fotinakis: Why not just `this.get('organization')`?

tests/integration/components/organizations/integrations/gitlab-settings-test.js
```diff
@@ -30,49 +28,60 @@ describe('Integration: GitLab Settings', function() {
     beforeEach(function() {
       const user = make('user');
       const organization = make('organization', 'withGitlabIntegration');
+      const organizationUser = make('organization-user', 'adminUser', {
```
fotinakis: Can just do `{organization, user}` here to shorten this in crazy ES6 style.

faun:
I'm leaving the CSS as-is until @PixelJanitor and I get a chance to pair on this. There are some unintended side-effects that we need to work through, and I think shipping this and iterating on it to simplify is the right approach here.

PR #609, opened 2018-06-25
Custom Auth0 domain (development only)

djones:
## Description

This PR adds Auth0 custom domain support to development only. This is so we can make some changes to our "Percy [DEV]" OAuth application and test this in dev, while allowing other engineers to have a working local login.

Technical docs: https://auth0.com/docs/custom-domains/additional-configuration#embedded-lock

## Reviewing this PR

Our custom domain for this Auth0 dev tenant is configured here:

screen shot 2018-06-25 at 11 04 28 am

## Post-Merge

Once deployed, our "Percy [DEV]" OAuth application needs to have its "Authorization callback URL" changed to "https://login-dev.percy.io/login/callback"

## Coming Up

The same change but with prod will be coming soon.


app/lib/lock-settings.js
```diff
@@ -1,4 +1,5 @@
 export default {
+  configurationBaseUrl: 'https://cdn.auth0.com',
```
djones: This will get rolled out to everyone (not just dev), but is safe and compatible with our current setup.

PR #613, opened 2018-06-26
Cross-browser updates

cadeParade:
## Description
Based on slack conversations, a few updates:
- Add green flash message when a browser is changed
- Add a link to docs in text above browser switcher buttons
- add “-” to “cross browser”

## Pictures
image 5
image 6


ballPointPenguin:
The flash message appears at the foot of the page on mobile-sized chrome but in the middle of mobile-sized firefox. I don't think that is a problem. But I do find it curious. https://percy.io/percy/percy-web/builds/819813?utm_campaign=percy&utm_content=percy-web&utm_source=github_status

maprules1000:
@ballPointPenguin that's an interesting catch. It's almost like firefox is rendering it at the bottom of some viewport instead of the bottom of the page.

cadeParade:
Thanks for catching that @ballPointPenguin . It is quite curious. I also can't reproduce this positioning locally, but using the firefox screenshot feature does reproduce it. It is reminiscent of this intercom issue https://app.intercom.io/a/apps/m37fs4zj/inbox/inbox/search/conversations/16873405025?q=footer. I think we won't fix it right now but I'll open a ticket to document these two cases.

PR #619, opened 2018-06-28
Browser background fix

PixelJanitor:


djones:
Annnnnd breathe out. That was a pretty interesting one!

PR #624, opened 2018-07-03
Remember destination url throughout login over two tabs

cadeParade:
## Description
If there is a scenario where:
1) User is logged out of percy.io AND
1a) User opens two non-home percy pages in two separate tabs
2) User logs in with the lock modal in one tab
3) Tab which the user logged into redirects correctly, second tab authenticates but redirects to home.

This PR makes step 3 authenticate and redirect properly in both tabs by 
1) Using `sessionStorage`(storage per tab) instead of `localStorage` to store redirect url.
2) Closing lock instance after authentication.

## Reviewing this PR
Lock is impossibly to test with Ember tests. We could add an e2e test though.


cadeParade:
Thanks so much for catching this. Great work. 

The solution above is not quite right -- we can't automatically redirect to home all the time, because say, if you're not logged in and on the docs page, you click "sign in" in top right, it opens the lock modal but doesn't change routes. If you close the modal with the code above, it redirects to home, where it should actually stay on docs route and close the modal there. 
This is the reason why I made that optional param `onCloseDestinationRoute` and only pass it in through he login route. That's the only route we want to do something different when the modal closes.

So I've accomplished this with a variation of the code snippet above (https://github.com/percy/percy-web/pull/624/files#diff-e47ce221b68bb71203191dfa77013a3bR74). 
I've also added `login` route to the "do not store" routes so if a user goes straight to `/login` and closes the modal, it doesn't redirect them back to `/login`. 

My gripe with this approach is a real stupid thing: importing `AUTH_REDIRECT_LOCALSTORAGE_KEY` from application route into ensure-stateful-login-mixin broke allllll the tests (ex: https://buildkite.com/percy/percy-web/builds/1780#529ff405-b16b-4844-9653-58dc533eda24/20-989) . I'm guessing this is related to importing something from a place the mixin is mixed in to? But I am not really sure why this is. So I moved it to `router`....I figured that was the second best place for it since it points to routes. This may be another pointer towards the redirect service you mentioned yesterday; I'm interested to hear your thoughts on that.

I'll be out the rest of the week so feel free to take this over the finish line if you want, or I'll just finish/merge next week.

app/router.js
```diff
@@ -2,16 +2,19 @@ import {on} from '@ember/object/evented';
 import EmberRouter from '@ember/routing/router';
 import config from './config/environment';
 
+export const AUTH_REDIRECT_LOCALSTORAGE_KEY = 'percyAttemptedTransition';
```
maprules1000: I'm not crazy about this, especially since this value isn't actually used in this file

maprules1000:
I have a thought about the pattern of declaring and importing constants in percy-web: what if we created a service that we stashed all the constants in and just imported it when they're needed? Something like https://medium.com/@quangtn0018/ember-js-set-up-constants-e3f31f5b8927 

We currently have constants everywhere, and it gets difficult to keep track of them. Stashing them all in `environment.js` isn't ideal either since there are many constants and they're not env specific. 

We can talk about this more and deal with it in a follow up PR, but we should clean up the constants pattern.

cadeParade:
Agree, we should have a different constants pattern. Injecting constants into everything automatically sounds potentially useful. But we don't want to be able to change constants. One thing that might work is use a JS Proxy object to define a custom `getter` and `setter` (but mostly `setter`) that will throw an error if you try to set a value.

```
  if (ENV.environment && ENV.environment !== 'production') {
    return new Proxy (allOurConstantsHash, {
      set(target, name) {
          Ember.assert(`Do not try to assign values to constants`);
      },
    });

  // If we're in production environment, export the POJO only
  // as Proxys are not available in some old browsers
  } else {
    return allOurConstantsHash;
  }
```

PR #627, opened 2018-07-03
[WIP] Proof of concept for docs redirect

cadeParade:
## Description
We're moving the docs to https://docs.percy.io/docs and need to redirect users from old urls to new ones.

To do this, intercept all requests to the docs route and block the application from continuing to develop and redirect to the corresponding new docs page. The new docs pages don't map 1:1 with the old routes, so each docs address will need to be looked up.

## Reviewing this PR


cadeParade:
Closed for https://github.com/percy/percy-web/pull/639

PR #636, opened 2018-07-12
Add simple notice of browser family upgrade.

fotinakis:
##  https://app.clubhouse.io/percy/story/2211/notification-on-build-when-current-browser-is-different-from-the-base-build-browser

A simple notification when the build has any browsers that have been upgraded, to clarify why "New Snapshots" are shown.

screen shot 2018-07-12 at 12 18 05 pm

Still needs:
- [ ] tests


app/models/build.js
```diff
@@ -193,6 +193,22 @@ export default DS.Model.extend({
     return this.store.findRecord('build', this.get('id'), {reload: true});
   },
 
+  browsersUpgraded: computed('browsers.[]', 'baseBuild.browsers.[]', function() {
+    const headBuildBrowsers = this.get('browsers') || [];
+    const baseBuildBrowsers = this.get('baseBuild.browsers') || [];
+    const browsersUpgraded = [];
+    headBuildBrowsers.forEach(headBrowser => {
```
fotinakis: Usually I would try hard to avoid writing a nested for-loop, but there are only ever like 2-4 things to process here so it's much clearer to write it this way.

app/routes/verify-email.js
```diff
@@ -7,7 +7,8 @@ export default Route.extend({
     code: '',
   },
   model(params) {
-    return $.ajax({type: 'PATCH', url: utils.buildApiUrl('emailVerifications', params.code)})
+    return $
```
fotinakis: Fixing an ongoing lint error forever.

app/templates/components/build-header.hbs
```diff
@@ -1,99 +1,116 @@
...
-    <div class="p-2 m-2 bg-gray-000 rounded border border-gray-100">
-      <div class="font-semibold mb-1">Your CI service or client failed to correctly upload build and snapshot resources.</div>
-      This build can be safely ignored. It will not be used as a baseline for visual diffs. If these failures happen frequently, check your CI or client logs for failures.
-    </div>
-  {{/if}}
-  {{#if (eq build.failureReason 'no_snapshots')}}
-    <div class="p-2 m-2 bg-gray-000 rounded border border-gray-100">
-      <div class="font-semibold mb-1">No snapshots were uploaded to this build, so there was nothing to compare. This could happen if your tests failed in your CI environment and nothing was uploaded to Percy.</div>
-      This build can be safely ignored. It will not be used as a baseline for visual diffs. If these failures happen frequently, check your CI or client logs for failures.
+  {{#if build.browsersUpgraded}}
```
fotinakis: This if-statement is the only thing really added to this file, everything else is just indented.

app/templates/components/build-header.hbs
```diff
@@ -1,99 +1,116 @@
...
-    </div>
-  {{/if}}
-  {{#if (eq build.failureReason 'no_snapshots')}}
-    <div class="p-2 m-2 bg-gray-000 rounded border border-gray-100">
-      <div class="font-semibold mb-1">No snapshots were uploaded to this build, so there was nothing to compare. This could happen if your tests failed in your CI environment and nothing was uploaded to Percy.</div>
-      This build can be safely ignored. It will not be used as a baseline for visual diffs. If these failures happen frequently, check your CI or client logs for failures.
+  {{#if (and build.isFinished build.browsersUpgraded)}}
+    <div class="p-2 bg-gray-000 border border-gray-100">
+      <strong>
+        {{#if (gt build.browsersUpgraded.length 1) }}
```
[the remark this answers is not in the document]
fotinakis: I don't think you can in this case, because it's not actually pluralizing any single string.

PR #640, opened 2018-07-18
Use stripe elements for subscription update

cadeParade:
## Description
We want to remove stripe-checkout and associated flows and use stripe-elements instead. @maprules1000 laid a great foundation for this with the `billing-card-updater` component, and here I've moved most of this logic into a mixin so it can be reused in the context of the subscription items.

### Main points
- Moved card updating logic from `billing-card-updater` to mixin
- Add logic to subscription-list-item to:
  - Expand subscription list item to show a credit card if no billing information on file
    - fill out credit card info and submit
    - cancel out of transaction with cancel button
  - Automatically change plan if billing information exists
  -  redirect to custom plan page if clicking "custom" button
- Remove stripe-checkout
- Write some integration tests that test up until the boundary of where it interacts with stripe

## Pictures
subscription-stripe-elements

## Reviewing this PR

_**Is Covered by Automated Tests?: YESISH**_


app/templates/components/organizations/billing-card-updater.hbs
```diff
@@ -4,17 +4,19 @@
...
     {{#if stripeError}}
       <p class="error">{{stripeError.message}}</p>
     {{/if}}
     <div class="flex justify-end mt-2">
-      <button class="btn btn-primary btn-shadow-purple" {{action 'updateCreditCard' stripeElement}} disabled={{if (not shouldShowSubmit) "disabled"}}>Update Card</button>
+      <button class="btn btn-primary btn-shadow-purple" {{action 'updateCreditCard' stripeElement planId}} disabled={{if (not shouldShowSubmit) "disabled"}} data-test-submit-card-button>Update Card</button>
     </div>
   {{/stripe-card}}
 {{else}}
-  <button class="btn btn-primary btn-shadow-purple" onclick={{action 'showCardInput'}}>Update Credit Card</button>
```
maprules1000: 🙌

app/templates/components/organizations/subscription-list-item.hbs
```diff
@@ -30,30 +30,39 @@
...
       </div>
       <div style="max-width: 150px" class="flex-1 pr-2 flex items-start justify-end">
-        {{#if (eq plan.id planData.id)}}
-          {{#if isSaving}}
-            {{saving-indicator isSaving=isSaving isSaveSuccessful=isSaveSuccessful}}
-          {{else}}
-            <a href="#" disabled="disabled" class="btn btn-primary shadow-purple-lg mb-0">Selected Plan</a>
-          {{/if}}
+
+        {{#if shouldShowCancel}}
```
maprules1000: should something be here if `shouldShowCancel` is true?

tests/integration/components/organizations/subscription-list-item-test.js
```diff
@@ -0,0 +1,174 @@
...
+    name: 'Business',
+    monthlyPrice: 849,
+    numDiffs: 200000,
+    extraDiffPrice: 0.006,
+    numTeamMembersTitle: '20 team members',
+    numWorkersTitle: '40 concurrent renderers',
+    historyLimitTitle: '1 year history',
+  };
+
+  const customPlanData = {
```
maprules1000: Should these be factory guy traits?

cadeParade: Well, these are taking the place of the items in subscription data service which are not actual models, but rather POJOs, so I chose not to put them in factory guy. Whether the subscription data service should be POJOs rather than plan models is another question. But this PR needs to stop somewhere so I drew the line. 😢

app/templates/components/organizations/subscription-list-item.hbs
cadeParade: Good call. `shouldShowCancel` is no longer necessary due to some previous refactoring so I'll remove it.

app/templates/components/organizations/billing-card-updater.hbs
```diff
@@ -4,17 +4,19 @@
...
+    data-test-billing-card-updater-stripe-card
     options=options
     change=(action 'checkCard')
     as |stripeElement stripeError|}}
     {{#if stripeError}}
       <p class="error">{{stripeError.message}}</p>
     {{/if}}
     <div class="flex justify-end mt-2">
-      <button class="btn btn-primary btn-shadow-purple" {{action 'updateCreditCard' stripeElement}} disabled={{if (not shouldShowSubmit) "disabled"}}>Update Card</button>
+      <button class="btn btn-primary btn-shadow-purple" {{action 'updateCreditCard' stripeElement planId}} disabled={{if (not shouldShowSubmit) "disabled"}} data-test-submit-card-button>Update Card</button>
```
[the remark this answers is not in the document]
cadeParade: Yes, but also this section does not show up unless the viewer is already a customer (i.e. already has a card on file) https://github.com/percy/percy-web/blob/master/app/templates/components/organizations/billing-section.hbs#L91

PR #645, opened 2018-07-25
Fix or silence warnings in the browser console

cadeParade:
## Description
This PR fixes several warnings that appear in the console.
- A new "unsafe" css binding was introduced with the billing page graph. This is now "safe".
- ~ember-moment is unhappy when being called with no args (aka `moment()`) so I've added a global parameter to allow this based on [their documentation](https://github.com/stefanpenner/ember-moment#global-allow-empty-dates).~ It turns out that our test factories are weirdly dependent on whatever inconsistent behavior `allowEmpty: false` allows. There were many percy diffs related to this change, so I've taken it out for now.
- I left a `data-test` attribute on a component with no `tagName`, so I've fixed that.

### ADDITIONALLY
The two types of warnings that our app is absolutely riddled with: 
- `You pushed a record of type 'project' with a relationship 'organization' configured as 'async: false'. You've included a link but no primary data, this may be an error in your payload.`
- ` You requested a record of type 'organization' with id 'percy' but the adapter returned a payload with primary data having an id of '2'. Use 'store.findRecord()' when the requested id is the same as the one returned by the adapter. In other cases use 'store.queryRecord()' instead`
ARE SILENCED.

A build page on Percy goes from 242 warnings down to 2 warnings (both about intercom and are probably related to my browser extensions).

As a bonus , the test output on buildkite is much more concise. (Sadly a bit less concise now that the moment warning is back :( ) (https://buildkite.com/percy/percy-web/builds/1990#e645e0ef-7a97-4b87-8a96-f6ca933e49c5)

### Questions:
- Do we have any objections to silencing the ember-data warnings? They are not fixable without a large refactor in the backend related to sending ids vs slugs. 


app/app.js
```diff
@@ -2,13 +2,24 @@ import Application from '@ember/application';
...
 import config from './config/environment';
+import {registerWarnHandler} from '@ember/debug';
 
 const App = Application.extend({
   modulePrefix: config.modulePrefix,
   podModulePrefix: config.podModulePrefix,
   Resolver,
 });
 
+registerWarnHandler((message, options, next) => {
```
maprules1000: I think this is fine for now, and it'll clean up buildkite logs in a major way, but I think it needs to be documented somewhere other than just here since these might end up preventing us from upgrading ember data at some point.

Maybe a clubhouse card, or adding a detailed comment in `deprecation-worflow.js` which is where we silence deprecations (not warnings)?

app/components/organizations/billing-usage-graph.js
```diff
@@ -14,11 +15,14 @@ export default Component.extend({
...
-      date: dayStat[0],
-      count: dayStat[1],
-      percentOfHeight: (dayStat[1] / yAxisCeiling) * 100,
-    }));
+    return dayStats.map(dayStat => {
+      const barHeight = (dayStat[1] / yAxisCeiling) * 100;
+      return {
+        date: dayStat[0],
+        count: dayStat[1],
+        percentOfHeight: htmlSafe(`--bar-height: ${barHeight}%`),
```
maprules1000: 👍

PR #648, opened 2018-07-25
Update billing page to have new plan titles

cadeParade:
## Description
The billing page needs to reflect the new titles from the update pricing page (Small, Medium, Large instead of Starter, Growth, Business). 
Since there are two extra tiers on the pricing page, but it doesn't make sense to put two extra rows on the billing page, I've changed the last row to be a generic "get more information" row.

Once that idea changed, it no longer made sense for the last row to be an actual `subscription-list-item` since the only thing in common was the title and button. So the remaining changes are moving that row out of the `subscription-list-item` component and updating tests.
The last row in `subscription-list` also now opens the support widget rather than transitioning to a page.

## Pictures


app/components/organizations/subscription-list-item.js
```diff
@@ -38,10 +32,7 @@ export default Component.extend(BillingUpdater, {
 
   actions: {
     handleSubscriptionSelection() {
-      this.get('organization.subscription');
-      if (this.get('isCustom')) {
```
maprules1000: so is there no longer an `isCustom` check?

app/templates/components/organizations/subscription-list.hbs
```diff
@@ -61,22 +61,23 @@
...
+          <button href="#" class="btn btn-primary shadow-purple-lg mb-0" disabled={{eq plan.id planData.id}} {{action "openIntercom"}} data-test-subscription-list-choose-enterprise>
+            Contact Us
+          </button>
+        {{/if}}
+      </div>
+    </div>
   </div>
 </div>
 
+
```
maprules1000: probs can remove this new line

app/components/organizations/subscription-list-item.js
cadeParade: `isCustom` was there only for the last row, which is now removed. The rows in subscription-list now all draw from data in `subscription-data` service.

cadeParade:
Ah, I see your "custom" point, which is larger than my implementation response above. Yes, we should not allow enterprise customers to downgrade without talking to support, however, that is not currently the way the site works, so it has also not changed in this PR. That line in the code was only to support having the last "custom" line in the subscription list that wouldn't trigger an automatic billing update. 

I am currently working on an update that will hide the subscription list totally from custom plan customers (See https://www.figma.com/file/Mo9dl2q8ErZNDdQLE79GN43z/Billing?node-id=0%3A1 -- see "Enterprise" board. This will do what you mentioned above :D

PR #649, opened 2018-07-26
Introduce Launch Darkly

cadeParade:
## Description
Allow use of launch darkly flags in `percy-web`.

*Q: Do we care about anonymous users?*
Launch darkly suggests giving each anonymous user a unique ID to do A/B tests or otherwise track them across the site. Not sure if we need that right now.

*Q: What should we use for user key?*
Currently set to `userHash` property which comes from here: https://github.com/percy/percy-api/blob/e158fed0b7d99d8c4fdc62ef26f81b329d2dfc50/app/models/percy/user.rb#L46
Is this reasonable to reuse for this purpose?

*Q: Do we need to hide any user attributes from Launch Darkly? (Name, email, org id, etc)*
https://docs.launchdarkly.com/v2.0/docs/private-user-attributes
If so, we'll need to find a different addon/roll our own.

Check out this user: https://app.launchdarkly.com/default/test/users/4705248ea83a67cc4f4a776c7f828b5f06854551b0187c46df24702fc20ed993 for the type of data we're looking at for the PR.


app/routes/application.js
```diff
@@ -29,6 +32,32 @@ export default Route.extend(ApplicationRouteMixin, EnsureStatefulLogin, {
     return this._loadCurrentUser();
   },
 
+  _loadLaunchDarkly: task(function*(currentUser) {
+    try {
+      const organizations = yield currentUser.get('organizations');
+      return this.get('launchDarkly').identify({
+        key: this.get('currentUser.userHash'),
```
djones: :+1:

app/routes/application.js
```diff
@@ -29,6 +32,32 @@ export default Route.extend(ApplicationRouteMixin, EnsureStatefulLogin, {
     return this._loadCurrentUser();
   },
 
+  _loadLaunchDarkly: task(function*(currentUser) {
+    try {
+      const organizations = yield currentUser.get('organizations');
```
maprules1000: 🎉 concurrency!

PR #651, opened 2018-07-26
Make sure admin mode is disabled for integration item test

maprules1000:
## Description
I noticed that when running the whole test suite in a browser that percy admin mode can cause the integration item test to fail, but when the test is run in isolation from the rest of the tests it passes. For some reason this is not encountered on ci.

In order to fix this weird state I'm explicitly clearing admin mode for this test, because it looks like some other test does not properly clear it.

## Reviewing this PR


cadeParade:
Thanks for doing this!

PR #660, opened 2018-08-02
Fix moment warnings

cadeParade:
## Description
As a followup to #645, this is a fix for the `moment` warnings that show up in the tests.

There were three root causes of these warnings:
- Bar graph on billing page was using `now` as an argument. This seemed to work when printing logs to the console, but it is an undocumented feature. I've changed it to use the `utc` helper which is documented on the `ember-moment` page, and that removes the warnings from the billing page.

- `build.baseBuild.createdAt`. Most test setups do not have a base build. When rendering certain components that look for the `baseBuild`, the test was trying to render a datetime that didn't exist. I've added a base build to each test case that needs it.

- `build.approvedAt`. Sometimes we want to render when the build is approved, but in the tests the `approvedAt` date was not set. Now it is.

Questions:
- Should we explicitly include `baseBuild` for builds that need it, or in the template check if it exists before displaying?
- Is there a way to get FactoryGuy to recognize a default property that is the same as itself? (ex: for baseBuild to be a default property, it would look something like: `baseBuild: FactoryGuy.belongsTo('build')` but this causes a stack overflow.


ballPointPenguin:
This all looks good to me. I'm not sure how to overcome the potential FactoryGuy recursion with build / baseBuild.

maprules1000:
Also, does it make sense to use the global `allowEmpty` property for the test environment? https://github.com/stefanpenner/ember-moment#global-allow-empty-dates

cadeParade:
Thanks both for your review. @maprules1000 setting `allowEmpty` to true in test environment is a great idea, I've done this.

PR #661, opened 2018-08-03
Clean up mirage config to check for correct user permissions

cadeParade:
## Description
While working towards publicly viewable projects (PVP), I needed to update quite a few of the mirage responses. 

This update:
- Centralizes the way we can produce errors for mirage responses (not related to PVP update, but nice)
- Checks if a user has access to an org (or if it has a public build) before returning an org.
- Checks if a user has access to a project (or if it public) before returning a project.

A side effect of updating these mirage responses to check for permissions is that a few other tests broke. 
In the case of the invite response, it started failing because the new `organizationUser` was not created by mirage, and therefore the user was not allowed to view the org before these updates.
In the case of some organization tests, there were two users set up with access to different orgs, but the tests expected one user to be able to access the other's org. This is now fixed also.

## Reviewing this PR
_**Is Covered by Automated Tests?: YES?**_

Mostly concerned with whether the permissions logic makes sense, which is contained in `_isOrgAllowed`, `_isProjectAllowed`, `_areAnyOrgProjectsPublic`, and `_isUserInOrg`.


mirage/config.js
```diff
@@ -27,18 +28,7 @@ export default function() {
...
-        {
-          errors: [
-            {
-              status: 'unauthorized',
-              detail: 'Authentication required.',
-            },
-          ],
-        },
-      );
+      return error401;
```
maprules1000: 🙌

mirage/config.js
```diff
@@ -313,4 +260,84 @@ export default function() {
...
+    const organizationUsers = schema.organizationUsers.where({userId: user.id});
+    const userOrgIds = organizationUsers.models.map(obj => obj.organizationId);
+    const allowedOrganizations = userOrgIds.map(id => schema.organizations.find(id));
+    const allowedOrganizationSlugs = allowedOrganizations.map(org => org.slug);
+
+    return allowedOrganizationSlugs.includes(orgSlug);
+  }
+}
+
+const error401 = new Mirage.Response(
```
maprules1000: can we move this variable definition to the top of the file with the others?

mirage/config.js
```diff
@@ -313,4 +260,84 @@ export default function() {
...
+    errors: [
+      {
+        status: 'unauthorized',
+        detail: 'Authentication required.',
+      },
+    ],
+  },
+);
+
+function error400({statusDetail = '', pointer = ' ', sourceDetail = ''} = {}) {
```
maprules1000: should this function name be `_error400`?

mirage/factories/subscription.js
```diff
@@ -21,4 +21,12 @@ export default Factory.extend({
       subscription.update({plan});
     }
   },
+
+  withTrialPlan: trait({
```
maprules1000: 👍

PR #665, opened 2018-08-09
Publicly viewable projects

cadeParade:
## Description
This PR does two main things
1. Standardize the handling of the cases where an org/project does not come back from the API regardless of authentication status.. Previously, the app would require a user to be authenticated when any org or project was being viewed. This has changed with some projects being publicly accessible. Now, if an org/project is returned from the API, it displays as normal (with one caviat -- see #2) and if the org/project doesn't come back, it prompts a login or shows an error, depending on the authentication status of the user. (https://github.com/percy/percy-web/pull/665/files#diff-b5806878a4e0dffbb46a22694fb6fe0e)

    To do this, the model for the routes (organization and project routes) are fetched in the `beforeModel` hooks. If the models come back, the loading of the routes continues. If they don't come back, they are handled by the file linked above.

2. Disables build/snapshot approve buttons in builds where the user is not authorized to update the build (when they are not logged in or are not a member of the organization). Testing for user membership in the org was actually a bit tricky, but the logic for how it happens is here: https://github.com/percy/percy-web/pull/665/files#diff-2820f5283474ede959fed1001a1fc938. `isUserMemberPromise` returns a promise which, after resolved, returns a boolean of whether the user is a member of the org. This property decides whether the builds/snapshots are approvable.

## Reviewing this PR

- [ ] Is it ok that it is asking for organization-users and failing? Should we stop the request altogether?
- [x] What is there is a public project with no builds? Presumably we should not show the getting started screen

## Manual testing setup: 
Make two user accounts (I will call them A and B)(I suggest one github and one username/email so it's easy to log into username/email in private browser)

For user A, make an organization.
Make the following projects for the organization:
- Public project with some builds. (X)
- Public project Y with no builds. (Y)
- Private project. (Z)

Ensure that all the above work while logged in as user A.
Visit project Y screen and make sure that the screen shows instructions for how to set up your first build.

Open a private window (or log out).
Visit url for all of the projects above.
Ensure that project X displays with builds, but approve buttons are disabled in build view and full-screen snapshot view
Ensure that project Y displays with the non-logged-in view of a project with no builds
Ensure that viewing project Z redirects you to login.

Visit organization URL. Make sure that only the public projects show up in the sidebar.

Log in as user B.
Visit url for all of the projects above.
Ensure that project X displays with builds, but approve buttons are disabled in build view and full-screen snapshot view
Ensure that project Y displays with the non-logged-in view of a project with no builds
Ensure that viewing project Z redirects you to error page.

Visit organization URL. Make sure that only the public projects show up in the sidebar.


app/lib/is-user-member-of-org.js
```diff
@@ -0,0 +1,18 @@
...
+  try {
+    return await org.get('_filteredOrganizationUsers');
+  } catch (e) {
+    return resolve(false);
+  }
+}
+
+export default function isUserMemberPromise(org) {
+  return DS.PromiseObject.create({
+    promise: _getOrganizationUsers(org).then(orgUsers => {
```
maprules1000: Since this is a little complex, and it's a helper fn that we could use elsewhere, could you add a comment explaining why it returns a `PromiseObject` and how it should be used?

cadeParade:
I agree that the error page could be nicer. I'm going to delegate that to a followup PR though, since that's been the state of the error page forever. I am thinking something like this: https://percy-io.slack.com/archives/C53NFE0HJ/p1535063223000100.

The public notice banner I think will get some more comments in beta, so I'm going to leave it for now. Do you mean the left alignment of the banner to the project title? It is technically exactly aligned. I've updated the color to be a bit lighter...

I'm going to merge for now to get this large chunk, but will definitely follow up addressing these concerns.

PR #669, opened 2018-08-14
Remove useless(?) approval-section component

cadeParade:
## Description
I cannot find any reason why there is a component wrapping the build approval button. Maybe there used to be a reason? But the conditional classes are no longer in our codebase, and the `approve` action is no longer used.


app/components/approval-section.js
```diff
@@ -1,20 +0,0 @@
...
-import Component from '@ember/component';
-
-export default Component.extend({
-  build: null,
-
-  approvedNow: false,
-  isApproved: alias('build.isApproved'),
-  classNameBindings: [
-    'classes',
-    'isApproved:ApprovalSection--approved',
```
maprules1000: huh. these css classes don't even exist any more

PR #671, opened 2018-08-17
[BUGFIX] Install GitHub app

maprules1000:
## Description
This fixes the styling of the installing loading state by adding back in some styles that were removed.

It also fixes a visual hang state where users might be prompted to link their github account before installation, despite already having linked that account.

### Before 

### After
screen shot 2018-08-21 at 4 24 51 pm


app/routes/organizations/organization/integrations/github.js
```diff
@@ -1,4 +1,13 @@
...
+import {inject as service} from '@ember/service';
 
-export default Route.extend(AuthenticatedRouteMixin, {});
+export default Route.extend(AuthenticatedRouteMixin, {
+  session: service(),
+
+  beforeModel() {
+    // If we don't force reload user on this page,
+    // we don't get the associated Identities
+    return this.get('session').forceReloadUser();
```
maprules1000: Fundamentally it looks like the api is not passing down the `identities` as part of the `user` request, and I'm not totally sure why. But triggering a reload of the `user` does get the identities.

maprules1000: This is the api line that is not working as expected: https://github.com/percy/percy-api/blob/master/app/controllers/api/v1/users_controller.rb#L11

maprules1000:
This is a style tweak and a quick technical fix for a slightly more complicated problem of making sure the api serializes `identities` with the `user`. Right now that code isn't working correctly, but it is slated to be fixed.

tests/acceptance/github-app-install-test.js
```diff
@@ -0,0 +1,87 @@
+import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
```
cadeParade: Minor:  prefer explicit import paths (as the line below is).
This one would be `percy-web/tests/helpers/setup-acceptance`.

tests/acceptance/github-app-install-test.js
```diff
@@ -0,0 +1,87 @@
...
+  setupAcceptance();
+
+  describe('without a linked github account', function() {
+    let organization;
+    setupSession(function(server) {
+      organization = server.create('organization', 'withNonGithubUser');
+    });
+
+    it('instructs a user to link a github account before installing the app', async function() {
+      await visit(`/organizations/${organization.slug}/integrations/github`);
```
cadeParade: This could be in the page object.

tests/acceptance/github-app-install-test.js
```diff
@@ -0,0 +1,87 @@
...
+
+    it('shows GitHub integration installation page', async function() {
+      await visit(`/organizations/${organization.slug}/integrations/github`);
+      expect(currentPath()).to.equal('organizations.organization.integrations.github');
+
+      await percySnapshot(this.test);
+    });
+
+    it('shows "Install Github" button when github is not installed', async function() {
+      await visit(`/organizations/${organization.slug}/integrations/github`);
```
cadeParade: Super minor, feel free to not act on this -- This line and the line below are repeated from the previous test. If you wanted to make it more DRY, you could put the visit in a `beforeEach`. And then testing that going to the same route gets to the same path is also duplicative. You can probably assume in the second test that going to the same route will have the same path.
Finally, having these two tests combined into one, with two Percy snapshots is reasonable -- it's not doing _that_ much, and acceptance tests are slow, so using one test to setup this whole scenario seems fine to me.
If you run the acceptance tests of the test suite, they each take about a half second to run on average, so the fewer acceptance tests we add at scale, the faster the tests will run.

tests/acceptance/github-app-install-test.js
cadeParade: These tests are great. Easy to read all the scenarios 👍

app/templates/components/organizations/github-integration.hbs
```diff
@@ -20,5 +20,5 @@
   </div>
   {{fa-icon "fa-spinner fa-pulse" classNames="OrganizationsGithubIntegration-leftAvatar"}}
   <header>Waiting for installation...</header>
-  Waiting for <strong>{{organization.githubIntegrationRequest.createdBy.login}}</strong> to install Percy in GitHub.
+  Waiting for <strong>{{organization.githubIntegrationRequest.createdBy.name}}</strong> to install Percy in GitHub.
```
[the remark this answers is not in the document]
maprules1000: It's been blank since that prop was removed from the user model I guess

tests/acceptance/github-app-install-test.js
maprules1000: Valid. I'll take a quick pass at a refactor

PR #679, opened 2018-08-22
[BUGFIX] Fix users data loading

cadeParade:
## Description
Visiting the organization users page had non-deterministic behavior where users would not always be present on the users page. 
This PR fetches the organization users on the users route and displays them in the template.

I am not sure why this route ever worked, I guess it was implicitly grabbing the users from the template, but this has stopped working recently (probably).
It is likely because of the recent serializer updates merged Friday (https://github.com/percy/percy-api/pull/579). It appears that `organization-users` relationship link sometimes doesn't come through even though it should, so the implicit `organization.get('organizationUsers')` stopped working. This solution gets the users explicitly in the model so the implicit relationship fetch does not need to happen any more.

While this fixes this particular problem, and the explicit model is a nice addition to the app, the disappearance of the relationship link needs further investigation.


app/routes/organizations/organization/users/index.js
```diff
@@ -0,0 +1,20 @@
...
+  model() {
+    const organization = this.modelFor('organizations.organization');
+    return hash({
+      organization,
+      users: this.store.query('organization-user', {organization}),
+    });
+  },
+
+  setupController(controller, resolvedModel) {
+    controller.setProperties({
```
maprules1000: 🙌

PR #681, opened 2018-08-24
Revert "Publicly viewable projects (#665)"

cadeParade:
This reverts commit 09b87d6b400ace7210bb1144ad83c26fda32f98a.
Reverted until authentication timeout issue has been resolved.

PR #682, opened 2018-08-25
Authenticate all routes

cadeParade:
Authenticate all routes until authentication bug is fixed.

Replaces https://github.com/percy/percy-web/pull/681


cadeParade:
🙏

PR #686, opened 2018-08-27
remove hard coded meta tag from index

cadeParade:


paytonealexis:
Awesome, thanks so much! 💃

PR #691, opened 2018-08-31
Implement Auth0 Silent Auth background token refreshing.

fotinakis:
## Description
Implement [Silent Auth](https://auth0.com/docs/api-auth/tutorials/silent-authentication) to fix session refresh issues, until it is supported in ember-simple-auth-auth0 ([issue](https://github.com/auth0-community/ember-simple-auth-auth0/issues/120)).

We perform Silent Auth once on first load, in the background, 10 seconds after the page has loaded, and only if no tab has performed silent auth in the last 30 minutes.

## Auth0 Background

- When you login you get an `id_token`. This is a JWT token with an expiration field on it.

- The `id_token` ALWAYS expires at the end of your configured JWT Expiration time. It is not automatically refreshed under any circumstances in a frontend SPA, "since for security reasons you cannot get a Refresh Token from the Implicit Grant (the OAuth flow typically used from Client-side Web Apps)." (https://auth0.com/docs/tokens/id-token#token-lifetime)
- You can get a new `id_token` anytime using auth0.js’s `auth0.checkSession` to fetch a new token in an iframe without users noticing — this is what “silent auth” does, which is currently unimplemented in ember-simple-auth-auth0.

Another interesting note about Simple Auth: it will never work in certain environments, such as when third-party cookies are disabled or with Safari ITP enabled. For this reason we do it in the background and inject into the session if it succeeds. Users with third-party cookies disabled or with Safari ITP enabled will always be logged out at the end of the JWT expiration time.

## Testing
**Is Covered by Automated Tests?: YES/NO** No.

Since we will be able to replace with this a native implementation at some point and this code is pretty isolated, opting for thorough manual testing and extensive inline comments and documentation. Many of the parameters and tuning was derived experimentally.

This has been thoroughly manually tested:
- With third-party cookies enabled
- With third-party cookies disabled (error is correctly handled)
- In an Auth0 environment where id_tokens would expire after 30 seconds. Can confirm that refreshing the page and waiting 10 seconds, a refresh request is made and the localstorage session gets updated with a new id_token and expiration.


app/authenticators/auth0-url-hash.js
```diff
@@ -1,14 +1,75 @@
...
+
+    // Auth0.checkSession injects an iframe and performs a redirects-only authentication flow,
+    // re-using the user's existing session to obtain a refreshed id_token.
+    auth0.checkSession(silentAuthOptions, (error, authResult) => {
+      if (error) {
+        Logger.error(error);
+        return;
+      }
+      authResult._lastSilentAuthRefresh = Date.now();
+      this.get('session').authenticate('authenticator:auth0-url-hash', authResult);
```
cadeParade: I suppose you did this on purpose, but I wrote a thing to ensure we get a `state` param from the API (https://github.com/percy/percy-web/blob/master/app/mixins/ensure-stateful-login.js#L9). I'm assuming this still works without a state param?

app/authenticators/auth0-url-hash.js
```diff
@@ -1,14 +1,75 @@
...
+
+      // Reset ember-simple-auth-auth0's session expiration timer.
+      // TODO: this is fragile, revisit when silent auth is natively implemented.
+      getOwner(this)
+        .lookup('route:application')
+        ._setupFutureEvents();
```
cadeParade: Where does this come from? What does it do?

XaserAcheron: Hi! Maintainer of `ember-simple-auth-auth0` here.

Short-ish version: there's a timer in the background that (currently) forces a log-out when the token expires, and the `_setupFutureEvents` function resets the expiration timer (see [here](https://github.com/auth0-community/ember-simple-auth-auth0/blob/develop/addon/mixins/application-route-mixin.js#L96) for the implementation). If this isn't called, the scheduled log-out will happen anyway when the original token expiration time is hit, which kinda defeats the purpose of the background refresh.

It's private API, but it won't change between now and when silent auth support is properly built in to `ember-simple-auth-auth0` (I'm working on it presently), so it should be safe.

fotinakis:
Closing this in preference for native support in https://github.com/percy/percy-web/pull/718

PR #706, opened 2018-09-06
Fix billing view bugs for custom plans.

fotinakis:


fotinakis:
Back to you @cadeParade with new tests.

PR #710, opened 2018-09-06
Add segment page view and event tracking

paulfri:
## Description
Adds analytics.js, pageview tracking, event tracking, user identification and grouping. Someone might have a better idea of what segmenting we want to do? Just copied the amplitude user properties for now.

Development only for now.

## Reviewing this PR
Click around, go to the [Segment event debugger](https://app.segment.com/percy/sources/percy_web_dev/debugger) to see the views/identifies/groups/events roll in.


app/router.js
```diff
@@ -30,6 +30,14 @@ const Router = EmberRouter.extend({
     }
     return true;
   }),
+
+  notifySegment: on('didTransition', () => {
```
fotinakis: Could rename `notifyGoogleAnalytics` above to `notifyAnalytics` and merge this into there so there's only 1 didTransition event listener.

app/services/analytics.js
```diff
@@ -52,6 +54,8 @@ export default Service.extend({
 
     this.organizationInstance.setUserId(null);
     this.organizationInstance.regenerateDeviceId();
+
+    this.segment.reset();
```
fotinakis: Where does `this.segment` come from? Is that supposed to be `window.analytics`?

app/services/analytics.js
```diff
@@ -98,5 +105,14 @@ export default Service.extend({
       );
       this.organizationInstance.logEvent(eventName, organizationEventProperties);
     }
+
+    if (organization) {
```
fotinakis: What do you think about syncing group information in the backend for all users?

paulfri: That seems okay? I'm not sure I understand the tradeoff, but it would definitely make these calls less noisy.

paulfri:
The Ember add-on looks pretty good and well-supported, but I'm not sure it makes sense to use while we still have the analytics service available and are still posting to GA/Amplitude manually. There's so little code here that I think we can get away with waiting to switch to the add-on until we reimplement those calls with Segment.

app/services/analytics.js
paulfri: yep, missed it

app/services/analytics.js
paulfri: So, at least for HubSpot, "Group calls only take effect when called via server-side libraries or mobile libraries, not via our client-side javascript library." So I think this is a no-op for now, but I'll leave it in until we figure that part out.

https://segment.com/docs/destinations/hubspot/

PR #714, opened 2018-09-10
Add margin top to demo form

Robdel12:
## Description

This adds some margin to the top of the form on the enterprise page so it's not right on top of the nav banner.

## Pictures
**Before**
image

**After**
image


cadeParade:
Hi @Robdel12, thanks for contributing! I'm going to pull this down and push it again to get the tests to run, but otherwise LGTM!

cadeParade:
Actually, only you can trigger a rebuild of this since it's on your fork. Could you try pushing again?

Robdel12:
Looks like that push didn't trigger the tests though. Does Buildkite do something similar to CircleCI where it doesn't allow builds from forks (for security reasons)?

cadeParade:
Thanks for being patient! We do have to trigger a test build manually, so I've finally done that and merged it. Thanks for the contribution!

PR #718, opened 2018-09-11
[WIP] Use Silent Auth support from ember-simple-auth-auth0.

fotinakis:
Blockers:
- [ ] If third-party cookies or Safari ITP are are disabled, normal auth is broken. Silent Auth needs to be optional.


config/environment.js
```diff
@@ -64,6 +64,25 @@ module.exports = function(environment) {
     'ember-simple-auth': {
       auth0: {
         logoutReturnToURL: '/',
+        silentAuth: {
+          // Automatically renew token every 30 minutes:
+          // renewSeconds: 1800,
+          renewSeconds: 10,
```
fotinakis: Must change this before merging.

PR #722, opened 2018-09-13
Remove `window.location.replace` make the back button work properly

Robdel12:
## Description

While clicking around the marketing site of percy I noticed the
redirect to the docs page uses `window.location.replace`, which
doesn't add to the browser history. So when I used the back button I
ended up two pages ago rather than the page I was actually on before.

I did a grep of the project to see where this was used and it doesn't
_seem_ like the replace is a specific design decision, but I could be
wrong! If so we can split this out into two different methods to
handle replacing and setting.

## Pictures

**Before**

**After**


Robdel12:
@cadeParade That makes total sense. After I cracked open the PR I also noticed there are some places where there's [downloads happening from a URL](https://github.com/percy/percy-web/blob/f1c3d5d832bf047b5dda374b02e91b3dc652d2f1/app/components/snapshot-viewer-header.js#L65-L77). Should those also not add to the history?

I totally can tackle this tonight / over the weekend 🎉

cadeParade:
Ah, yes, another good catch. Those should also not add to the history 👍 👍

Robdel12:
Sweet, I'll get it fixed up 💯

Robdel12:
Heh, I lied about doing it over the weekend. I decided to stay off a computer instead. I'm going to try and crush this sometime this week 💪🏼

cadeParade:
Looking good. If you check out the tests locally, especially [this test](https://github.com/percy/percy-web/blob/master/tests/unit/services/session-test.js#L103), you'll notice that it redirects to home. This is because the util function needs to be stubbed in the test. You can see an example of it [here](https://github.com/percy/percy-web/blob/master/tests/unit/services/session-test.js#L96). This is the reason we made the original command a util method, so we could prevent the tests from redirecting.

cadeParade:
Thanks so much for seeing this through!

cadeParade:
Also your commit messages are boss 👍

Robdel12:
@cadeParade HA thank you! I make sure I try to explain myself :p 

I appreciate the patience on this! 🎉

PR #728, opened 2018-09-18
Show a disabled badge when integration is broken

faun:
[Clubhouse Card](https://app.clubhouse.io/percy/story/2279/when-a-personal-access-token-becomes-invalid-alert-the-user-that-there-s-a-problem)
===

When an integration has a status of 'unauthorized', add a badge to
signify the integration is disabled and action is required.

---

See also: [Backend PR](https://github.com/percy/percy-api/pull/621)


tests/pages/components/integration-item.js
```diff
@@ -14,6 +16,8 @@ export const IntegrationItem = {
   contactButtonLink: attribute('href', SELECTORS.CONTACT_BUTTON),
 
   hasInstallButton: isPresent(SELECTORS.INSTALL_BUTTON),
+  hasUnauthorizedClass: isPresent(SELECTORS.UNAUTHORIZED_CLASS),
```
cadeParade: Would prefer the class you're checking for to be a string literal in `isPresent`. The `SELECTORS` hash is for selectors :D

tests/integration/components/organizations/integrations/integration-item-test.js
```diff
@@ -135,4 +135,22 @@ describe('Integration | Component | organizations/integrations/integration-item'
...
+      this.set('organization', organization);
+
+      this.render(hbs`{{organizations/integrations/integration-item
+        integrationName="gitlab"
+        integrationStatus="unauthorized"
+        organization=organization}}`);
+    });
+
+    it('has a badge to signify the integration is disabled', function() {
+      expect(IntegrationItem.hasUnauthorizedClass).to.equal(true);
```
cadeParade: Is this class doing anything? It seems like it is not. `status-unauthorized` does not have any styles associated with it, and also Percy test would catch it. Anyway, it seems like we can remove this class from the tests and the template itself with no ill effects?

PR #735, opened 2018-09-19
Update user menu to look nicer with new header.

cadeParade:
As the title says, this PR does two things: 
- Update user menu to look nicer with new header. (and also just rewrite it with way simpler styling)
- Updates the main menu dropdown and the user menu dropdown to have the click target cover the whole area of the `li`

Before: 

After:
fullscreen_9_18_18__6_02_pm


app/components/user-menu-light.js
```diff
@@ -0,0 +1,4 @@
+import Component from '@ember/component';
+import TargetApplicationActionsMixin from '../mixins/target-application-actions';
+
+export default Component.extend(TargetApplicationActionsMixin);
```
fotinakis: `TargetApplicationActionsMixin` is super old and very scary/brittle/fragile/terrible (i know because I wrote it).

Can we use the `route-action` helper all the time now and drop support for TargetApplicationActionsMixin?

tests/pages/components/user-menu-light.js
```diff
@@ -0,0 +1,10 @@
+import {create} from 'ember-cli-page-object';
+import ClickDropdownTrigger from 'percy-web/tests/pages/helpers/click-basic-dropdown-trigger';
```
fotinakis: This should be `clickDropdownTrigger` so we know it's a function and not `ClickDropdownTrigger` which looks like a class.

cadeParade:
@fotinakis Updated

PR #742, opened 2018-09-20
"Product" header menu opens on hover

cadeParade:
Taken fully from here: http://ember-basic-dropdown.com/docs/content-events

menu-open-on-hover


app/components/marketing/menu-bar.js
```diff
@@ -1,15 +1,38 @@
...
+    closeLater(dropdown) {
+      this.closeLater.perform(dropdown);
+    },
+
+    prevent() {
+      return false;
     },
   },
+
+  // Wait 3 seconds after mouse leaves dropdown for it to close.
```
fotinakis: It's waiting for 300 ms, not 3 s, right?

cadeParade: Um yep, that is what I meant

PR #743, opened 2018-09-20
Replace user avatar with "Go to App" button in marketing nav.

cadeParade:
Pass down `redirectToDefaultOrganization` to each instance of the marketing header. 
One alternate option that is much less code change is using 
`<a class="btn btn-primary ml-3" {{action (route-action "redirectToDefaultOrganization")}}>Go to App</a>`
But this is a bit magical, doesn't make it clear where it is coming from, and also would require extra setup in integration tests. I think the more verbose method as proposed in this PR is more clear and idiomatic. 

https://percy.io/percy/percy-web/builds/1024888/view/62450960/1280?browser=chrome&mode=head

### DIFF
--- a/app/templates/components/marketing/menu-bar.hbs
+++ b/app/templates/components/marketing/menu-bar.hbs
@@ -58,7 +58,7 @@
     {{#link-to "enterprise" class="ml-3 text-purple-600 hover:no-underline hover:text-purple-800"}}Enterprise{{/link-to}}
     {{!-- {{#link-to "customers" class="ml-3 text-purple-600 hover:no-underline hover:text-purple-800"}}Customers{{/link-to}} --}}
     {{#if session.isAuthenticated}}
-      {{user-nav user=session.currentUser class="ml-3"}}
+      <a class="btn btn-primary ml-3" {{action redirectToDefaultOrganization}}>Go to App</a>
     {{else}}
       <a class="ml-3 text-purple-600 hover:no-underline hover:text-purple-800 cursor-pointer" {{action "clickLogin"}}>Login</a>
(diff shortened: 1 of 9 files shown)


app/templates/components/marketing/menu-bar.hbs
```diff
@@ -58,7 +58,7 @@
     {{#link-to "enterprise" class="ml-3 text-purple-600 hover:no-underline hover:text-purple-800"}}Enterprise{{/link-to}}
     {{!-- {{#link-to "customers" class="ml-3 text-purple-600 hover:no-underline hover:text-purple-800"}}Customers{{/link-to}} --}}
     {{#if session.isAuthenticated}}
-      {{user-nav user=session.currentUser class="ml-3"}}
+      <a class="btn btn-primary ml-3" {{action redirectToDefaultOrganization}}>Go to App</a>
```
fotinakis: Does this need quotes? `"redirectToDefaultOrganization"`?

app/templates/features.hbs
```diff
@@ -1,3 +1,6 @@
 {{title "Features"}}
-{{marketing/marketing-page page=model class="marketing-features-page"}}
+{{marketing/marketing-page
+  page=model
+  redirectToDefaultOrganization=(route-action "redirectToDefaultOrganization")
```
fotinakis: Since this is always the same, we should push this down into the `menu-bar` component. That should be what `route-action` helps us do is not have to pass this in everywhere.

cadeParade:
The add-on authors explicitly say it is not meant for use in components
https://github.com/DockYard/ember-route-action-helper/blob/master/README.md#overriding-route-action-for-integration-tests

PR #744, opened 2018-09-20
Enable local feature flags.

fotinakis:
This sets `local: true` in dev & test, which allows you to run things like:

```
> ld.variation('new-pricing-plan', 'plan-a') // return the current value of the feature flag providing a default if it doesn't exist
 
> ld.setVariation('new-pricing-plan', 'plan-x') // set the variation value
 
> ld.enable('apply-discount') // helper to set the return value to `true`
> ld.disable('apply-discount') // helper to set the return value to `false`
 
> ld.allFlags() // return the current list of feature flags and their values
 
> ld.user() // return the user that the client has been initialized with
```

https://www.npmjs.com/package/ember-launch-darkly#local-feature-flags


cadeParade:
I don't really want to hard code feature flag names into the environment file. Can we just use `setVariation` in the console when you want to have the flag true?

fotinakis:
Per discussion, just going to enable dynamic local handling and set flags manually in dev.

PR #752, opened 2018-09-21
Marketing: fix CTA links. Fix hijacking relative links.

fotinakis:


app/components/marketing/marketing-page.js
```diff
@@ -2,16 +2,12 @@ import Component from '@ember/component';
 
 export default Component.extend({
   click(event) {
-    // Support third-party links that Ember doesn't know about. This emulates the behavior of
-    // link-to for smooth in-app transitions, but for links that come from Contenful and rendered
+    // Support third-party relative links that Ember doesn't know about. This emulates the behavior
```
fotinakis: Links are relative now in contentful, so adjusting this.

cadeParade: Thanks for updating the comment!

PR #755, opened 2018-09-21
hero adjustments

PixelJanitor:


cadeParade:
The enteprirse hero text is broken :( https://percy.io/percy/percy-web/builds/1027594/view/62660614/1280?browser=firefox&mode=diff

cadeParade:
But also yay for Percy!

PR #760, opened 2018-09-21
Switch layout columns based on number of texts

cadeParade:


app/components/marketing/content-block.js
```diff
@@ -0,0 +1,17 @@
+import Component from '@ember/component';
+import {readOnly} from '@ember/object/computed';
+import {computed} from '@ember/object';
+
+export default Component.extend({
+  supportingTexts: readOnly('block.supportingTextSections'),
```
fotinakis: Maybe supportingTexts --> supportingTextSections?

PR #761, opened 2018-09-24
Use percy-agent

djones:
This PR swaps `percy-web` to use a version of `ember-percy` which utilizes `percy-agent`. 

The following visual differences are expected: https://percy.io/percy/percy-web/builds/1028858

* The home page renders slightly differently.
* We now correctly serialize all input fields. So you see some input fields going from no input, to some input.


package.json
```diff
@@ -32,6 +32,7 @@
   "license": "MIT",
   "devDependencies": {
     "@ember/optional-features": "^0.6.1",
+    "@percy/agent": "0.0.41",
```
fotinakis: Seems like it should be a dependency of `ember-percy`, right?

djones: We've discussed this one before. Previously we landed on keeping it separate. I can't remember the reasons for that tbh.

I'm happy to move it. It's one less thing to do to upgrade. Any thoughts on why this would be a good or bad thing?

djones: One nice thing about `percy-agent` making a dependency on `ember-percy` is that it would allow us to guarantee compatibility with `percy-agent`. It would not be possible to get into a situation where say `ember-percy` was out of date and `percy-agent` was way ahead and had breaking changes.

PR #767, opened 2018-09-25
Update features styles for supporting text links

cadeParade:


fotinakis:
This ended up breaking the styles for the old vs. new bullets, so I just made all of them have a consistent bullet style. Also fixed styling of bullets on dark.

PR #777, opened 2018-09-25
Redirect to default org when user visits signup or login route when a…

cadeParade:
No longer rely on ember-simple-auth unauthenticatedRouteMixin to redirect if a user is authenticated when they visit `login` or `signup`. Instead intercept the request and redirect them to their default org.


app/routes/unauthenticated-base.js
```diff
@@ -0,0 +1,13 @@
+import Route from '@ember/routing/route';
+import {inject as service} from '@ember/service';
+
+export default Route.extend({
+  session: service(),
+  beforeModel(transition) {
+    this._super(...arguments);
+
+    if (this.get('session.isAuthenticated')) {
+      transition.send('redirectToDefaultOrganization');
```
fotinakis: I think it'd be more traditional to use `this.send` since you're on a route? I can't find any documentation of `transition.send`.  And might be good to return this if it's a promise?

app/routes/unauthenticated-base.js
```diff
@@ -0,0 +1,13 @@
+import Route from '@ember/routing/route';
+import {inject as service} from '@ember/service';
+
+export default Route.extend({
+  session: service(),
+  beforeModel(transition) {
+    this._super(...arguments);
```
fotinakis: Dug around a bit in the ember-simple-auth mixin that this replaces and noticed that you might want to `return` this in an else block below, otherwise it might not do anything to just call super in this hook.

See as reference:
https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/unauthenticated-route-mixin.js#L89

app/routes/unauthenticated-base.js
cadeParade: Using `this.send` throws an error when a user loads `signup` or `login` route directly. `“Can’t trigger action 'redirectToDefaultTransition ’ because your app hasn’t finished transitioning into its first route, you can call .send() on the Transition object passed to the model/beforeModel/afterModel hooks”`
The solution to this is to use `transition.send`.

cadeParade:
Back to you @fotinakis

PR #783, opened 2018-09-26
Allow cmd/ctrl click to open links in new tab

cadeParade:


app/components/marketing/marketing-page.js
```diff
@@ -9,14 +11,14 @@ export default Component.extend({
       // The target link is part of this app. Hijack the transition and stop bubbling.
       let path = event.target.href.split(window.location.origin)[1].split('#')[0];
 
-      this.send('navigate', path);
+      // If ctrl or cmd are pressed, open a new tab.
+      if (event.ctrlKey || event.metaKey) {
+        window.open(path);
```
fotinakis: Alternatively I think you can just `return true;` work here, which might be more cross-device compatible and accessible because it'll just do the browser default.

PR #786, opened 2018-09-26
Remove updated-marketing-site flag

cadeParade:
Removes all the places where the marketing page feature flag is checked.


app/routes/index.js
```diff
@@ -1,17 +1,8 @@
 import Route from '@ember/routing/route';
-import ResetScrollMixin from '../mixins/reset-scroll';
 import metaTagLookup from 'percy-web/lib/meta-tags';
-import {inject as service} from '@ember/service';
 
-export default Route.extend(ResetScrollMixin, {
```
maprules1000: Why remove the reset scroll mixin?

cadeParade: I've tested all these routes and they no longer need it.
I think this is needed if you're switching between child routes, like between `route/new` and `route/edit` where there is a parent template between `new` and `edit`. In that case I believe the scroll wouldn't reset. These are all top level routes so it always resets.

PR #787, opened 2018-09-27
Remove unnecessary component wrappers

PixelJanitor:
- Removes application template wrapper div
- Removes template only wrapper divs

https://github.com/emberjs/ember-optional-features

### DIFF
--- a/app/components/browser-switcher-badge.js
+++ b/app/components/browser-switcher-badge.js
@@ -1,0 +1,3 @@
+import Component from '@ember/component';
+
+export default Component.extend({});
--- a/app/components/changelog-post-list.js
+++ b/app/components/changelog-post-list.js
@@ -1,0 +1,3 @@
+import Component from '@ember/component';
+
+export default Component.extend({});
--- a/app/components/changelog-post.js
+++ b/app/components/changelog-post.js
@@ -1,0 +1,3 @@
+import Component from '@ember/component';
+
+export default Component.extend({});
--- a/app/components/commit-table.js
+++ b/app/components/commit-table.js
@@ -1,0 +1,3 @@
+import Component from '@ember/component';
+
+export default Component.extend({});
--- a/app/components/marketing/customer-quote.js
+++ b/app/components/marketing/customer-quote.js
@@ -1,0 +1,3 @@
+import Component from '@ember/component';
+
+export default Component.extend({});
--- a/app/components/marketing/faq-block.js
+++ b/app/components/marketing/faq-block.js
@@ -1,0 +1,3 @@
+import Component from '@ember/component';
+
+export default Component.extend({});
--- a/app/components/marketing/footer-cta.js
+++ b/app/components/marketing/footer-cta.js
@@ -1,0 +1,3 @@
+import Component from '@ember/component';
+
+export default Component.extend({});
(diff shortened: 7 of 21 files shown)


cadeParade:
Why do we need all these extra js files?

PixelJanitor:
Not sure - Michael and I are discussing this as we speak. The ember optional features did that when it ran itself. I just deleted one of them and everything still worked.

maprules1000:
Ok I have a good idea about what's going on. First, the `"template-only-glimmer-components": true` flag treats `hbs` files the do not have corresponding `js` files like glimmer components. This means 3 things:

1. It will not have a wrapping ember-view div
2. The template will not be able to use {{this}} as it will attempts to reference a js file that does not exist. (it looks like ember has been instantiating an empty component in the background, but that does not happen when it's treated like a glimmer component)
3. Any `class` attributes that were previously passed into the components will not be applied since there is no longer a wrapping div for those classes to be applied to.

So it looks like we might need some of those empty component files.

maprules1000:
@PixelJanitor I'm testing this out locally by deleting the added `js` files one by one and seeing the resulting changes and it looks like it's going to be a bit of grunt work to get everything up and running.

Every template file without a `js` file has to access variables with `@`. So if you want to remove, say, `changelog-post.js` then you have to go through `changelog-post.hbs` and add `@` in front of every variable that gets passed into the component.

Without the `js` files you have to reference the variable in a more direct way.

app/templates/components/marketing/footer-cta.hbs
```diff
@@ -2,27 +2,27 @@
...
-            {{content.smallerText}}
+            {{@content.smallerText}}
           </p>
         {{/if}}
       </div>
     </div>
     <div class="btn-toolbar flex flex-1 lg:justify-end">
-      {{#if (and content.secondaryButtonText content.secondaryButtonLink)}}
+      {{#if (and @content.secondaryButtonText @content.secondaryButtonLink)}}
         <a href="{{content.secondaryButtonLink}}" class="btn-md flex items-center justify-center border border-purple-500 rounded link-white">
```
maprules1000: Look like you missed this one

app/templates/components/marketing/footer-cta.hbs
```diff
@@ -2,27 +2,27 @@
...
-          {{content.secondaryButtonText}}
+          {{@content.secondaryButtonText}}
         </a>
       {{/if}}
       <a href="{{content.primaryButtonLink}}" class="btn btn-white btn-md flex items-center text-center border border-purple-100 px-2 py-1">
```
maprules1000: this one too

app/templates/components/marketing/menu-bar.hbs
```diff
@@ -8,31 +8,27 @@
...
               linkRoute="how-it-works"
               icon="marketing/green-tag-page"
               linkColor="text-green-600"
             }}
             {{!-- {{marketing/menu-bar-dropdown-item
-              tagName=""
               text="Integrations"
               linkRoute="integrations"
               icon="marketing/git"
             }} --}}
```
maprules1000: can this whole commented out block be removed?

app/templates/components/projects/quickstart-code.hbs
```diff
@@ -20,7 +20,7 @@ bundle install
 {{#if project.tokens.isPending}}
```
maprules1000: Missed this too

PR #788, opened 2018-09-27
Update meta tags for new routes

paytonealexis:


app/lib/meta-tags.js
```diff
@@ -37,6 +55,13 @@ const headTags = {
     ),
   ],
 };
```
PixelJanitor: This line needs to be removed for the proper nesting I believe

PR #790, opened 2018-09-28
New default org routes

cadeParade:
## Description
This PR solves a problem that will come up in the new banner for the PVP project. In it, we will have a button that says "Jump to your projects". This could link to the current `redirectToCurrentOrganization` method, but this method uses localStorage preferentially which has **the most recently visited org** stored in it. This means that whenever a user clicked that button, they would be taken to the same org they were on, which would not be their org.
So we need two options for redirecting to "default" org -- 1) Redirect to most recently viewed org and 2) Redirect to default org _that belongs to the viewer_.

To this end, this PR introduces two new things: 
1) A service that controls which org should be the default org a user should be redirected to. This method has been a pain for a while, but is exacerbated by the PVP project. The service has a method of the same name, `redirectToDefaultOrganzation` but now has an optional parameter `useMostRecentOrg`. When this is false, it skips the localStorage lookup.
2) Two new routes, one that redirects to default org and one that redirects to most recent org. With these routes we can reference them in `link-to`s in the template and not worry about passing any actions down or importing the service.

## Reviewing this PR
Whatchu think of these patterns?

_**Is Covered by Automated Tests?: Yesish -- the service functionality is tested through the new route acceptance tests.**_


app/components/goto-app.js
```diff
@@ -13,7 +14,7 @@ export default Component.extend({
   classNameBindings: ['classes'],
 
   click() {
-    this.sendAction('redirectToDefaultOrganization');
```
maprules1000: YES. Goodbye sendAction! You will not be missed!

app/templates/features.hbs
```diff
@@ -2,6 +2,5 @@
 {{marketing/marketing-page
   page=model
   transitionTo=(route-action "transitionTo")
-  redirectToDefaultOrganization=(route-action "redirectToDefaultOrganization")
```
maprules1000: 🙌

app/routes/application.js
```diff
@@ -149,24 +150,7 @@ export default Route.extend(ApplicationRouteMixin, EnsureStatefulLogin, {
   },
 
   _redirectToDefaultOrganization() {
```
maprules1000: If this lives in a service, do we still need this action in this route?

maprules1000: There's an associated `redirectToDefaultOrganization` in the action block that consumes this. Can both be removed, or is there a good reason to keep them around?

maprules1000:
It looks like there's a missing service injection in the tests when I run it all in a browser. 

The `application-test.js` unit test needs `service:redirects` added to the needs array.

cadeParade:
We resolved @maprules1000 comments in a slack conversation, the summary was:
1) "It looks like there's a missing service injection in the tests when I run it all in a browser. The application-test.js unit test needs service:redirects added to the needs array."
This is correct but it's pretty weird that the tests pass remotely without this?

2)"Also, it looks like the new header manages to pop out of the testing window and stick to the top of the browser window? I don't think it's a blocker, but it's a change."
This is not new with this PR, the build page header is styled to stick to the top of the screen, but percy snapshots still are ok.

app/routes/application.js
cadeParade: Yep, good call, removed all remaining references to this action and the action itself 👍

PR #798, opened 2018-10-01
Upgrade husky and ember-cli-dependencies-workflow

cadeParade:
Replaces https://github.com/percy/percy-web/pull/795 and https://github.com/percy/percy-web/pull/797 while resolving conflicts.


maprules1000:
This looks good, but I think that some husky configuration needs to be moved around. I'll investigate and add to this branch.

PR #803, opened 2018-10-02
Fix tests

maprules1000:
`percy-agent` currently swallows buildkite test failures 😬 so these errors snuck through.


app/templates/components/snapshot-viewer.hbs
```diff
@@ -1,45 +1,53 @@
...
-    activeBrowser=activeBrowser
-    updateActiveBrowser=updateActiveBrowser
-    isBuildApprovable=isBuildApprovable
-  }}
+<div
+  class="SnapshotViewer mb-2
+    {{if isFocus 'SnapshotViewer--focus'}}
+    {{if (not isExpanded) 'SnapshotViewer--collapsed'}}
+    {{if isActionable 'SnapshotViewer--actionable'}}"
+  data-test-snapshot-viewer
```
maprules1000: I am not ok with this, but since the `attributeBinding` pattern isn't working any more, and these class names need to be directly referenced through the page object test handler, this is the best solution I could come up with.

maprules1000:
Overall I am not happy that the attributeBindings pattern that we have been using no longer works. I am not sure why, but I have fixed these tests so that they will no longer be broken on master.

I'm going to dig only a little more to see if I can find out why attributeBindings is no longer working as expected. If I can't come up with a good reason, and if we agree that this is not the best path forward, then we can roll back #787 since that is the source of these issues.

app/templates/components/snapshot-viewer.hbs
maprules1000: I think it feels not ok because it's a change from the successful way things worked before, which is frustrating IMO, but may be good in the long run.

The page objects aren't really the problem. They are a bit of a complication, but not the source of the issue. I thought the `attributeBinding` was also the issue, but it looks like `classNames` is also not properly being picked up at the time `ember-cli-page-objects` is looking for the info.

Basically, with glimmer template syntax enabled `ember-cli-page-objects` does not like for any selectors to be declared in js. It looks like it's a timing issue more than anything, but I can't find any bugs about it, and I'm not really inclined to dig into it any further than I have.

Bottom line, if we choose this I'm 💯 ok with that, but I would also want to go through the whole code base and change out the remaining instances where classes, tags, and attributes are declared in JS. Now, that sounds like a ton of work, but I honestly think I can get it done in under 2 hours, as long as there are no major complications.

maprules1000:
Might have figured out the source of the problem and a fix. The PR should be much simpler.

maprules1000:
Ok I know more about this issue now. It boils down to a scope problem. I couldn't ID the change that made it a problem, but `ember-cli-page-objects` prepends the top level object scope to all child element selectors which results in being unable to find the objects. It looks like it comes down to these steps deep in the `ember-cli-page-objects` code: 

https://github.com/san650/ember-cli-page-object/blob/master/addon-test-support/-private/execution_context/integration.js#L120
https://github.com/san650/ember-cli-page-object/blob/master/addon-test-support/-private/helpers.js#L144

It looks like something funky happens when we specify a page object's scope. The selector should not have the parent scope selector in it since the `testContext` is already within the parent element. Thus the scope selector won't work because the context is already established.

I poked at several possible fixes and if it fixes the integration test then it breaks the acceptance test. I'm not going to spend any more time on it. I thought there might be a simple solution but I haven't been able to discover it yet.

cadeParade:
Thanks for digging into this! 
Maybe I am missing something, but it seems like all this is to be expected...the `template-only-glimmer-components` flag means that a component no longer has a DOM element. Therefore `classNames` and `attributeBindings` cannot work as they were written before in a js file, because there is no DOM element to bind to. The `classNames` and `attributeBindings` assignments in a component js file just silently do nothing. Putting them in the template as is done in this PR is afaict the _only_ way to have the classes and attributes be assigned to DOM elements.

maprules1000:
Closed in favor of #806

maprules1000:
@cadeParade the problem with this change is that some components declare DOM events in the js file. So if there's no longer a wrapping element for those events to attach to then we need to rethink how all of those events work, and some of them are not trivial. 

ex: https://github.com/percy/percy-web/blob/master/app/components/snapshot-viewer-full.js#L61
and other bits.

On top of that, really, I would like to change the pattern across the whole app at once, and that would add extra time to the effort, but I think it is necessary so we don't leave the codebase in a weird in-progress state where we have both old and new patterns.

Wrapping div removal will be native to ember eventually, so it's fine if we don't have it right now.

PR #804, opened 2018-10-03
Add tests and component for commit table

faun:
## Description
The commit table source code links are not working correctly. They should link to the repo branch & commit SHA, but they do not. This PR fixes the broken functionality and adds a component JS file and page objects, as well as an integration test.

## Reviewing this PR

Ideally, there should be no change in functionality here, aside from restoring the functionality before https://github.com/percy/percy-web/pull/787. If you see anything that's different, let me know.


app/components/commit-table.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default Component.extend({
+  build: null,
+  commit: null,
+  isRepoLinked: alias('build.isRepoLinked'),
+  commitUrl: computed('build.repo', 'commit.sha', function() {
+    let repo = this.get('build.repo');
+    let commit = this.get('commit');
+    if (repo && commit) {
+      return `${repo.htmlUrl}/commit/${commit.sha}`;
```
cadeParade: If we're trying to make this generic for different SCMs, we should make sure `commit` and `tree` (on line 20) are also generic, since I believe you said those particular strings were not applicable to bit bucket.

faun: Fixed in 789f1f2

PR #808, opened 2018-10-04
Fix console warnings for style bindings

cadeParade:
In the marketing website push, we got a little sloppy with style attribute bindings. This makes them "safe" so the console and tests are not filled with warnings anymore.


maprules1000:
(I tested it locally too)

PR #809, opened 2018-10-04
Update example instructions so all builds aren't auto approved

maprules1000:
The branch name is `local/test` on the example-quickstart repo. I also considered `local/quickstart` and just `local`.

If the branch name is changed I'll have to also change it on the example-quickstart repo


app/templates/components/projects/quickstart-code.hbs
```diff
@@ -2,7 +2,7 @@
   Follow these steps to get the demo project running. First clone the Git repository:
 </p>
 
-{{#code-block lang="nohighlight"}}git clone https://github.com/percy/example-quickstart.git
+{{#code-block lang="nohighlight"}}git clone -b local/test https://github.com/percy/example-quickstart.git
```
faun: I made `local/test` a protected branch on the example-quickstart repository.

faun:
On second thought, it's probably better to checkout master locally and then create a new branch as a separate command. This is more in alignment with the standard git flow that most people are familiar with.

app/templates/components/projects/quickstart-code.hbs
maprules1000: Thanks!

PR #822, opened 2018-10-09
[Ember upgrade - Part 1] Updates to data loading for Ember upgrade

cadeParade:
## Description
In preparation for the Ember upgrade, [a few changes to the API serializers](https://github.com/percy/percy-api/pull/650) must be made. These changes, in turn, require a few updates to how percy-web is handling its data.

There were two main things that needed to be changed. 
1) In the case of the billing route, this query needed to be told explicitly to query the server each time it is run. That's because with the ember-data updates, the app would default to finding the organization, which is already loaded, and not run the query with the usage stats included. Adding `reload: true` forces the query to happen whenever the billing model hook is run.

2) The rest of the changes are a little more subtle. A problem manifested that sometimes a snapshot was returned from a query _without_ including the comparisons belonging to it. This is a problem because in the snapshot model, comparisons are not async, so it always expects a snapshot's comparisons to be present.
It turns out that snapshots from another build are sometimes loaded as `includes` -- a comparison includes head and base snapshots which are sent in the `includes` array of a response. Having these extra snapshots present did not previously bother the app, presumably because the snapshot linkage data was not present and therefore it was just a floating orphaned entity, but with the [`include_data:true` on the comparisons line of snapshot serializer](https://github.com/percy/percy-api/pull/650/files#diff-80ec3bbd447be1ce420be0b08eabcd18R12) combined with the ember-data upgrade, it breaks. 

So, the rest of the changes are to address this problem. 
- `hasMany('snapshots')` -- Moving away from this method in general as it is unreliable. Instead find all instances in the store and filter them.
- `allSnapshotsAreLoaded` -- this check no longer worked because it counted how many snapshots of the build were in the store vs how many the build says it has. Now that there are snapshots that might have been loaded originating from other builds, it is broken. 
Removing the logic surrounding this check does mean that occasionally a user will reload all unchanged snapshots (for instance if they load the unchanged snapshots, navigate to another build, then load the same build again). I imagine this doesn't happen so often as to matter much to users, but I'm open to hear otherwise and find another way of caching. 
- add `.filterBy('isChanged')` -- The original filter method no longer worked because it fetched all snapshots for a build in the store -- including ones loaded as head/base snapshots from another build. It now filters snapshots by their build _and_ whether they have diffs.
- `isUnchangedSnapshotsVisible`  -- This property was not resetting on route change so the state of the snapshot list was not resetting between builds. 

## Reviewing this PR

_**Is Covered by Automated Tests?: YES?**_ The tests still pass...

[Part 1] - This PR
[Part 2](https://github.com/percy/percy-api/pull/650)
[Part 3](https://github.com/percy/percy-web/pull/826)


app/components/build-container.js
```diff
@@ -70,31 +71,17 @@ export default Component.extend(PollingMixin, {
...
-      yield this.get('snapshotQuery').getUnchangedSnapshots(this.get('build'));
-      loadedSnapshots = this._getLoadedSnapshots();
-    } else {
-      // If the cached snapshots are used, the loading spinner does not display because it
-      // completes this operation in the same run loop. So pause for half a second to
-      // give the user some indication we have processed their click.
-      yield timeout(500);
-    }
+    yield this.get('snapshotQuery').getUnchangedSnapshots(this.get('build'));
+    loadedSnapshots = this._getLoadedSnapshots();
```
maprules1000: I like that this is much simpler now

app/components/build-container.js
```diff
@@ -70,31 +71,17 @@ export default Component.extend(PollingMixin, {
...
   _getLoadedSnapshots() {
     // Get snapshots without making new request
-    return (
-      this.get('build')
-        .hasMany('snapshots')
-        .value() || []
-    );
+    return this.get('store')
+      .peekAll('snapshot')
+      .filterBy('build.id', this.get('build.id'));
```
maprules1000: 🎉

PR #828, opened 2018-10-11
Infinity Loading for Project Page

PixelJanitor:
## Description
This PR adds pagination to the builds page in the form of infinite scrolling using the `ember-infinity` addon. There is also a placeholder build item that shows up while fetching more builds from the server and disappears when there are no more builds to be fetched.

## Pictures
Build loading state
screen shot 2018-10-17 at 3 19 17 pm

## Reviewing this PR
I've added an acceptance test to explicitly test the infinite scrolling functionality.

Note: This will be merged after #838


tests/acceptance/project-test.js
```diff
@@ -265,6 +267,26 @@ describe('Acceptance: Project', function() {
       expect(currentPath()).to.equal('organization.project.index');
     });
 
+    it('hides the loader when there are less than 50 builds', async function() {
```
maprules1000: Unfortunately there's no reliable way to wait for all the builds to load on the page to test that the infinity loader component does its job.

However, we can test if it properly shows on the page based on the number of builds on the server. That means data loading is working properly and the `afterInfinityModel` hook is functioning as intended.

cadeParade: You can set a delay for for requests if you want: http://www.ember-cli-mirage.com/docs/v0.1.x/server-configuration/#timing 
It is also possible to make a pending promise, check something, and then allow it to fulfill. I haven't thought about this scenario enough to know if that would work but it is _possible_.
But also this seems fine.

cadeParade: ex: https://github.com/percy/percy-web/blob/master/tests/integration/components/snapshot-approval-button-test.js#L76

tests/acceptance/project-test.js
```diff
@@ -265,6 +266,26 @@ describe('Acceptance: Project', function() {
...
+      await ProjectPage.visitProject(urlParams);
+
+      // expect infinity loader to be present but hidden
```
cadeParade: I think `isPresent` vs `isVisible` would give you the granularity you're looking for.

app/lib/paginated-ember-infinity-model.js
```diff
@@ -0,0 +1,20 @@
+import InfinityModel from 'ember-infinity/lib/infinity-model';
+import {get, set} from '@ember/object';
+
+const ExtendedInfinityModel = InfinityModel.extend({
+  buildParams() {
+    let params = this._super(...arguments);
+    params['page[cursor]'] = get(this, '_cursor'); // where `this` is the infinityModel instance
```
cadeParade: fance

app/lib/paginated-ember-infinity-model.js
```diff
@@ -0,0 +1,20 @@
...
+    return params;
+  },
+  afterInfinityModel(builds, params) {
+    // assumes the list is ordered
+    // if the server response is less than the request assume we've loaded all the data
+    const canLoadMore = builds.get('length') >= params.get('perPage');
```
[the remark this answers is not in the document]
cadeParade: Maybe put some of this comment in the code for future reference.

tests/acceptance/project-test.js
maprules1000: Well the issue is that once the loader enters the viewport it makes a call to the server to load more builds, and `ember-infinity` handles that call all on it's own. It's possible to check mirage and make sure that call was made, but it's not really possible to check that all of those things rendered to the page. 

But really if anything rendered to the page then the server call is working just fine.

PR #838, opened 2018-10-16
Refactor project status icon

maprules1000:
## Description
When installing `ember-infinity` to enable infinite scrolling on the project builds page, I ran into a [backtrack re-rendering error](https://github.com/emberjs/ember.js/issues/13948) where the project status icons data would change state while the page was rendering. Ember does not like this and it had been deprecated since 1.13, and removed in 2.10. 

I think the loading of the infinity model triggered this change, but also the build was being loaded through a query call in computed property on a model. So this had to be fixed.

## Reviewing this PR

I know it's not ideal to load data in a component, but since project status comes from a build we have to sometimes. The code is written to prefer locally loaded data and only run the query if the local data is not available.


app/components/project-list-item.js
```diff
@@ -0,0 +1,36 @@
...
+  serverRecentBuild: computed.readOnly('_getMostRecentBuild.lastSuccessful.value.firstObject'),
+
+  localRecentBuild: computed('project.builds', function() {
+    const allBuilds = this.get('store').peekAll('build');
+
+    const projectBuilds = allBuilds.filter(build => {
+      return build.get('project.id') === this.get('project.id');
+    });
+
+    const sortedBuilds = projectBuilds.sort((a, b) => {
```
cadeParade: Could probably use sortBy
```suggestion
    const sortedBuilds = projectBuilds.sortBy('id');
```

cadeParade: Can the suggestion thing do multiple lines? Dunno

app/components/project-list-item.js
```diff
@@ -0,0 +1,36 @@
+import Component from '@ember/component';
+import {computed} from '@ember/object';
+import {task} from 'ember-concurrency';
+import {inject as service} from '@ember/service';
+
+export default Component.extend({
+  store: service(),
+
+  recentBuild: computed.or('serverRecentBuild', 'localRecentBuild'),
```
cadeParade: Would this not favor the first argument (aka the one that goes to the server) over the second? I think so based on this: https://www.emberjs.com/api/ember/3.4/functions/@ember%2Fobject%2Fcomputed/or
"The computed property will return the first truthy value" -- so it would have to check `serverRecentBuild` every time?

app/components/project-list-item.js
```diff
@@ -0,0 +1,36 @@
...
+  serverRecentBuild: computed.readOnly('_getMostRecentBuild.lastSuccessful.value.firstObject'),
```
cadeParade: ~Presumably this works, but how does `_getMostRecentBuild` get called if there is no `perform`?~
This is very clever

app/models/project.js
```diff
@@ -27,11 +27,6 @@ export default DS.Model.extend({
 
   projectBrowserTargets: DS.hasMany('projectBrowserTargets', {async: false}),
 
-  _lastBuild: computed('organization', 'slug', 'builds', function() {
```
cadeParade: 🎉

app/templates/components/project-page.hbs
```diff
@@ -2,12 +2,10 @@
...
       {{#each sortedProjects as |project|}}
-        <li class="flex items-center mb-2 {{if project.isDisabled 'opacity-25'}}" {{action 'toggleSidebar'}}>
-          <div class="status-indicator flex-no-shrink mr-2 is-{{dasherize project.lastBuild.state}} is-{{if project.lastBuild.isFinished (dasherize project.lastBuild.reviewState)}}"></div>
-          {{#link-to "organization.project.index" project.organization.slug project.slug class="font-semibold link-gray truncate"}}
-            {{project.name}}
-          {{/link-to}}
-        </li>
+        {{project-list-item
+          project=project
+          toggleSidebar=(action 'toggleSidebar')
```
cadeParade: P4 Double quotes

app/components/project-list-item.js
maprules1000: so this won't trigger 'serverRecentBuild' to run, but if the state of `serverRecentBuild` changes the template will re-render the `recentBuild` but it won't trigger a re-render if `localRecentBuild` is first. 

I think that's what is up but I'm going to double check.

app/components/project-list-item.js
maprules1000: I tried using `sortBy` but it wasn't working for some reason. I'll take another whack at it.

app/templates/components/project-page.hbs
maprules1000: ?

app/components/project-list-item.js
maprules1000: Oh, `sortBy` didn't work because this is an array of classes, not simple objects. So I need to `get` the ids to sort on.

app/components/project-list-item.js
maprules1000: Actually, it works the same either way. I'll switch it up since, logically, it should prefer local over server.

PR #846, opened 2018-10-18
Refactor showSupport 

cadeParade:
There were a bunch of `showSupport` actions being passed around, unclear where they were from or where they were going. Many were not being used. Also, several of them were broken.

This PR makes one method for showing intercom and imports it directly where it is needed. 

The one exception is that the error template needs it and there's no route or js file for it so it still needs the application route action.


app/components/organizations/integrations/gitlab-settings.js
```diff
@@ -19,12 +20,12 @@ export default Component.extend({
...
 
   actions: {
-    openIntercom() {
+    showSupport() {
       const messageText =
         "Hi! I'd like to run GitLab Self-Managed behind a firewall." +
         ' Do you have information for me on how to get started?';
 
-      window.Intercom('showNewMessage', messageText);
+      this.get('intercom').showIntercom('showNewMessage', messageText);
```
maprules1000: 🙌

app/services/intercom.js
```diff
@@ -18,4 +18,14 @@ export default Service.extend({
       });
     }
   },
+
+  showIntercom(...args) {
```
maprules1000: This is great!

PR #847, opened 2018-10-19
Fix organization switcher

cadeParade:
[This ancient ticket I made](https://app.clubhouse.io/percy/story/1716/test-for-org-switcher-nav-that-data-loads)
[This more recent ticket I made](https://app.clubhouse.io/percy/story/2452/organization-switcher-component-does-not-update-when-a-user-creates-a-new-org-without-a-hard-refresh)

## Description
Somewhere along the way this query got broken again. But in a slightly subtler way. It would not show an organization that was recently created without a page refresh. This is a bug that is present in production right now (even before the most recent upgrade) for an unknown amount of time. So this fixes it, plus adds a test so hopefully this regression does not happen again.

## Reviewing this PR


app/adapters/application.js
```diff
@@ -91,6 +91,11 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
       return utils.buildApiUrl('buildSnapshots', buildId);
     }
 
+    if (requestType === 'query' && modelName === 'organization' && query.user) {
+      delete query.user;
+      return utils.buildApiUrl('userOrganizations');
```
maprules1000: Everything else makes sense, but what's this all about?

cadeParade: We discussed this briefly in slack but, since asking for `user.get('organizations')` didn't trigger a re-fetch of the data, I had to explicitly call out to user organizations endpoint. That is a special endpoint (`api/v1/user/organizations` -- note there is no user id in the url), so to use the `store.query` method, a custom handler had to be made.

cadeParade: Here's the slack link if you're interested. https://percy-io.slack.com/archives/DAH3H44RM/p1539714077000100

maprules1000: Ah yes, I remember now 👍

PR #852, opened 2018-10-19
Add webhook-event model

paulfri:
Just some boilerplate for an upcoming PR. Opening separately to cut down on the PR size.


app/models/project.js
```diff
@@ -24,7 +24,7 @@ export default DS.Model.extend({
 
   builds: DS.hasMany('build', {async: true}),
   tokens: DS.hasMany('token', {async: true}),
-  webhookConfigs: DS.hasMany('webhookConfig', {async: false}),
+  webhookConfigs: DS.hasMany('webhookConfig', {async: true}),
```
paulfri: The endpoint necessary for this was added in https://github.com/percy/percy-api/pull/674

app/models/webhook-config.js
```diff
@@ -3,6 +3,7 @@ import utils from 'percy-web/lib/utils';
 
 export default DS.Model.extend({
   project: DS.belongsTo('project'),
+  webhookEvents: DS.hasMany('webhook-event'),
```
cadeParade: It looks like in the API you're always including `webhook-events` that belong to a `webhook-config`. So this should be `async: false`. If `async: true`, ember-data will sometimes try to ask for webhook-events via the webhook-event link ([which doesn't exist](https://buildkite.com/percy/percy-web/builds/3128#2d8419ed-f358-4fb8-8e36-adf2f3840239))  and that will make ember-data sad.

PR #857, opened 2018-10-22
Clear admin mode on logout

cadeParade:
Since
- admin mode lets you see organizations that you're not a member of, AND
- now with PVP when you are logged out you can also see organizations you are not a member of

This is confusing for people with admin mode enabled when they are fully logged out. This situation results in everything being in a logged out state except the approval buttons and lack of PVP banner.

This PR clears admin mode when a user is logged out so if viewing a public project, it should always show the banner and disable the approval buttons.


cadeParade:
The tests passed in their respective places but github webhooks are down so didn't report.

https://buildkite.com/percy/percy-web/builds/3152

PR #859, opened 2018-10-22
Users screen: add a users count

unicornzero:
## Description
Adding a user count to the Users list

## Pictures
screen shot 2018-10-22 at 4 06 22 pm

## Reviewing this PR

_**Is Covered by Automated Tests?: no


app/templates/organizations/organization/users/index.hbs
```diff
@@ -7,7 +7,7 @@
   </p>
   <div class="bg-white rounded shadow-1 p-2 mb-2">
     <div class="flex justify-between items-center mb-2">
-      <h2 class="text-xl font-semibold mb-0">{{capitalize organization.name}} users</h2>
+      <h2 class="text-xl font-semibold mb-0">{{capitalize organization.name}} users ({{usersCount}})</h2>
```
cadeParade: I would probably use this, although the above method definitely works too:
```suggestion
      <h2 class="text-xl font-semibold mb-0">{{capitalize organization.name}} users ({{users.length}})</h2>
```
Routes deal with loading data, so the main things that you want to deal with in a route are the data that you get. Users count isn't really data that you've loaded it's more like a thing you get after you load the main data blocks.

If this was a component, I would not hesitate to set `usersCount` as a property, this just happens to be a route/controller/template combo which is a bit of a strange thing and eventually being deprecated, so I'd lean towards asking for length in the template in this case.

unicornzero:
We don't have UX on this feature yet. This might not even be the right change to make. I'm not quite sure.

cadeParade:
The update seems reasonable to me!

maprules1000:
@unicornzero rebasing should remove those percy diffs

PR #861, opened 2018-10-23
Recent webhook deliveries UI

paulfri:
## Description
Implement UI for viewing recent webhook events. We show the last 10 events delivered to a webhook configuration.

## Pictures
screen shot 2018-10-23 at 4 32 14 pm
screen shot 2018-10-23 at 4 32 27 pm
screen shot 2018-10-23 at 4 32 49 pm

## Reviewing this PR
Go to the webhook configuration form (within project settings) and check out the new views. Simply creating and updating a webhook configuration should create some events by default.

_**Is Covered by Automated Tests?: YES/NO**_Yes.


app/components/webhook-event.js
```diff
@@ -0,0 +1,59 @@
...
+    'Webhook delivery failed because of an unknown error. Please ' +
+    'contact support if this issue persists.',
+};
+
+export default Component.extend({
+  isOpen: false,
+
+  isRequest: true,
+
+  isSuccess: computed('webhookEvent', function() {
```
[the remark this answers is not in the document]
cadeParade: Also, it seems like you're recalculating this property here on this component, when it is already calculated on the `webhookEvents` model?  If they are the same, I would use the model version.
https://github.com/percy/percy-web/pull/861/files#diff-867d789c9b6656cabec02e1473e9a606R16

paulfri:
There was a visual bug caused by my merge with @PixelJanitor's stuff that Percy caught, so the screenshots are slightly out-of-date.

app/templates/components/projects/webhook-config-list.hbs
```diff
@@ -12,9 +12,17 @@
...
-        <div class="p-2 border-b" data-test-webhook-config-item>
+        <div
+          class="p-2 {{unless (eq webhookConfig filteredWebhookConfigs.lastObject) "border-b"}}"
+          data-test-webhook-config-item
+        >
           <div class="flex items-center justify-between">
             <div class="flex items-center mr-2">
+              {{status-dot
+                status=webhookConfig.status
+                class="mr-2"
```
cadeParade: I would rather not introduce this pattern. If we need there to be classes on the component div, we should
- have the component have the div (take off `tagName=""` and use `classes="mr-2") or
- put the class on the div inside the component or
- surround the component with a div with the class we want

app/templates/organization/project/webhooks/webhook-config.hbs
```diff
@@ -23,5 +21,21 @@
         saveSuccess="webhookConfigUpdated"
       }}
     </div>
+
+    {{#if model.webhookConfig.webhookEvents}}
```
cadeParade: It looks like, based on the model file, `webhookEvents` for a `webhookConfig` are async now. When asking for a relationship like this, we've always got to be aware of whether or not this data has loaded. I think this would become a race condition if the request for webhookEvents is very slow. The template would always pass the `if` because there would be a promise object that is present as `model.webhookConfig.webhookEvents` and then when it got to the `each`, there would be nothing. The template will not update automatically when a promise has finished, so to make this safer I would probably either ask for the webhookEvents in the model and block rendering on that data, or use the (`await` helper )[https://github.com/fivetanley/ember-promise-helpers] on line 25.

tests/integration/components/webhook-event-test.js
```diff
@@ -0,0 +1,111 @@
...
+      webhookEvent=webhookEvent
+      deliveryUrl=deliveryUrl
+    }}`);
+  });
+
+  it('displays webhook event data', function() {
+    percySnapshot(this.test.fullTitle());
+
+    expect(WebhookEvent.id).to.equal(webhookEvent.id);
+    expect(WebhookEvent.deliveryUrl).to.equal(`POST ${this.get('deliveryUrl')}`);
```
cadeParade: I would probably assign `deliveryUrl` to a variable like you did with `webhookEvent` here. This is because asking for `this.get('foo')` in a test is asking for a property on the test context. It's not actually asking the component instance. Since we don't have a lot of control over the test context, it's safer to compare against variables that we've set rather than something on the test context.

tests/integration/components/webhook-event-test.js
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  context('when detail view is opened', function() {
+    it('displays request headers and payload', function() {
+      webhookEvent = make('webhook-event', {
+        requestHeaders: {'X-Request': 'yep'},
+        requestPayload: 'request!!',
+      });
+      this.set('webhookEvent', webhookEvent);
+      this.render(hbs`{{webhook-event
```
cadeParade: This context runs after the `beforeEach` which also renders the component, so it looks like you're rendering two components with each test.

PR #873, opened 2018-10-30
[bugfix] Fix inifinity scroll links

maprules1000:
## Description
When scrolling through infinite builds sometimes the links for the newly loaded batch of builds breaks.

The root cause of the error is refreshing the project. Somehow the linked project at `build.project` ends up unavailable for the newly rendered build cards. Passing in the project directly fixes this issue.

## Reviewing this PR

Unfortunately, it's not feasible to reproduce the timing issue in tests :(


tests/integration/components/build-card-test.js
```diff
@@ -14,7 +15,11 @@ describe('Integration: BuildCard', function() {
 
   beforeEach(function() {
     setupFactoryGuy(this.container);
+    setupRouter(this.container);
```
maprules1000: The necessary container context is only available in `beforeEach`, not the `before/beforeAll` hook :-/

cadeParade: Could we put this in the relevant `describe` `beforeEach` (That would be line 26 I believe) so the setup is not required for every test? And also leave a comment above `setupRouter` that it's only needed for the url test.

Also, has this test ever failed? If you stop passing `project` in the template, would this test fail?

maprules1000: Can confirm that it does fail if there is no project passed into the component

Robdel12:
Looks like some factory data needs to be frozen. So close! ☺️

cadeParade:
Ok, @maprules1000 and I had a lengthy pairing session on this and learned that the build polling in `project-container.js` is interfering with the infinite scroll querying. 

One thing we tried was replacing the build query in the poller to use infinity model. This did work, and we saw the correct builds coming back from the server (including new ones!) but the template did not update with them (because the template was dependent on the model hook, which only runs once).
Then we tried setting builds property on the component after the reload finished with the new builds, but this was super broken -- there was not the right number of builds, they were unsorted, and a seemingly random selection of builds showed up in the list.

Another I tried was putting the poll action on the route and passing it down, so the component would call the route action when it wanted to refresh. Then the route action would set the build objects on the controller , and the template would read from those builds, rather than`model.builds`. This worked in theory but made all the build data disappear on refresh (although the cards were strangely all still there)

Overall, what I think is happening is that the way the builds are fetched (`project.get('builds').reload()`) is interfering with the infinity model. Asking for builds this way is saying to Ember "give me all the builds for this project, and all the builds that come back _are_ all the builds for this project", therefore, when the poll gets back, Ember believes the returned builds are the only builds the project should know about. So all but the first 100 builds become orphaned after then first refresh (which returns 100 builds).

We did not find a working solution to this, so we'll use @PixelJanitor's original proposal for explicitly passing `project` into the build card component as a slightly-less-than-optimal fix.

tests/integration/components/build-card-test.js
cadeParade: We talked about this on VC, but we found that the test was not setting up a relationship between project and build, and therefore the test would fail before the changes in this PR. However, project and build _should_ have a relationship (although this bug manifests in them not having a relationship) so setting up the test that way seems wrong. When the project/build relationship was set up correctly, the test passed before and after the changes in this PR. 

To fix this bug in a reasonable time, we'll remove this test and keep the rest.

PR #901, opened 2018-11-15
Make Builds Polling and Infinity Scrolling Play Nice

maprules1000:
## Description
Build polling was not playing nice with the ember infinity models that are used for infinite scrolling. I still don't know exactly what the issue is, but it boils down to the fact that infinity models are an array proxy wrapped with some sugar from `ember-infinity` to make everything run seamlessly. 

The fix is to make the store the source of truth and treat polling _and_ `ember-infinity` simply as a way to load data. This means that `builds` must be peeked from the store, filtered by project, and then ordered and "cleaned". They have to be "cleaned" to make sure only an ordered and continuous list of builds are shown on the page. If they are not "cleaned" then some random non-sequential builds show up at the end of the page: e.g. the 2nd to last build number could be 314 and the very last one might be 297 since we're getting all builds for the project in the store. 

## Reviewing this PR

_**Is Covered by Automated Tests?: YESish**_

This is covered by automated tests to a degree, but infinity data loading is tricky to reliably test.

The most important things to verify that can't easily be tested:
1. Make sure the first builds are automatically loaded when running a new project.
2. Make sure infinity scrolling works for projects with > 50 builds.
3. Make sure the infinity loader disappears when going from a project with > 50 builds to one with < 50, and comes back when going back to the first project.
4. Make sure the infinity loader reappears when going from a project with < 50 builds to one with > 50, and goes away when going back to the first project.

Also see https://github.com/ember-infinity/ember-infinity/issues/335 about fundamental polling issue.


app/routes/organization/project/index.js
```diff
@@ -48,11 +48,6 @@ export default Route.extend(ResetScrollMixin, {
         project_slug: project.get('slug'),
       };
       this.analytics.track('Project Viewed', organization, eventProperties);
-
-      // If transitioning back to this page after first load, background reload the builds.
-      if (project.get('builds').isFulfilled) {
-        project.get('builds').reload();
-      }
```
maprules1000: This is unnecessary since builds are reloaded by the infinity model when a user navigates to a project.

app/templates/components/project-container.hbs
```diff
@@ -59,14 +59,14 @@
...
   {{#if builds}}
-    <div class="build-list">
-      {{#each builds as |build|}}
-        {{build-card build=build project=project}}
-      {{/each}}
-      {{#infinity-loader infinityModel=builds hideOnInfinity=true}}
+    {{#each builds as |build|}}
+      {{build-card build=build project=project}}
+    {{/each}}
+    {{#if canLoadMore}}
```
maprules1000: The infinity-loader's internal show/hide logic is broken so this is a workaround of sorts.

tests/acceptance/project-test.js
```diff
@@ -335,24 +335,66 @@ describe('Acceptance: Project', function() {
...
       }
 
-      server.create('build', 'withSnapshots', {project, createdAt: _timeAgo(60, 'days')});
-      server.create('build', 'expired', {project, createdAt: _timeAgo(30, 'hours')});
-      server.create('build', 'failed', {project, createdAt: _timeAgo(3, 'hours')});
-      server.create('build', 'failedWithTimeout', {project, createdAt: _timeAgo(25, 'minutes')});
+      server.create('build', 'withSnapshots', {
+        project,
+        createdAt: _timeAgo(60, 'days'),
+        buildNumber: 1,
```
maprules1000: For some reason mirage is not great at reliably counting up in sequence. It might be a speed thing. Adding explicit build numbers more accurately simulates the actual api response.

app/components/project-container.js
```diff
@@ -31,9 +37,39 @@ export default Component.extend(PollingMixin, {
     return this._refresh();
   },
 
+  builds: computed('isRefreshing', 'infinityBuilds._loadingMore', function() {
```
maprules1000: This is where the magic happens. We no longer use the infinityModel as a source of truth to render the builds on the page. We use the builds in the store, but this means they have to be filtered by project, sorted, and then checked for continuity before being shown on the page.

app/components/project-container.js
```diff
@@ -31,9 +37,39 @@ export default Component.extend(PollingMixin, {
...
+    return this._sortAndCleanBuilds(filteredBuilds);
+  }),
+
   actions: {
     refresh() {
       this._refresh();
     },
   },
+
+  _sortAndCleanBuilds(builds) {
```
maprules1000: This orders the builds and checks that there are only builds with continuous build number: i.e. `5,4,3` not `5,4,3,1`. The latter state can come about because of the many different ways data is loaded into the store from the api.

app/components/project-container.js
```diff
@@ -31,9 +37,39 @@ export default Component.extend(PollingMixin, {
...
+    // sort the builds in decending order
+    const sortedBuilds = builds.sortBy('buildNumber').reverse();
+
+    let continuousBuilds = [];
+    for (let i = 0; i < sortedBuilds.length; i++) {
```
maprules1000: I would be so happy if someone could recommend another way to do this 🙏

app/components/project-container.js
cadeParade: Let's pull this out (because it is duplicated below) and then also it would be possible to write unit tests for this function.

app/components/project-container.js
```diff
@@ -1,28 +1,34 @@
...
   tagName: 'main',
   classNames: ['project-container flex-1 border-l pb-8'],
   attributeBindings: ['data-test-project-container'],
   'data-test-project-container': true,
+  buildsLimit: INFINITY_SCROLL_LIMIT,
 
-  _refresh() {
+  canLoadMore: computed.not('infinityBuilds.reachedInfinity'),
+
+  async _refresh() {
```
cadeParade: 🎉

app/components/project-container.js
```diff
@@ -1,28 +1,34 @@
...
-          .then(() => {
-            if (!this.isDestroyed) {
-              this.set('isRefreshing', false);
-            }
-          });
-      });
+    const project = await this.get('project').reload();
+    const buildCount = this.get('buildsLimit');
+
+    // run the initial 50 builds query again instead of reloading 100 builds
```
cadeParade: Could we put a little more context in this comment (that maybe references the route , or at least what is referred to by "again" and "reloading")

app/components/project-container.js
cadeParade: You might be able to `for..of` use https://stackoverflow.com/a/32101207

app/components/project-container.js
```diff
@@ -31,9 +37,39 @@ export default Component.extend(PollingMixin, {
...
+    // sort the builds in decending order
+    const sortedBuilds = builds.sortBy('buildNumber').reverse();
+
+    let continuousBuilds = [];
+    for (let i = 0; i < sortedBuilds.length; i++) {
+      const currentBuildNumber = sortedBuilds[i].get('buildNumber');
+
+      if (i === sortedBuilds.length - 1) {
+        continuousBuilds.push(sortedBuilds[i]);
+      } else if (currentBuildNumber - sortedBuilds[i + 1].get('buildNumber') === 1) {
```
cadeParade: Optional: One thing that might make this a bit more comprehensible could be breaking this check into a method with a descriptive name like `_isBuildNumDiffGreaterThanOne` or something like that... `_isBuildNumMoreThanOneAway`..
idk, but a well named method could get the gist across without having to dive into all the indexing.

app/components/project-container.js
cadeParade: Could you also add a comment describing why this is necessary (Because base builds are loaded along with every build and therefore there could be a big gap in the build numbers if all of them were displayed and that looks bad)?

tests/acceptance/build-test.js
```diff
@@ -211,7 +211,7 @@ describe('Acceptance: Build', function() {
     expect(BuildPage.isDiffsVisibleForAllSnapshots).to.equal(false);
 
     await BuildPage.clickProject();
-    await ProjectPage.builds(1).click();
+    await ProjectPage.builds().click();
```
cadeParade: It seems like this is clicking every build now?

app/templates/components/project-container.hbs
```diff
@@ -59,14 +59,14 @@
   {{loading-page}}
 {{else}}
   {{#if builds}}
-    <div class="build-list">
```
maprules1000: Need to add this back in

tests/acceptance/build-test.js
maprules1000: Yeah I'll looks back into this. I think it defaults to the first one, but then it would just be better to change that `1` in to `0`.

app/components/project-container.js
maprules1000: So now that it's abstracted to the utils and broken up into 2 functions I'm not really worried about this. Yes, it looks like something out of a high school freshman's very first computer science homework, but that's not a problem because it works and I got a 💯on that assignment.

app/components/project-container.js
```diff
@@ -1,28 +1,36 @@
...
   classNames: ['project-container flex-1 border-l pb-8'],
   attributeBindings: ['data-test-project-container'],
   'data-test-project-container': true,
+  buildsLimit: INFINITY_SCROLL_LIMIT,
 
-  _refresh() {
+  canLoadMore: computed.not('infinityBuilds.reachedInfinity'),
+
+  async _refresh() {
     this.set('isRefreshing', true);
```
cadeParade: Not required for this PR but just realized since this is now using ember-concurrency, we could use the `isRunning` property...it might also remove the need to check if the component is destroyed?

app/lib/paginated-ember-infinity-model.js
```diff
@@ -7,15 +8,29 @@ const ExtendedInfinityModel = InfinityModel.extend({
...
     // if the server response is less than the request assume we've loaded all the data
     // >= is used to prevent a race condition where we might load more than the request number of
     // projects due to chunks of data being loaded in many ways.
-    // assumes list from server is ordered starting with most recent build
     const canLoadMore = builds.get('length') >= params.get('perPage');
 
     set(this, 'canLoadMore', canLoadMore);
-    set(this, '_cursor', builds.get('lastObject.id'));
+
+    // need to set the pagination cursor based on the oldest build already loaded into the store,
```
cadeParade: This is slightly wrong now...it sets the cursor based on the oldest _continuous_ build?

tests/integration/components/project-container-test.js
```diff
@@ -100,6 +104,7 @@ describe('Integration: ProjectContainer', function() {
       await this.render(hbs`{{project-container
         project=project
         builds=infinityBuilds
```
cadeParade: These all have both `builds` and `infinityBuilds` -- but it looks like only `infinityBuilds` is being passed to `project-container`?
https://github.com/percy/percy-web/pull/901/files#diff-ab50d10e82844df10a6af25e173647feR22

cadeParade:
Also, here is an example of how polling might be tested: https://github.com/percy/percy-web/blob/master/tests/acceptance/build-test.js#L92, it might be useful to write a case like this to make sure project polling isn't broken...
LMK if you're interested and want to pair on it

app/lib/paginated-ember-infinity-model.js
maprules1000: Yes, otherwise some builds might be skipped.

maprules1000: If the buildNumbers go `79,78,77,76,70` and we set the cursor with `70` then builds `75,74,73,72,71` will never be loaded.

cadeParade: Right, the comment just says it sets it on the _oldest_ build, which is slightly incorrect now.

maprules1000: Ah. You make valid point. I fix.

tests/integration/components/project-container-test.js
maprules1000: Yes, because this is an integration tests the Ember Store isn't setup so the `builds` computed property, which peeks builds in the store, doesn't run properly.

PR #914, opened 2018-11-27
Update Resource Widget

maprules1000:
## Description
Updates the resource widget based on internal discussions
- Prevent `Static Sites` from wrapping
- Remove deprecated `Percy Anywhere` framework
- Add `Puppeteer` under the ruby language
- Change button location
- Add analytics to track when users click on the docs and demo project buttons
- Fix typo


app/components/resource-widget.js
```diff
@@ -78,5 +78,19 @@ export default Component.extend({
         this.set('selectedFramework', null);
       }
     },
+
+    trackDocsVisit() {
+      const organization = this.get('organization');
+      const request = this.get('selectedFramework').toLowerCase();
+
+      this.get('analytics').track('Framework Docs Visited', organization, {request});
```
cadeParade: Dunno if this was discussed, but `request` is usually talking about some request/response data, rather than a framework requested. Maybe `frameworkRequested`? But anyway, not that strong of feelings about it.

maprules1000: I can rename it really quick.

PR #919, opened 2018-11-30
Update org creation screen

cadeParade:
## Description
This PR adds an email field to the org creation form. 

The main functional addition was adding an `isFirstOrganization` property passed down from the route (because it requires a query). All of the user email behavior depends on this property being `true`.

The `organization-new` form now has a user email input. This field is backed by a new changeset generated from `currentUser`. Submitting the form saves both the organization (via the current `form/base` `save` method and also POSTS to the user endpoint to update the user's email.

Additionally, I've refactored the model for the form a bit -- the new organization was previously created within the form component, but this made testing quite hard (because you have to stub the save method) and since it was within the component it was difficult (and bad form) to reach into the test component and stub a property on an internal model.

Instead, I've created the new organization in the `organizations/new` route, and called it part of the `model` which is where data for routes usually goes. Additionally, it allows us to get all the information we need from the github subscription data in the route and not have to pass it further along to several components.

The tests are a bit funny because the async relationship computed property on user model for identities causes infinity problems in tests, so to get something out the door, I've taken some liberties. But this relationship definitely needs to be refactored, we've wasted enough of our life on it.

## Questions
- What should be the behavior if user save fails?
    - Should it block the flow? **NO**
- Should email prompt be shown only to users who sign up with GitHub? **YES**

## Pictures
https://percy.io/percy/percy-web/builds/1209279/view/77236353/1280?browser=chrome&mode=diff

## Reviewing this PR
_**Is Covered by Automated Tests?: YES, fairly well**_


djones:
> Should email prompt be shown only to users who sign up with GitHub?

Yes, I think so. Would like to avoid extra steps during sign up that are not relevant.

tests/acceptance/organization-test.js
```diff
@@ -57,6 +58,25 @@ describe('Acceptance: Organization', function() {
...
 
+    it('can create new organization and user email when creating first org', async function() {
+      // simulate the user having no organizations
+      server.db.organizationUsers.remove(1);
+      server.db.organizations.remove(1);
+      await visit('/organizations/new/');
+      // I am truly sorry
+      this.owner.__container__
+        .lookup('service:session')
+        .set('currentUser.githubIdentity', {provider: 'github'});
```
maprules1000: I accept this reality.

app/components/forms/organization-new.js
```diff
@@ -1,36 +1,63 @@
...
+      // Save the organization
+      this.send('save');
+      // If there's an email field, also save that
+      if (this.get('isFirstOrganization')) {
+        try {
+          const newEmail = this.get('userChangeset.email');
+          await this.get('userChangeset').save();
+          this.get('flashMessages').success(`Check ${newEmail} to verify it!`);
+        } catch (e) {
+          this.get('flashMessages').danger(`NO: ${e}`);
```
maprules1000: Maybe a nicer error message?

app/components/forms/organization-new.js
```diff
@@ -1,36 +1,63 @@
...
+  userChangeset: computed('currentUser', function() {
+    const validator = UserEmailValidations;
+    this.set('currentUser.email', '');
+    return new Changeset(this.get('currentUser'), lookupValidator(validator), validator);
   }),
+
+  actions: {
+    async customSave() {
+      // Save the organization
+      this.send('save');
```
maprules1000: Hold up. Ok, so I begrudgingly understand that this is needed because of the way the form components are composed, but, a thought: could you add a private helper function that calls the base save action in the base form component and then call that helper function from this action? It's verbose, but less gross. (Yes, that was intentional 😄). 

Anything that would be more explicit than throwing out `send` or `sendAction` would be better, but I'm honestly not even sure if those will work. 

If they don't, then adding a comment saying where that `save` functionality is expected to be would be very useful. I mean, I know that it's on the base form component which this particular form is extending, but this action passing isn't as common as the deprecated idk-send-back-up-to-parent-component-and-hope-for-the-best pattern.

app/components/forms/organization-new.js
maprules1000: Or even an error message that shows a link to the profile page that opens in a new background tab?

cadeParade: Very reasonable

PR #926, opened 2018-12-10
Percy power selects

PixelJanitor:
screen shot 2018-12-10 at 5 28 51 pm

screen shot 2018-12-10 at 5 27 35 pm


cadeParade:
There are some interesting differences: 
https://percy.io/percy/percy-web/builds/1241372/view/79646856/1280?browser=chrome&mode=diff
https://percy.io/percy/percy-web/builds/1241372/view/79647790/1280?browser=chrome&mode=diff

and also in one like this: https://percy.io/percy/percy-web/builds/1245347/view/80001350/1280?browser=chrome&mode=diff, the dropdown is pushed to the right, is this intentional?

PR #929, opened 2018-12-12
Show duplicate build number in project container

maprules1000:
## Description
Juriy from Wework reached out to Tim (via email with subject "Cancelling Percy builds") about infinite scrolling being broken for https://percy.io/WeWork/Spacestation, but only once an arbitrary page was reached.

It turns out that builds 1251825 and 1251826 shared the same build id (18114) and this duplication was causing all builds below the duplicated 18114 to not be shown in the interface. 

[This function](https://github.com/percy/percy-web/blob/915c5be4b8/app/lib/utils.js#L134) was checking to make sure the build numbers were sequential. It is now updated to accomodate for the rare case when build number are the same.

A test was also added to cover this case.

## Reviewing this PR


app/lib/utils.js
```diff
@@ -134,6 +134,6 @@ export default {
   _isNextBuildNumberOneLess(sortedBuilds, i) {
     const currentBuildNumber = sortedBuilds[i].get('buildNumber');
     const previousBuildNumber = sortedBuilds[i - 1].get('buildNumber');
-    return previousBuildNumber - currentBuildNumber === 1;
+    return previousBuildNumber - currentBuildNumber <= 1;
```
timhaines: @maprules1000 I think this line deserves a comment explaining what it handles - it's a bit non-obvious, and a new dev to the project could easily mistake it for an error.  What do you think?

maprules1000: Yup! Good call

PR #930, opened 2018-12-12
Implement new Users and Invites UI

unicornzero:
## Description
This is a big PR. It was originally several stories, but as the feature evolved, it was not feasible to break this down into smaller PRs. Sorry! <3

In short, we've completely redone the organization users interface. This includes moving UI code from templates to components, new components, a bunch of tests, and a lot more. 

A user can see all users and invites in the organization or leave the organization. Admins can invite users if the organization don't have a seat limit that they've exceeded. Admins can also copy invite urls, cancel invites, and change user permissions.

## Pictures

No section headers when there are no invites:
screen shot 2018-12-12 at 2 03 32 pm

With invites:
screen shot 2018-12-12 at 2 05 01 pm

When on the /invite path when invites are allowed:
screen shot 2018-12-12 at 2 05 20 pm

Organization has a seat limit and that seat limit has been met or exceeded:
screen shot 2018-12-12 at 2 05 55 pm
(Please ignore the "trial" warning at the top of the screens. Dev environment issue.)


app/components/percy-btn.js
```diff
@@ -0,0 +1,13 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+  tagName: '',
+
```
cadeParade: A few thoughts about this component (cc @PixelJanitor ):

- Let's add some sane defaults here for all the properties that could be passed in. A way to document the API and also set some defaults.

- Let's not use `onWhatever` names for actions. In some cases (like `onclick`), these actions are implemented in native javascript and may interfere with native event bubbling which can be a big, but subtle, problem. Instead we could use something like `handleTap`.

- In the same vein, passing in`disabled` seems like it is a native HTML attribute, where in this case it is not. It is an Ember property being passed to a component. A preferable way to handle it IMO would be to pass `isDisabled` into the component (which name makes it clear that it is not a native HTML attribute), and have `disabled={{isDisabled}}` inside the component template. Additionally, both Ember and HTML add some implicit behavior to elements with the `disabled` attribute, so it is best to be very very explicit about where we want that applied.

- Why the choice of `tap`? `tap` in a web development context usually refers to touch screen events specifically, and while we do want this action to happen when the button is pressed, the name is a bit misleading. I was looking for the separate `onClick` method to see what the different behavior would be for click event instead of tap.

- Should `onTap` be optional (as it is in the current state)? Does it make sense to have a button show up that doesn't do anything?

app/templates/components/percy-btn.hbs
```diff
@@ -0,0 +1,15 @@
+<button
+  data-test-percy-btn
+  type="{{if type type "button"}}"
+  name="button"
+  disabled={{disabled}}
+  class="
+    percy-btn
+    {{class}}
+    {{if primary-btn "percy-btn-primary"}}
```
cadeParade: These properties also for the `why-kebab-case-all-of-a-sudden` discussion :)

tests/acceptance/join-test.js
```diff
@@ -9,15 +9,22 @@ describe('Acceptance: Join', function() {
...
+    const user = server.create('user', {name: 'Basil Cat'});
     server.create('invite', {
       id: 'valid-code',
       organization: anotherOrganization,
+      fromUser: user,
+    });
+    server.create('invite', {
+      id: 'expired-code',
+      isExpired: true,
+      organization: anotherOrganization,
```
cadeParade: You may not need the organization and fromUser link anymore? Not sure though.

tests/acceptance/manage-users-test.js
```diff
@@ -0,0 +1,147 @@
...
+      expect(UsersHeader.seatCount.text).to.equal(text);
+      expect(UsersHeader.billingLink.isVisible).to.equal(true);
+      expect(UsersHeader.supportLink.isVisible).to.equal(true);
+      expect(UsersHeader.inviteButton.text).to.equal('Invite new users');
+    });
+
+    it('renders all users', async function() {
+      const numberOfUsers = 5;
+      const users = server.createList('user', numberOfUsers);
+      users.map(user => {
```
cadeParade: This implementation gets the job done and this is not a very important point, but for the learnings it is returning an un-needed array (bc `map` returns an array).
An alternative which would fit the purpose _slightly_ better would be `users.forEach(user => {do the thing})` which will just execute the thing in the function but not return any array.

tests/acceptance/manage-users-test.js
```diff
@@ -0,0 +1,147 @@
...
+  let adminUser;
+  let numberOfInvites;
+  let confirmationAlertStub;
+
+  setupSession(function(server) {
+    organizationName = 'Meow Mediaworks';
+    seatsUsed = 3;
+    seatLimit = 10;
+    seatsRemaining = 10;
+    // using 'withSponsoredSubscription' to avoid a trail expiration error in the snapshots
```
cadeParade: `trail` -> `trial`

tests/acceptance/manage-users-test.js
```diff
@@ -0,0 +1,147 @@
...
+      seatsRemaining,
+      name: organizationName,
+    });
+    adminUser = server.create('user');
+    server.create('organizationUser', {
+      organization,
+      user: adminUser,
+      role: 'admin',
+    });
+    numberOfInvites = 2;
```
cadeParade: P10: I would probably move this declaration up to where the rest of the constants are declared.

tests/acceptance/manage-users-test.js
```diff
@@ -0,0 +1,147 @@
...
+      fromUser: adminUser,
+    });
+  });
+
+  describe('when currentUser is an Admin', function() {
+    beforeEach(function() {
+      const sessionServiceStub = Service.extend({
+        currentUser: adminUser,
+      });
+      this.owner.register('service:session', sessionServiceStub, 'sessionService');
```
cadeParade: This should not be necessary in an acceptance test. If it is necessary, let's work together to refactor it out.

tests/acceptance/manage-users-test.js
```diff
@@ -0,0 +1,147 @@
...
+
+        await UsersHeader.sendInvitesButton.click();
+
+        expect(currentRouteName()).to.equal('organizations.organization.users.index');
+        expect(ManageUsersList.inviteCards().count).to.equal(numberOfInvites + 1);
+        await percySnapshot(this.test.fullTitle());
+      });
+
+      it('cancelling from the invite form returns user to users route', async function() {
+        await visit(`/organizations/${organization.slug}/users`);
```
cadeParade: You can put visit calls in page objects. ex: https://github.com/percy/percy-web/blob/master/tests/pages/project-page.js#L21

tests/integration/components/organizations/manage-users-list-test.js
```diff
@@ -0,0 +1,149 @@
...
+  let organization;
+  let organizationUsers;
+  let invites;
+  let numberOfUsers;
+
+  beforeEach(function() {
+    setupFactoryGuy(this);
+    ManageUsersList.setContext(this);
+    organization = make('organization');
+    numberOfUsers = 5;
```
cadeParade: Hoist dat variable plz

tests/integration/components/organizations/manage-users-list-test.js
```diff
@@ -0,0 +1,149 @@
+import {it, describe, beforeEach} from 'mocha';
+import {make, makeList} from 'ember-data-factory-guy';
+import {percySnapshot} from 'ember-percy';
+import {setupRenderingTest} from 'ember-mocha';
+import hbs from 'htmlbars-inline-precompile';
+import moment from 'moment';
+import setupFactoryGuy from 'percy-web/tests/helpers/setup-factory-guy';
+import ManageUsersList from 'percy-web/tests/pages/components/organizations/manage-users-list';
+
```
cadeParade: These tests are wonderful and thorough, but I would be remiss in not pointing out that Percy's strength is rendering tests, and a lot of this sort of thing can be shunted to Percy snapshots. I'd encourage you to think about snapshot strategy in the future, it might save you some time!

tests/integration/components/organizations/user-card-test.js
```diff
@@ -0,0 +1,346 @@
...
+        organization.set('currentUserIsAdmin', true);
+      });
+
+      describe('Leave organization', function() {
+        it('requests confirmation and calls destroyRecord', async function() {
+          await this.render(
+            hbs`{{organizations/user-card organizationUser=organizationAdminUser}}`,
+          );
+          const orgUserStub = sinon.stub(organizationAdminUser, 'destroyRecord');
+          find('.user-card-actions').classList.add('opacity-100');
```
cadeParade: Is this necessary? If so, can you add a comment as to why, as you did in the previous case? (there is no percy snapshot in this test)

tests/integration/components/organizations/users-header-test.js
```diff
@@ -0,0 +1,221 @@
...
+
+  beforeEach(function() {
+    setupFactoryGuy(this);
+    UsersHeader.setContext(this);
+    organization = make('organization', {
+      name: organizationName,
+      seatsUsed,
+      seatLimit,
+    });
+    this.set('isInviteFormAllowed', false);
```
cadeParade: Let's put all the `this.set`s together, so we can clearly see what we are assigning to the test context.

tests/integration/components/organizations/users-header-test.js
```diff
@@ -0,0 +1,221 @@
...
+    organization.set('hasSeatsRemaining', true);
```
cadeParade: Can this property be added to line 22?
`make('organization', {name: 'sldjf', hasSeatsRemaining: true, etc})`?

tests/integration/components/organizations/users-header-test.js
```diff
@@ -0,0 +1,221 @@
...
+    this.set('organization', organization);
+  });
+
+  it('displays basic information', async function() {
+    organization.set('currentUserIsAdmin', true);
```
cadeParade: Is this necessary for this case?

tests/integration/components/organizations/users-header-test.js
```diff
@@ -0,0 +1,221 @@
...
+    expect(UsersHeader.supportLink.isVisible).to.equal(true);
+    expect(UsersHeader.inviteButton.text).to.equal('Invite new users');
+  });
+
+  describe('when user is admin', function() {
+    beforeEach(async function() {
+      organization.set('currentUserIsAdmin', true);
+    });
+
+    describe('when isInviteFormAllowed=false (on /users page)', function() {
```
cadeParade: Probably don't need to mention `on /users page`. The component does not care about what route it's on, it only cares what the value of that property is.

app/models/organization.js
```diff
@@ -52,6 +53,16 @@ export default DS.Model.extend({
 
   organizationUsers: DS.hasMany('organization-user'),
 
+  seatLimit: DS.attr(),
+  seatsUsed: DS.attr(),
+  seatsRemaining: DS.attr(),
+
+  // when you remove seats, it's not recalculating this for the associated form
```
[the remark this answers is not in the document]
cadeParade: Also if you still need a greater-than comparator, there is a computed macro `gt` (https://www.emberjs.com/api/ember/3.4/functions/@ember%2Fobject%2Fcomputed/gt) which can do this for you. Just remember the second argument must always be a literal value.

cadeParade:
Also, please check out the percy build and approve ones that are correct. There are lots of diffs!

app/components/forms/organization-invite.js
```diff
@@ -30,6 +36,8 @@ export default BaseFormComponent.extend({
         () => {
           // Fully reset the model + changeset if saved successfully.
           this.set('model', this.newModel());
+          this.get('store').query('invite', {organization: this.get('organization')});
+          this.get('router').transitionTo('organizations.organization.users');
```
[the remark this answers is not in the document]
unicornzero: AWESOME!

My tests just validated that I can replace:
```
          this.set('model', this.newModel());
          this.get('store').query('invite', {organization: this.get('organization')});
          this.get('router').transitionTo('organizations.organization.users');
```

With just this!
```
          this.get('router').transitionTo(
            'organizations.organization.users',
            this.get('organization.id'),
          );
```

A++ on pointing this out and a fantastic explanation.

app/components/organizations/invite-card.js
```diff
@@ -0,0 +1,42 @@
...
+  }),
+  isAdmin: alias('organization.currentUserIsAdmin'),
+  store: service(),
+  actions: {
+    onCopyInviteUrlToClipboard() {
+      this.get('flashMessages').success('Invite URL was copied to your clipboard');
+    },
+    cancelInvite() {
+      const invite = this.get('invite');
+      const orgId = this.get('organization').get('id');
```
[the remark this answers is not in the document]
unicornzero: Perfect example of something I missed when reviewing my own code, of something I didn't know when I first started this feature. I'm glad to have other eyes on this now! :D

app/components/organizations/invite-card.js
```diff
@@ -0,0 +1,42 @@
...
+
+      const confirmationMessage = `Are you sure you want to cancel the invitation to
```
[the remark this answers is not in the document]
unicornzero: Sure! I'd originally thought of it as being part of the if (I broke it out because it was too long for the if to be all one line) but I can see why it might read better with the line break moved down one. Thanks!

app/templates/components/organizations/manage-users-list.hbs
```diff
@@ -0,0 +1,26 @@
+<div data-test-manage-users-list>
+  <div data-test-manage-users-column-headers class="bg-white flex px-4 mb-sm border-b z-1">
+    <div style="min-width: 200px" class="flex-1 text-sm text-gray-300 font-semibold pr-2">Name</div>
```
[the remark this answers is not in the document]
unicornzero: I'm not sure if this was me or @PixelJanitor so I'll defer to his styling expertise.

app/templates/components/organizations/invite-card.hbs
```diff
@@ -0,0 +1,43 @@
...
+      }}
+      <div class="truncate">
+        <div data-test-invite-email class="font-semibold leading-none truncate">
+          {{invite.email}}
+        </div>
+        <div class="text-sm text-gray-400 italic truncate">Expires <span data-test-invite-expiration>{{moment-format invite.expiresAt 'MMM D, Y'}}</span></div>
+      </div>
+    </div>
+    <div data-test-invite-role style="min-width: 100px" class="flex-1 leading-none">
+      <LabelTag @primary-tag={{eq "Administrator" invite.roleTitle}}>
```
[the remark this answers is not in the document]
unicornzero: I want to be excited, but I don't know what this comment means. :)

app/models/organization.js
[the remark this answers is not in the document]
unicornzero: Ah, nice, looks like `gt` handles null how I want it to. `hasSeatsRemaining: gt('seatsRemaining', 0),` works.

app/models/organization.js
```diff
@@ -52,6 +53,16 @@ export default DS.Model.extend({
...
+  hasSeatsRemaining: computed('seatsRemaining', function() {
+    // If seatsRemaining is null, this will always return false
+    return Number(this.get('seatsRemaining')) > 0;
```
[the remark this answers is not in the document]
unicornzero: I did try `parseInt` in this context but wasn't crazy about the NaN result for a comparison. But, the ember `gt` helper eliminates the need for either. Yay!

app/routes/organizations/organization/users/index.js
```diff
@@ -1,29 +1,15 @@
...
-    });
+    return this.modelFor('organizations.organization');
   },
 
   setupController(controller, resolvedModel) {
     controller.setProperties({
-      users: resolvedModel.users,
-      organization: resolvedModel.organization,
+      organization: resolvedModel,
+      isInviteFormAllowed: false,
```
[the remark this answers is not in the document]
unicornzero: Oh, yes! That's right! Now that it's a simple boolean value I can just add that in.

app/templates/components/forms/organization-invite.hbs
```diff
@@ -3,30 +3,39 @@
     {{form-fields/textarea
         property="emails"
         title="Email addresses"
+        id="inviteFormEmailAddresses"
         changeset=changeset
         autofocus=true
+        placeholder="Separate each email with a space or a new line. Invitations expire after one week."
         validateProperty=(action "validateProperty" changeset "emails")
+        keyUp=(action "resetSaveButton")
```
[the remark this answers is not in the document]
unicornzero: If you fill in the form badly and click the `Send Invites` button, you see error text below the form and the button state changes from purple to red and the text changes to `Failed`. It was confusing that you used to have to click `Failed` after your edited your typo and wanted to retry sending invites. Now, if you have a failed request, when you start editing the input again (correcting your error), the button changes back to purple `Send Invites` to make it more clear you can click it to try sending again.

tests/acceptance/join-test.js
unicornzero: We don't need organization now. It gets weird if I take out fromUser though :)

tests/acceptance/manage-users-test.js
unicornzero: Thanks for the learnings!

tests/acceptance/manage-users-test.js
unicornzero: You're right. Looks like it was an early remnant of me trying to make things work that wasn't actually part of making things work. Deleted lines 82-85

app/templates/components/label-tag.hbs
```diff
@@ -1,3 +1,8 @@
-<div class="label-tag ml-1">
-  {{labelText}}
+<div
```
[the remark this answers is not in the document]
PixelJanitor: slack-imgs

app/components/percy-btn.js
PixelJanitor: While I'm making these adjustments, it's coming back to me why I did what I did in some of these cases. When calling a component, using `disabled=` is natural since it's a common html element attribute. A dev wouldn't need to go looking through the docs to see what prop they need to use for disabled if it's just the expected attribute. Just like we don't pass something like `classNames=` to the component. It's very natural to just write `class=` and expect it to add those styling classes. So my intention was to make it simpler and more predictable. The `copy-button` does this the same way too, and when I was updating the `percy-btn` props, the `copy-btn` reminded me why I did this. Thoughts?

PixelJanitor: Any use of the word tap should be in reference to touch devices only. `handleClick` would probably be a better prop here unless we can get away with the native `onClick` :)

tests/integration/components/organizations/manage-users-list-test.js
unicornzero: Yes. That's a good point. Which of my tests seem particularly redundant? I could delete some if it adds to the cost of maintaining them.

tests/integration/components/organizations/users-header-test.js
unicornzero: Nope! You have a *very* good eye! :)

app/components/percy-btn.js
unicornzero: I named `onTap` and am happy to have us change its name to anything else. 

The name itself came out of habit as we had cause to need to build a click/tap wrapper for our app (which was mostly used on mobile devices) and added `onTap` handlers to everything. I knew I wanted to avoid a name collision with `onClick`, so I just picked the name of what came to mind. I'm 100% fine with changing it.

@cadeParade what do you think of @PixelJanitor 's `handleClick` suggestion?

tests/integration/components/organizations/manage-users-list-test.js
cadeParade: Well, basically all of the assertions in 'renders a user correctly' below could be deferred to Percy. All of these things test what is displayed on the screen, so any change in those properties would also show up in Percy.
I don't subscribe to depending wholly on Percy for tests -- for one, the feedback cycle is much longer than local unit tests, and for two, there are some things we need to test that are not visual. 
My usual approach would be snapshot the component in a few of the most important states, which would cover things like displaying texts, avatars, etc, then write some assertions for things that are changing in the logic of my component (disabled states, dropdowns, if some text changes, etc). 
In this particular case I am not too worried about maintenance cost. The properties you're testing for should not change very often and are pretty easy to figure out where they come from if they do change.

app/components/percy-btn.js
cadeParade: On the topic of `disabled`, I still think it is more clear to use `isDisabled` or another indicator that it is not the actual property that is making a thing disabled. For instance, you might want to use the `isDisabled` property for some logic in the JS file for more than setting the actual attribute on an element.

app/templates/components/organizations/manage-users-list.hbs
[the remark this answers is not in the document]
PixelJanitor: This was definitely me. This is a style specific to this component, so no need to break it out into a css class to manage elsewhere. There are a few of these in percy cases scattered, and there's no need to be afraid of using inlines for single use cases. It doesn't cause a maintainability issue since it's only relevant to it's specific use. Anytime a style like this is shared I create a class for it for sure.

app/templates/components/user-avatar.hbs
```diff
@@ -1,5 +1,7 @@
-{{#if (or showPlaceholder (not user))}}
-  <img src="images/placeholder-avatar.jpg">
-{{else}}
-  <img src="{{user.avatarUrl}}">
-{{/if}}
+<div data-test-user-avatar style="width: 40px; height: 40px;" class="rounded-full overflow-hidden {{class}}">
```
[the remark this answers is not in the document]
PixelJanitor: See other msg

app/templates/components/forms/organization-invite.hbs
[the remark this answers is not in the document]
unicornzero: The errors come from entering the text incorrectly, so I think we're okay with just putting it on the textarea. 

Also, checkboxes don't seem to work with keyUp and I'm not sure what to use there or if it's worth trying to find something else that would

app/components/percy-btn.js
unicornzero: I've tackled `handleClick`. I've not tackled `disabled` vs `isDisabled`. @PixelJanitor ?

PixelJanitor: I pushed commits earlier that fixed handlClick and the isDisabled stuff 👍

tests/integration/components/organizations/manage-users-list-test.js
unicornzero: Thanks! That's helpful.

app/templates/components/organizations/user-card.hbs
```diff
@@ -1,30 +1,40 @@
-<div data-test-user-card class="user-card border-b cursor-pointer">
-  <div class="p-2 flex items-center" {{action "toggleExpanded"}}>
-    {{user-avatar user=organizationUser.user linked=true class="mr-1"}}
-    <div>
-      <div class="leading-none mb-sm font-semibold">
+<div data-test-user-card class="user-card bg-white border-b">
+  <div class="px-4 py-3 flex items-center">
+    <div style="min-width: 200px" class="flex flex-1 items-center pr-2">
+      <UserAvatar @user={{organizationUser.user}} @class="flex-none mr-2" @tagName="" />
```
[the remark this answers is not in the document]
PixelJanitor: Angled brackets all converted back to curly's!

app/templates/components/label-tag.hbs
```diff
@@ -1,3 +1,8 @@
...
+  class="
+    label-tag
+    {{if primary-tag "label-tag-primary"}}
```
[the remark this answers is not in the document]
PixelJanitor: changed

app/components/percy-btn.js
PixelJanitor: percy-btn defaults are something I agree with, but will be addressed in a separate PR.

app/templates/components/percy-btn.hbs
PixelJanitor: I'm going to address these in a Percy-btn PR. Would like to discuss some naming patterns for all of these props but don't want to block this PR

app/components/organizations/user-card.js
```diff
@@ -1,18 +1,45 @@
...
+  currentUser: readOnly('session.currentUser'),
+  role: readOnly('organizationUser.role'),
+  roleTitle: readOnly('organizationUser.roleTitle'),
+  isAdmin: readOnly('organizationUser.organization.currentUserIsAdmin'),
+
+  deleteNameText: computed(function() {
+    if (this.get('isCurrentUser')) {
+      return 'yourself';
+    } else {
+      return this.get('organizationUser').user.name;
```
cadeParade: This syntax works (`.user.name`) because Ember is in the middle of switching away from the `get` syntax, but this syntax does not work in every case: see https://www.emberjs.com/blog/2018/04/13/ember-3-1-released.html#toc_es5-getters-for-computed-properties-2-of-4
Let's change this to use `get` soon so we can be consistent and safe.

unicornzero: Will do! Thanks! I'll put it in my next PR!

PR #931, opened 2018-12-13
Fix user/identities relationship

cadeParade:
## Description
This PR has all the writeup of this: https://github.com/percy/percy-web/pull/919 PLUS

The `identities` relationship is now fetched in the route of the org creation screen. The `identities` relationship originally caused this feature grief because the computed properties are async. I've refactored this feature to no longer depend on the async identity computed properties but instead load them explicitly in the route so we can know they're there.

## Reviewing this PR


app/components/forms/organization-new.js
```diff
@@ -1,36 +1,70 @@
...
+  userChangeset: computed('currentUser', function() {
+    const validator = UserEmailValidations;
+    this.set('currentUser.email', '');
+    return new Changeset(this.get('currentUser'), lookupValidator(validator), validator);
   }),
+
+  actions: {
+    async customSave() {
+      // Save the organization
+      this.send('save');
```
maprules1000: Could you add a comment here to say this is sending 'save' to the base form component?

app/models/identity.js
```diff
@@ -1,8 +1,12 @@
...
+import {equal} from '@ember/object/computed';
 
 export default DS.Model.extend({
   user: DS.belongsTo('user', {async: false}),
   provider: DS.attr(),
   uid: DS.attr(),
   nickname: DS.attr(),
+
+  isGithubIdentity: equal('provider', 'github'),
+  isAuth0Identity: equal('provider', 'auth0'),
```
maprules1000: 🙌 This is a much more sensible place for these checks to live

app/components/forms/organization-new.js
```diff
@@ -1,36 +1,70 @@
...
+      // If there's an email field, also save that
+      if (this.get('isFirstOrganization')) {
+        try {
+          const newEmail = this.get('userChangeset.email');
+          await this.get('userChangeset').save();
+          this.get('flashMessages').success(`Check ${newEmail} to verify it!`);
```
maprules1000: Actually, could you also add a simple analytics event here? I'm curious about how often this field will be used.

PR #936, opened 2018-12-17
Show branch selector even without SCM setup

maprules1000:
## Description
Shows the branch selector in the project container even if the organization does not have a source code management system connected.


cadeParade:
Actually, what about the case where there are no branches? I believe that is why I put it behind the scm  condition, I had assumed that if there was no scm there would be no branches. But will the dropdown still show up when there are no branches? I think it should not.

maprules1000:
@cadeParade what if we hide the branch selector when the number of branches is 1 or less?

cadeParade:
SGMT

PR #937, opened 2018-12-17
Add component defaults and camel casing to `percy-btn` props

PixelJanitor:


unicornzero:
Looks like this may have had some unintended side-effects of turning some of the users ui buttons purple: https://percy.io/percy/percy-web/builds/1264913/view/81556398/1280?mode=diff&browser=chrome&snapshot=81556398

PixelJanitor:
Yep! Those fixes are in progress. I wondered why our cancel button was so prominent so I was glad to see that it was something I did and wasn't on master :)

PR #940, opened 2018-12-18
When user leaves an org, redirect to their default org

unicornzero:
## Description
When a user leaves an organization, the app used to leave them on the users page of the organization they had just left. Instead, we can redirect them to their default organization.


app/components/organizations/user-card.js
```diff
@@ -35,6 +36,9 @@ export default Component.extend({
         return;
       }
       this.get('organizationUser').destroyRecord();
+      if (this.get('isCurrentUser')) {
+        this.get('router').transitionTo('index');
```
cadeParade: Is index the best place to take them? Maybe we could take them to another one of their organizations if they have one? 
You could do this with something like this:
```
this.get('router').transitionTo('default-org');
```

That should theoretically take the user to the first item in their organization list.
LMK if this does not work, happy to help!

tests/integration/components/organizations/user-card-test.js
```diff
@@ -198,23 +198,6 @@ describe('Integration: UserCard', function() {
         organization.set('currentUserIsAdmin', true);
       });
 
-      describe('Leave organization', function() {
```
cadeParade: These tests seem useful still?

app/components/organizations/user-card.js
unicornzero: Oh, I do like that better. I will give that a go.

unicornzero: Works like a champ! Much nicer. Thanks!

tests/integration/components/organizations/user-card-test.js
unicornzero: We talked through this and discovered that stubbing transitionTo on the router was particularly tricky because it doesn't behave like a "normal" service. So, we're going to leave these tests out.

tests/acceptance/manage-users-test.js
```diff
@@ -112,6 +112,19 @@ describe('Acceptance: ManageUsers', function() {
       await percySnapshot(this.test.fullTitle());
     });
 
+    describe('leaving organization', function() {
+      it('redirects user to / ', async function() {
```
unicornzero: trying to update this to properly test for default routes and I'm having mirage/db difficulty

PR #942, opened 2018-12-18
improve messaging when a user cannot send invites

unicornzero:
## Description
When a user cannot send invitations we want better messaging.

If the `[Invite new users]` button is disabled on the `/users` page because the user is not an admin, we want to provide a tooltip that explains this.

If a user goes to the `/users/invite` page and cannot see the form, we want to display messaging indicating that they either do not have admin permissions or that there are no seats available.


tests/integration/components/organizations/users-header-test.js
```diff
@@ -28,25 +28,25 @@ describe('Integration: UsersHeader', function() {
...
+        isInviteFormAllowed=isInviteFormAllowed
+      }}`);
+
+      const text = `You’ve used ${seatsUsed} of ${seatLimit} seats available.`;
+      expect(UsersHeader.organizationName).to.equal(organizationName);
+      expect(UsersHeader.seatCount.text).to.equal(text);
+      expect(UsersHeader.billingLink.isVisible).to.equal(true);
+      expect(UsersHeader.supportLink.isVisible).to.equal(true);
+      expect(UsersHeader.inviteButton.text).to.equal('Invite new users');
+    });
```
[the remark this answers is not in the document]
unicornzero: Ah, I'd taken the snapshot out of this one since other tests capture the same thing, but I could add it back in. :)

unicornzero:
Pending design update from @PixelJanitor

PixelJanitor:
Almost finished with my part. Didn't realize you had another PR for this. Will move my code over to here! :)

PixelJanitor:
image
Less scary permission notice has been pushed to this PR

unicornzero:
Todo: Add an acceptance test that adds a percy snapshot of the full screen showing the note.

app/templates/components/organizations/users-header.hbs
```diff
@@ -22,5 +22,17 @@
     <div class="px-4 pb-4" data-test-invite-form-wrapper>
       {{forms/organization-invite organization=organization}}
     </div>
+  {{else if isInviteFormAllowed}}
+    <div data-test-invite-permission-error class="px-4 pb-3">
+      {{#if isMember}}
+        <div class="reminder-note">
+          <strong>Note:</strong> Inviting new users requires organization admin permissions.
+        </div>
+      {{else if cannotAddSeats}}
```
maprules1000: Is there a possibility this can get into a state where`isInviteFormAllowed` is true and `isMember` and `cannotAddSeats` are both false?

If that is the case then there will be a weird spacing on the page since the div will render with padding but will have no content.

But maybe that state isn't possible to hit?

maprules1000: Also, I'm not totally sure I understand `isInviteFormAllowed`? Is that saying the invite form is visible?

unicornzero: @maprules1000 `isInviteFormAllowed` means that you're on the `/users/invite` route, but we didn't want to explicitly state that you're on a route in the variable name.

unicornzero: If `isInviteFormAllowed` is true but `isMember` and `cannotAddSeats` are false, then we catch it on line 21-23 because that means `showInviteForm` is true.

maprules1000: Ok sounds good to me @unicornzero !

PR #943, opened 2018-12-18
Update the plans shown on the billing page to v3

timhaines:
## Description
New pricing model launch


Robdel12:
Super tiny thing but the widows on the pricing table make my eye twitch a bit. 
image

Looks like (in a future PR probably) we could make the table a wee bit wider

PR #944, opened 2018-12-18
freezeMoment in Users UI tests

unicornzero:
## Description
freezeMoment in UI tests for Percy snapshots


tests/acceptance/manage-users-test.js
```diff
@@ -1,5 +1,6 @@
 import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
 import {currentRouteName} from '@ember/test-helpers';
+import freezeMoment from '../helpers/freeze-moment';
```
cadeParade: Let's use a direct path -- you can start every import from within `percy-web` with `percy-web/`. This string is set in package.json. Things are as you expect except you don't have to put `percy-web/app/...`, `percy-web/routes/etc` will do, and for tests, you can do `percy-web/tests/integration/etc`.

So this would be `percy-web/tests/helpers/freeze-moment`

PR #947, opened 2018-12-19
[WIP- style changes first] When changing subscription, show user limit error if plan change fail…

unicornzero:
…s due to user limits

## Reviewing this PR


app/services/subscriptions.js
```diff
@@ -16,12 +16,16 @@ export default Service.extend({
...
     savingPromise.then(
       () => {},
-      () => {
+      adapterErrors => {
+        const firstError = adapterErrors.errors[0];
+        const isQuotaError = firstError.status === 'quota_exceeded';
+        const quotaError = firstError.detail;
+        const generalError =
+          'A Stripe error occurred! Your card may have been declined. Please ' +
+          'try again or contact us at hello@percy.io and we will help you get set up.';
```
maprules1000: Should this be hello@percy.io, or support@percy.io?

unicornzero: I don't know. That's what was there before. @Robdel12 , what say you?

app/services/subscriptions.js
```diff
@@ -16,12 +16,16 @@ export default Service.extend({
 
     savingPromise.then(
       () => {},
-      () => {
+      adapterErrors => {
+        const firstError = adapterErrors.errors[0];
```
cadeParade: Can we make this lookup safer? If there's no `errors`property on `adapterErrors` it will explode. Ugly, but something like `adapterErrors` && `adapterErrors.errors` && `adapterErrors.errors[0]` would work.

cadeParade: Or you can use generic `get` (so import `get` from `@ember/object`) and then `get(adapterErrors, 'errors') && get(adapterErrors, 'errors')[0]` might work. `get` won't throw an error if the property you're getting is undefined.

https://www.emberjs.com/api/ember/3.4/functions/@ember%2Fobject/get

unicornzero: Thanks! I wasn't confident about this line and wasn't sure of the best approach. Thanks for giving me some options!

app/services/subscriptions.js
Robdel12: I'd say `support@` just to make it explicit. FWIW I _think_ both go to me 😆

app/services/subscriptions.js
unicornzero: Good call. `const firstError = get(adapterErrors, 'errors') && get(adapterErrors, 'errors')[0];` did the trick!

PR #948, opened 2018-12-20
Create new org, auto create demo project, redirect to demo project

cadeParade:
## Description
This PR:
- Adds a second button on the new organization screen to take a user to a demo project
- When they click said button, they are forwarded to `new-demo` route, which kicks of a request to the API for a new demo project.
- Manages the loading/disabled state of the two buttons separately.
- Puts back form functionality removed in this PR https://github.com/percy/percy-web/pull/943 (placeholder email address so users with only a github address will have the option to enter their work email address)
- Adds a `demo` badge to demo projects.
- Adds a button to top right of demo projects so users can easily navigate to make their own project.
- Redirects to second build for demo project if it exists, and the project page if at least two builds do not exist.

## Pictures
See Percy build: https://percy.io/percy/percy-web/builds/1276278

## Reviewing this PR


app/templates/components/forms/organization-new.hbs
```diff
@@ -41,10 +60,10 @@
     </div>
     {{form-fields/submit
       value="Start new project"
-      classes="percy-btn percy-btn-primary"
-      submit=(action "customSave")
-      isSaving=isSaving
-      disabled=isSubmitDisabled
+      classes="percy-btn data-test-project-submit-button"
```
maprules1000: I think we want this button to be the secondary white color. Right now both buttons are the primary purple color

cadeParade:
The last two commits are adding a feature flag to the "Start demo project" button. It is off by default, but will allow us to switch this path on easily once the API is ready. 
The last commit converts a file to use a helper I wrote for launch darkly in integration tests. The other integration test was using `setupComponentTest` instead of the newer `setupRenderingTest`. Converting one to the other required adding a bunch of `async` and `await` calls to the test, but the test is otherwise unchanged.

PR #951, opened 2019-01-03
update favicon

PixelJanitor:
Time for Pency to go:
image
 becomes
image


djones:
@PixelJanitor looks like many of those diffs were the copyright year changing 🤦‍♂️

PR #952, opened 2019-01-04
[WIP] Prev/next snapshot on fullscreen view

cadeParade:
## Description
Add a previous/next button (+ keyboard shortcuts) to fullscreen mode -- it will navigate to the previous or next snapshot in the list (same comparisonMode, browser, width). It cycles through all the snapshots with diffs, it does not venture into no-diff land (similar to keyboard nav on snapshot list)

## Reviewing this PR

## TODO: design/UI review/discussion


app/components/snapshot-viewer-full.js
```diff
@@ -3,6 +3,14 @@ import {alias} from '@ember/object/computed';
 import Component from '@ember/component';
 import filteredComparisons from 'percy-web/lib/filtered-comparisons';
 
+const KEYS = {
```
maprules1000: This makes the navigation logic *way* easier to read

app/routes/organization/project/builds/build/snapshot.js
```diff
@@ -124,4 +173,34 @@ export default Route.extend(ResetScrollMixin, {
...
+      return;
+    }
+
+    const currentSnapshotIndex = snapshots.mapBy('id').indexOf(snapshotId);
+    let nextSnapshotIndex = isNext ? currentSnapshotIndex + 1 : currentSnapshotIndex - 1;
+
+    // A user has navigated past the last snapshot
+    if (nextSnapshotIndex > snapshots.length - 1 && isNext) {
+      nextSnapshotIndex = 0;
+      this.get('flashMessages').info("You've been looped to the beginning");
```
maprules1000: Maybe "This is the beginning" or "You're now at the beginning". Same with the end, but I don't have a very strong opinion about it.

app/controllers/organization/project/builds/build.js
```diff
@@ -54,4 +54,12 @@ export default Controller.extend({
       isSnapshotsLoading: false,
     });
   },
+
+  _unchangedSnapshots: [],
```
maprules1000: I'd move this to right under `orderedBrowserSnapshots` for file readability.

timhaines:
Had a little play with this. Awesome to see it cycles through snapshots with no changes when they're expanded in the build too.  👍 

Are we consciously deciding to show the buttons and keep the keyboard active when there's only 1 change?

cadeParade:
@timhaines It is sort of conscious...It is because it is a bit technically tricky for the fullscreen snapshot to actually know whether there are prev/next snapshots available. It is definitely possible but I made the call to not invest the time right now. Happy to approach it in a followup if you think the flash message is not sufficient.

PR #959, opened 2019-01-14
Demo Project Error Polling

maprules1000:
## Description
Sometimes there may not be a demo project readily available for assignment. In those cases users will see a very gentle error message telling them a project is being created. The "error page" polls the server in the background and will proceed with a correct redirect when a project has been assigned.

## Pictures
https://percy.io/percy/percy-web/builds/1324122/view/86471951/1280?browser=chrome&mode=diff

## Reviewing this PR


app/routes/organizations/organization/projects/new-demo.js
```diff
@@ -1,16 +1,48 @@
+import Ember from 'ember';
 import Route from '@ember/routing/route';
 import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import {task, timeout} from 'ember-concurrency';
+
+const MAX_REFRESH_REQUESTS = 20;
+const WAIT_TIME = 10000;
```
djones: this should be probably something like 10 mins. Roughly the time it would take to completely create a demo project with 3 builds.

cadeParade:
I am snooping, but here's two examples of testing polling behavior:
https://github.com/percy/percy-web/blob/master/tests/acceptance/project-test.js#L103
https://github.com/percy/percy-web/blob/master/tests/acceptance/build-test.js#L91
:D

app/routes/organizations/organization/projects/new-demo.js
```diff
@@ -24,4 +68,19 @@ export default Route.extend(AuthenticatedRouteMixin, {
       this.transitionTo('organization.project', organization.get('slug'), demoProject.get('slug'));
     }
   },
+
+  saveDemo(demo) {
+    return demo.save().then(
```
cadeParade: We are moving more towards `async/await` or ember concurrency -- is there a reason you've used `then` here instead of either of those?

app/routes/organizations/organization/projects/new-demo.js
```diff
@@ -1,16 +1,60 @@
...
+      } else {
+        // if there is no error then clear errors and route to demo project
+        this.controller.set('hasError', false);
+        this.fetchDemoProject(demoProject);
+        return;
+      }
+    }
+  }).drop(),
+
+  async fetchDemoProject(demoProject) {
```
cadeParade: `fetchDemoProject` is actually not fetching the demo project anymore, it is getting the builds, so we should name it something like `fetchDemoBuilds`.

cadeParade: It also transitions so `fetchDemoBuildsAndTransition`?

app/templates/organizations/organization/projects/new-demo.hbs
```diff
@@ -1,3 +1,20 @@
-<div class="container">
-  {{loading-page}}
-</div>
+{{#if hasError}}
+  <div class="container-full container-vh bg-white-pattern flex items-center justify-center">
+   <!-- recommendation : the integrations icon with a circle or the gear icon with a circle  -->
```
cadeParade: Is this comment still needed?

tests/acceptance/organization-test.js
```diff
@@ -96,7 +97,43 @@ describe('Acceptance: Organization', function() {
     });
 
     describe('creating a new org with demo project', function() {
-      it('redirects to demo project when project has no builds', async function() {
+      it('shows a custom loading screen when there is no project available', async function() {
```
cadeParade: This seems like it's testing for an error screen rather than a loading screen, so maybe the title should be updated?

app/routes/organizations/organization/projects/new-demo.js
```diff
@@ -1,16 +1,60 @@
...
+
+    for (_numPollRequests; _numPollRequests < MAX_REFRESH_REQUESTS; _numPollRequests) {
+      const saveRetry = yield this.saveDemo(demoProject);
+      const error = saveRetry.error;
+
+      if (error) {
+        if (Ember.testing) {
+          return;
+        } else {
+          yield timeout(WAIT_TIME * MILLISECONDS_IN_MIN);
```
cadeParade: So it only polls every 10 minutes?
Is that useful? Am I reading it wrong?

app/routes/organizations/organization/projects/new-demo.js
cadeParade: Or maybe just `transitionToProject` because getting the builds is a prerequisite for transitioning.
In any case the naming of this method could be improved.

app/routes/organizations/organization/projects/new-demo.js
maprules1000: Yeah, it'll only poll every 10 min. That's about the estimated time to generate a demo project in the api. We might be able to tweak these numbers after release too.

app/routes/organizations/organization/projects/new-demo.js
maprules1000: 👍

app/templates/organizations/organization/projects/new-demo.hbs
maprules1000: Not really

app/routes/organizations/organization/projects/new-demo.js
maprules1000: I think it might be possible. They could just go to another tab and do other things, and then when they come back and **bam** demo project.

maprules1000: This is also supposed to be a very edge case error state

app/routes/organizations/organization/projects/new-demo.js
```diff
@@ -1,16 +1,60 @@
...
+      } else {
+        // if there is no error then clear errors and route to demo project
+        this.controller.set('hasError', false);
+        this.fetchDemoProject(demoProject);
+        return;
+      }
+    }
+  }).drop(),
+
+  async transitionToDemo(demoProject) {
```
cadeParade: The places where this is called will also have to be updated.

maprules1000: 🤦‍♂️

PR #962, opened 2019-01-15
Light trial bar everywhere now

PixelJanitor:
image


PixelJanitor:
There is, I just didn't catch that the link changed. Saw that they trial bar was there, but just missed it from all the red red red red red red red red red

PR #964, opened 2019-01-17
[WIP] Free plan usage banner

unicornzero:
**This PR is dependent upon an [API change](https://github.com/percy/percy-api/pull/778)**
### Description
As a Percy user on a free plan, I want to be aware of how much free usage I have left this month.

### Pictures
screen shot 2019-01-17 at 2 51 54 pm
screen shot 2019-01-17 at 2 52 02 pm
screen shot 2019-01-17 at 2 52 30 pm
screen shot 2019-01-17 at 2 52 51 pm
screen shot 2019-01-17 at 2 54 50 pm

### Reviewing this PR
_**Is Covered by Automated Tests?: YES/NO**_ - Not yet, working on that next.


unicornzero:
@cadeParade realized there's an edge case for both this banner and the trail banner causing the page not to load if a non-member views a public project. We want to get a fix for that before we ship this.

app/models/subscription.js
```diff
@@ -32,4 +32,12 @@ export default DS.Model.extend({
   trialDaysRemaining: computed('trialEnd', function() {
     return Math.round(moment(this.get('trialEnd')).diff(moment(), 'days', true));
   }),
+  currentUsageRatio: DS.attr(),
+  currentUsagePercentage: computed('currentUsageRatio', function() {
+    const percentage = parseFloat(this.get('currentUsageRatio')) * 100;
+    return percentage < 1 ? Math.ceil(percentage) : Math.floor(percentage);
+  }),
+  hasIncludedSnapshotsRemaining: computed('currentUsageRatio', function() {
```
cadeParade: There is a computed macro `lt` for less than. So this could be 
`hasIncludedSnapshotsRemaining: lt('currentUsageRatio', 1),`

app/styles/components/_notice_bar.scss
```diff
@@ -1,4 +1,4 @@
-.trial-bar {
+.notice-bar {
```
cadeParade: Text color and background color can be set with tailwind classes. 
This could be set on some div in `notice-bar` with 
`class="text-gray-700 bg-gray-300"`
The background color in particular does not exactly match `--very-light-gray`* so I would choose whatever grey color that looks good.
This file and these css variables are very old, so I've approximated these colors. I think that it's preferable to use these utility classes rather than the css variables, but @PixelJanitor can weigh in for sure).

\* The way you can tell they are not the same color is because -- if you grep for `--very-light-gray:` you'll see a hex code assigned `#e9e9e9`. Then if you check `dev.percy.local:4200/tailwind` you can see the hex codes for the "100" series colors. There's no `e9e9e9` in that list, so I know that it can't be exactly matched.

tests/acceptance/organization-test.js
```diff
@@ -41,7 +41,8 @@ describe('Acceptance: Organization', function() {
       await percySnapshot(this.test);
     });
 
-    it('can create new organization and update org switcher when creating second org', async function() { // eslint-disable-line
+    it('can create new org and update org switcher when creating second org', async function() {
+      // eslint-disable-line
```
cadeParade: This is probably an automatically generated diff by prettier.
If it moved `// eslint-disabled-line` to the next line automatically, the preceding line is either not too long anymore, or the linter will fail. So anyway, either this comment needs to be on the previous line still or the comment needs to be deleted.

tests/acceptance/notice-bar-test.js
```diff
@@ -0,0 +1,176 @@
...
+        let project = server.create('project', {organization});
+        let build = server.create('build', {project});
+
+        urlParams = {
+          orgSlug: organization.slug,
+          projectSlug: project.slug,
+          buildId: build.id,
+        };
+
+        this.organization = organization;
```
cadeParade: We prefer to assign a scoped variable (such as line 13 -- `let urlParams` rather than assigning something to the test context. There are still some references in the code base of this pattern, but they are the old way. 
The reason for this is because (at least at some point in Ember's past) assigning properties to the test context caused memory leaks and therefore flaky tests.

tests/acceptance/notice-bar-test.js
```diff
@@ -0,0 +1,176 @@
...
+      });
+
+      describe('in top header', function() {
+        it('appears on a project page', async function() {
```
cadeParade: Let's make this test `appears on project page and links to billing`.
Acceptance tests can be a little beefier than integration or unit tests because each `it` block requires the test runner to tear down and start up a new application instance. So doing a few more commands in an acceptance test is better for test performance in the long run.
ex:
```
        it('appears on a project page and links to billing', async function() {
          await ProjectPage.visitOrg({orgSlug: this.organization.slug});

          expect(NoticeBar.message.isVisible).to.equal(true);
          expect(NoticeBar.billingLink.isVisible).to.equal(true);

          await NoticeBar.billingLink.click();
          expect(currentRouteName()).to.equal('organizations.organization.billing');
        });
```

cadeParade: This file could be consolidated a bit by defining a helper method (it must be defined in the test context, so after line 9):

```
  async function checkNoticeBar(context) {
    expect(NoticeBar.message.isVisible).to.equal(true);
    expect(NoticeBar.billingLink.isVisible).to.equal(true);

    await percySnapshot(context.test);
    await NoticeBar.billingLink.click();

    expect(currentRouteName()).to.equal('organizations.organization.billing');
  }
```
then in a test:
```
 it('appears on a project page', async function() {
          await ProjectPage.visitOrg({orgSlug: this.organization.slug});
          await checkNoticeBar(this);
  })
```

mirage/serializers/organization.js
```diff
@@ -1,13 +1,21 @@
...
 export default JSONAPISerializer.extend({
-  include: [
-    'subscription',
-    'versionControlIntegrations',
-    'subscription.currentUsageStats',
-    'githubIntegrationRequest',
-    'githubIntegrationRequest.createdBy',
-  ],
+  include: function() {
+    const users = this.registry.schema.db.users;
```
cadeParade: There is one more case where the includes would be nothing -- it is when there is a user is logged into their own percy account but is viewing a project that belongs to an organization they don't have access to.
This fix is probably good enough for now, but now you know :/

app/templates/components/notice-bar.hbs
```diff
@@ -0,0 +1,7 @@
+{{#if (await isUserMember)}}
+  {{#if organization.subscription.isFree}}
+    {{free-usage-bar organization=organization}}
```
cadeParade: Since the two bars are purely markup now and they are fairly short, it doesn't seem like there's much of an advantage to have them in separate components rather than putting the HTML for both of them inline in `notice-bar`. I don't feel that strongly though.

app/models/subscription.js
unicornzero: Thanks! Forgot to look for computed macros. :)

mirage/serializers/organization.js
unicornzero: Right. Noted. And agreed. :/

app/templates/components/notice-bar.hbs
unicornzero: That sounds good to me. I am not crazy about how it is right now and cutting the number of files sounds good.

tests/acceptance/notice-bar-test.js
unicornzero: Oh, that is very helpful insight! Thanks!

tests/acceptance/notice-bar-test.js
unicornzero: Thanks; I really appreciate how you spell out new-to-me things. Super helpful!

PR #967, opened 2019-01-24
Reposition org switcher

cadeParade:
## Description
To improve UX for users, we want to improve the navigation between organizations and organization settings. 

To do this, we've moved the nav-switcher component from the top left of the screen to the top right inside the user menu.

We've also added a link to `fixed-top-header` called `Dashboard` which will redirect you to a recent project or organization from many places in the app which was previously not possible.

## Reviewing this PR
_**Is Covered by Automated Tests?: YES?**_


tests/integration/components/fixed-top-header-test.js
```diff
@@ -0,0 +1,58 @@
+/* jshint expr:true */
+/* eslint-disable no-unused-expressions */
```
maprules1000: Why is this disabled for this file?

cadeParade: The old copy-pasta

PR #973, opened 2019-01-28
Remove new-org-with-no-projects screen

cadeParade:
## Description
This PR removes the screen that looks like this: 
slack_-_percy

And instead forwards you to the new project page.

The code that does this is this line: https://github.com/percy/percy-web/pull/973/files#diff-50cae791df7cab4bad4779409b327d97R26
This line also sets up the next PR to be able to centralize project picking logic into one place: https://github.com/percy/percy-web/pull/967/files#diff-e26bb49418f164214024ca15b7af545aR39

The rest of the updates are updating tests, and tidying up things related to the tests a bit.


app/templates/components/form-fields/submit.hbs
```diff
@@ -1,7 +1,7 @@
 <button
   class="{{classes}} relative {{if isSaving 'is-loading'}} {{if isSaveSuccessful 'is-successful'}} {{if (eq isSaveSuccessful false) 'is-failed'}}"
   disabled={{disabled}}
-  onclick={{action "perform"}}
+  {{action "perform"}}
```
cadeParade: Don't use browser events, keep all our events in the ember run loop.

app/templates/components/forms/project-new.hbs
```diff
@@ -3,12 +3,13 @@
...
-        validateProperty=(action "validateProperty" changeset "name")
-        autofocus=true
-        placeholder="widgets-web-app"
+      property="name"
+      title="Project name"
+      changeset=changeset
+      validateProperty=(action "validateProperty" changeset "name")
+      autofocus=true
+      placeholder="widgets-web-app"
+      testLabel="project-name-input"
```
cadeParade: These diffs are tabs + the the addition of `testLabel` line

mirage/config.js
```diff
@@ -121,17 +121,14 @@ export default function() {
...
-      // The create-demo request from the app includes almost no information --
-      // the server provides it. So we have to make up all the information we want to
-      // be returned on the project here.
-      const demoProjects = schema.projects.all().models.filterBy('isDemo', true);
-      attrs.name = `My cool demo project ${demoProjects.length}`;
-      attrs.slug = `my-cool-demo-project-${demoProjects.length}`;
-      attrs.fullSlug = `${organization.slug}/${attrs.slug}`;
-      attrs.isEnabled = true;
-    }
+
```
cadeParade: This logic has been generalized, because it's important for the router to have a project and full slug on every project. It turns out that there was not a test correctly creating a project that was not a demo so this was never a problem before. But I updated the `create project` test to route to the new project page, so now we need the slugs in all the created projects in tests. 
This makes more sense anyway IMO.

PR #974, opened 2019-01-30
Do not show dynamic status dot for any except active project

cadeParade:
## Description
The dots in the project sidebar are sometimes problematic for performance reasons. 
We decided to show the most recent build status in the dot for only the active project.


maprules1000:
The code looks good, and it performs as expected when running agains prod, but the diffs don't look like they should.

Maybe some data also needs updating?

maprules1000:
Also for some reason this breaks on the demo org

cadeParade:
Chatted in slack with @maprules1000 -- the reason the demo org was misbehaving was related to his local setup. The Percy diffs on a second look are correct.

PR #977, opened 2019-01-31
Put projects settings under project header

cadeParade:
## Description
To make navigation between settings and builds easier, we are adding a new button set in the project header where a user can navigate between a project's builds and settings.
This required rearranging several routes, but very little functionally has changed.
The routes are now:

```
Organization
    project
        index (displays builds -- fixed-top-header, project-sidebar, project-container)
        settings (fixed-top-header, project-sidebar, project-header)
            index (displays settings forms)
```

The reason the project sidebar and project header are not in the `organizations.project` route, is because the individual build and fullscreen snapshot view are _also_ under `organizations.project` and we don't want the sidebar to display for those views. So the views still have to be pieced together with components rather than taking advantage of Ember's `outlet` functionality.

See: this hilarity: https://percy.io/percy/percy-web/builds/1376416/view/90275183/1280?browser=chrome&mode=diff


app/routes/organization/project/settings/index.js
```diff
@@ -0,0 +1,89 @@
+import Route from '@ember/routing/route';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import {hash} from 'rsvp';
+import {inject as service} from '@ember/service';
+import utils from 'percy-web/lib/utils';
+
+export default Route.extend(AuthenticatedRouteMixin, {
```
cadeParade: This file is most of what `app/routes/organization/project/settings.js` used to be. Nothing has changed except the model hook returns less stuff.

app/router.js
```diff
@@ -102,9 +102,8 @@ Router.map(function() {
     // Don't add anything else in this top-level namespace, we want to allow users to own the whole
     // projects namespace. Org-level settings and such should go in the above "organizations" route.
     this.route('project', {path: '/:project_id'}, function() {
-      this.route('settings');
-      this.route('integrations', function() {
-        this.route('webhooks', {}, function() {
+      this.route('settings', {}, function() {
```
cadeParade: There is/was nothing at the integrations nor webhooks route, so I removed an intermediate as it seems useless. I didn't remove `integrations` route because I can see a future where we have different kinds of integrations.

cadeParade: I will undo this.

tests/pages/components/project-sidebar.js
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const ProjectSidebar = {
+  projectLinks: collection({
+    itemScope: SELECTORS.SIDEBAR_PROJECT_ITEM,
+    item: {
+      projectName: text(),
+    },
+  }),
+  // toggleProjectSidebar: clickable(SELECTORS.TOGGLE_PROJECT_SIDEBAR),
```
cadeParade: get rid of this?

cadeParade:
Why does project name disappear when switching from builds to project settings?

PR #997, opened 2019-02-06
Demo Updates

maprules1000:
## Description
This PR incorporates the feedback discussed in these google docs.

[Doc 1](https://docs.google.com/document/d/1CKVhEhAsJPfICu-XkwBeqYqSXYNjhP8CoC5JLdmNXQA/edit#)  [Doc 2](https://docs.google.com/document/d/1A0j71u9JWlHCxRv7TcjEB_Jfnfk7EmGJsZUnY8gXqsI/edit)


app/templates/components/comparison-viewer.hbs
```diff
@@ -15,7 +15,7 @@
           anchorPlacement='left-corner'
           key="three"
           title="Visual Diff"
-          message="These are the that have changed due to changes in code. We call this a visual diff. To toggle between the visual diff and the new screenshot that makes up that diff you can click the screenshot on the right or just type “d”."
+          message="These red pixels highlight the visual changes underneath—we call this a “visual diff. To toggle the visual diff on or off click it (or press “d”)."
```
cadeParade: Is this an em-dash? It looks weird in GitHub but if you've checked it then 👌

app/templates/components/demo-tooltip.hbs
```diff
@@ -3,7 +3,7 @@
     {{#attach-popover
       class='tip-container'
       showOn='click'
-      hideOn='click'
+      hideOn='clickout'
```
cadeParade: Does this mean that it doesn't hide when you click inside it? Is that desirable?

maprules1000: This means it will hide when you click outside of the tooltip, including clicking on another tooltip. It won't hide if you click inside it of if you click on the anchor again, but it will hide if the 'got it' or 'hide all' button is clicked.

PR #1002, opened 2019-02-07
Pricing slider

cadeParade:
## Description
See [document](https://docs.google.com/document/d/1luqJ_h3r1mxRjYVX9SuS6lxmDYDxl8EZcZ7zM-zBGYA/edit#)  for information.

I've also removed some of the unneeded contentful stuff surrounding the pricing cards (although the model still needs to remain -- there will be a followup PR that removes the model after the pricing cards have been removed from contentful).

@PixelJanitor has also made a heroic effort to style the slider across browsers, I mean, this stuff is complicated. 🎉


app/tailwind/components/range-slider.css
```diff
@@ -0,0 +1,53 @@
+/* stylelint-disable selector-no-qualifying-type */
+/* stylelint-disable property-no-vendor-prefix */
+
+input[type=range] {
+  width: 100%;
+  -webkit-appearance: none;
+}
+
+input[type=range]:focus {
+  outline: none;
```
Robdel12: 😢 Can we apply some other method of showing this control has focus? Doesn't have to be an outline but it'd be nice to know where my keyboard focus is while I tab through the page.

PR #1012, opened 2019-02-13
Use LD development environment to control local feature flags.

cadeParade:
## Description
It is not ideal to manage feature flags in two places.
There are two problems: 
- It is hard to tell what the state of your environment should be since there are two places to manage flag state
- Some folks are using the production API all the time and only running the frontend locally, so the flags set in `config/environment` do not apply to their environment.

This PR takes advantage of the separate flag environment `Development` so we can have one place to manage feature flags.


config/environment.js
```diff
@@ -27,9 +27,7 @@ module.exports = function(environment) {
       preventDuplicates: true,
     },
 
-    launchDarkly: {
-      clientSideId: '5b453a435b18c32c7440a5bd',
-    },
+    launchDarkly: {},
```
maprules1000: Could this be removed?

PR #1013, opened 2019-02-14
Add snapshot grouping logic and tests

cadeParade:
Logic for dividing snapshots into groups by fingerprint. 
The main complication about this process was that the snapshots are already sorted via our snapshot sort algorithm. I did not want to sort all the possible arrays again by that process.
So this logic takes the pre-sorted snapshots and divides them into groups, including the cases of having a null or undefined value as the fingerprint as well as having only one snapshot for a fingerprint.


maprules1000:
I fixed the test contentful model that I think was breaking your tests.

PR #1016, opened 2019-02-15
Marketing Video

maprules1000:
Adds the marketing video contentful model and wires it all up (with tests).


app/routes/index.js
```diff
@@ -6,7 +6,7 @@ export default Route.extend({
 
   model() {
     return this.get('store').queryRecord('marketing-page', {
-      'fields.pageName': 'Home',
+      'fields.pageName': 'Home-MAP-Video-Block',
```
maprules1000: This will be reverted before being shipped. It's here for development.

maprules1000:
The tests will fail until this is ready for merge. It's a contentful thing.

app/components/marketing/video-block.js
```diff
@@ -0,0 +1,11 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+  isVideoModalVisible: false,
+
+  actions: {
+    toggleVideoModal() {
```
cadeParade: If you would like to reduce lines of code, and delete this whole file, you can use this fancy helper in the template to toggle properties: `{{action (toggle "isVideoModalVisible" this)}}`.

In your template below it would look like:
`<div class="relative rounded overflow-hidden shadow-y-lg-purple-600" {{action (toggle "isVideoModalVisible" this)}} data-test-video-block-image>`

But if you like the JS toggle better, that is fine, I don't mind either way. Just wanted to let you know about the possibility :)

It comes from here: https://github.com/DockYard/ember-composable-helpers#toggle

app/models/video-block.js
```diff
@@ -0,0 +1,29 @@
...
+  subheader: attr(),
+
+  videoEmbedUrl: attr(),
+  classes: attr(),
+
+  imageUrl: readOnly('mainImage.file.url'),
+  imageDescription: readOnly('mainImage.description'),
+  isImagePresent: notEmpty('imageUrl'),
+
+  isImageCentered: equal('imagePosition', 'Center'),
```
cadeParade: Do we need all these image properties? Will we have more than one video block where it might be aligned differently?

maprules1000: We could. I just kept them in to be as similar to and flexible as the content block as possible. It also accurately reflects the model in Contentful.

app/components/marketing/video-block.js
maprules1000: Oh neat! I'll do that then :)

app/components/marketing/video-block.js
```diff
@@ -2,10 +2,4 @@ import Component from '@ember/component';
 
 export default Component.extend({
   isVideoModalVisible: false,
```
cadeParade: You can remove this whole file if you use the template helpers. If we don't automatically set a value for `isModalVisible` it will evaluate to false and toggle correctly.

PR #1024, opened 2019-02-20
Snapshot grouping UI V1

cadeParade:
## Description
A very early version of snapshot grouping UI.

This UI uses the snapshot grouping method merged previously (#1013 ) and introduces the idea of a `snapshot-block`.

Previously, `snapshot-list` was a list of single snapshots -- a homogenous list. Now, we are introducing `snapshot-groups` intermingled with `snapshot-list`. Snapshot-list is now heterogeneous.  But the two types of snapshot items need to be displayed  in almost exactly the same way and have many of the same actions available (approve, expand full screen, switch width, etc). 
So I've made a check in the template to see whether the "snapshot-block" is an array (aka a group) or not and decided which component to display based on that.

This approach has also been carried through to the tests -- the `snapshot-list` page-object now also has a concept of snapshot-block and can decide whether it should ask a snapshot object or group object to carry out the action requested.

## Reviewing this PR

_**Is Covered by Automated Tests?: YES-ish**_
There are fairly thorough tests for snapshot-list ordering and the group approval button. Other actions (such as expand to full screen, switch widths, collapse/expand group) are not tested as they may be updated soon based on results from the user testing.

TODO in followup PRs:
- Expand/collapse behavior when a group is approved
- Keyboard nav of snapshot list with groups
- Finalization of designs (what actions will be available on the header, location of buttons, etc)


tests/acceptance/build-test.js
```diff
@@ -684,11 +709,12 @@ describe('Acceptance: Demo Project Build', function() {
     await BuildPage.visitBuild(urlParams);
 
     const tooltipElement = await findAll('.ember-attacher').firstObject;
-
-    expect(BuildPage.demoTooltips.length).to.be.at.least(2);
+    expect(BuildPage.demoTooltips.length).to.equal(5);
```
cadeParade: This test was passing, even though I had broken the index logic. I added these lines so the test would fail if the snapshot viewer component was not receiving an index value.

config/environment.js
```diff
@@ -175,6 +175,10 @@ module.exports = function(environment) {
     ENV.moment = {
       allowEmpty: true,
     };
+    ENV.launchDarkly.local = true;
```
cadeParade: TIL even though there is a default arg as false in the launch darkly addon , it throws a warning if you fall back to the default value which is _extremely noisy_. So we can set the default value of feature flags in tests in this hash.

app/components/group-approval-button.js
```diff
@@ -0,0 +1,25 @@
...
+
+  actions: {
+    async approveGroup() {
+      set(this, 'isLoading', true);
+      await get(this, 'createReview')(get(this, 'approvableSnapshots'));
+      // The time between when we get the response back from the server and the time
+      // it takes Ember to process and render the returned data is actually quite long (~1s)
+      // on production. So setting these properties here disguises that transition time and
+      // prevents jank in the Approve/Approved button state
+      this.set('isGroupApproved', true);
```
maprules1000: Maybe make this `set(this,...` to be consistent

app/components/snapshot-group.js
```diff
@@ -0,0 +1,41 @@
...
+    return this.get('snapshots').every(snapshot => {
+      return snapshot.get('isApproved');
+    });
+  }),
+
+  filteredComparisons: computed('coverSnapshot', 'activeBrowser', 'groupSelectedWidth', function() {
+    return filteredComparisons.create({
+      snapshot: this.get('coverSnapshot'),
+      activeBrowser: this.get('activeBrowser'),
+      snapshotSelectedWidth: this.get('userSelectedWidth'),
```
maprules1000: Not blocking at all, but the other js files use `get()` and `set()` but this one is using `this.get()` and `this.set()`

tests/acceptance/build-test.js
maprules1000: That's cool, I didn't want to hard code counts into here in case the number of tooltips changed, but it looks like it'll hold steady for a bit so 👍

app/components/group-approval-button.js
cadeParade: yep, good call. Still getting used to this syntax.

PR #1025, opened 2019-02-21
Video Updates

maprules1000:
## Description
Moves the header and subheader to the top of the video content block and darkens the overlay a little bit.


tests/integration/components/marketing/video-block-test.js
```diff
@@ -12,8 +12,8 @@ describe('Integration: Marketing/VideoBlock', function() {
 
   const videoBlock = {
     isImagePresent: true,
-    isImageRightAligned: true,
-    isImageCentered: false,
+    isImageRightAligned: false,
```
cadeParade: Is it possible for the image to be right or left aligned now? My (very tenuous) understanding from the slack channel is that it's always centered now. If so, can we remove the un-needed properties?

maprules1000: It was not possible when the header was above the image and the subheader was below, but now both the header and subheader are above the image, so left and right alignment works again 🎉

maprules1000: Do you think it makes sense to add some snapshots to check for that? I'm on the fence about it since this won't be a very common component.

cadeParade: nah, I think it's fine.

PR #1026, opened 2019-02-21
Expand/collapse behavior of snapshot groups + keyboard nav

cadeParade:
## Description
This PR adds (a) keyboard nav, (b) auto expand/collapse behavior to a snapshot list that includes groups.

Since the keyboard nav and auto expand/collapse behavior should be identical between snapshot viewer components and snapshot group components, I've made a parent object that contains the logic for keyboard nav and expand/collapse behavior. Now, both `snapshot-viewer` and `snapshot-group` component inherit from this parent class and share this logic. 

TODO in followup PRS:
- Keyboard nav through groups with all snapshots visible
- Hide/show all diffs button and keyboard shortcut
- Lazy loading for groups
- Design updates


app/components/snapshot-group.js
```diff
@@ -1,41 +1,30 @@
...
-    });
-  }),
-
-  actions: {
-    updateSelectedWidth(value) {
-      set(this, 'userSelectedWidth', value);
-    },
+  click() {
+    this.send('expandBlock');
+    this._super(...arguments);
```
maprules1000: Hmm, this is really clunky. Would it be possible to use an action here instead of `click()`? I'm not sure what the right pattern is here, but this doesn't feel like it.

You might be able to call the `expandBlock` action directly in the component and then the `click()` event in  `SnapshotListItem` should fire?

I think this might be a goo opportunity to just try to move away from `click()` altogether. It's basically an old skool action that hasn't been deprecated imo.

app/components/snapshot-list-item.js
```diff
@@ -0,0 +1,119 @@
...
+        setProperties(this, {
+          isExpanded: true,
+          isUnchangedSnapshotExpanded: true,
+        });
+      }
+      // Wait a tick for the above properties to be set on unchanged snapshots, so the snapshot will
+      // become fully expanded before scrolling. If we didn't wait for this, the component would
+      // scroll to a height based on the closed snapshot viewer height rather than the opened one.
+      next(() => {
+        window.scrollTo(0, this.$().get(0).offsetTop - 48); // 48px - snapshot viewer header height
```
maprules1000: Could you move this number to a const before `Component.extend`?

app/components/snapshot-list.js
```diff
@@ -85,9 +85,15 @@ export default Component.extend({
       function(e) {
         if (this.get('isKeyboardNavEnabled')) {
           if (e.keyCode === KEYS.DOWN_ARROW) {
-            this.set('activeSnapshotId', this._calculateNewActiveSnapshotId({isNext: true}));
+            this.set(
```
maprules1000: Hmm. So, I think if you're going to use the new `get/set` syntax in a file you might as well replace it everywhere in a file so there isn't mixes usage.

app/components/snapshot-viewer.js
```diff
@@ -1,119 +1,16 @@
...
-import filteredComparisons, {hasDiffForBrowser} from 'percy-web/lib/filtered-comparisons';
-import InViewportMixin from 'ember-in-viewport';
+import {or, readOnly} from '@ember/object/computed';
+import SnapshotListItem from 'percy-web/components/snapshot-list-item';
 
-export default Component.extend(InViewportMixin, {
-  // required params
-  allDiffsShown: null,
-  build: null,
+export default SnapshotListItem.extend({
```
maprules1000: Wow this cleans up this file realllll nice. 👌

app/components/snapshot-list-item.js
cadeParade: Yes.

app/components/snapshot-list.js
cadeParade: Yes.

app/components/snapshot-group.js
cadeParade: Thanks for calling this out. This was a holdover from an earlier revision. I've removed it entirely from this component, it is now depending on the click action in the parent component (`snapshot-list-item`).

I do not really agree that `click` action should not be used. The `click` action and other component-level actions are still in the ember guides, and allow you to define actions whenever a user interacts with anything in your component rather than choosing which element to attach the action to. https://guides.emberjs.com/release/components/handling-events/

In this case, we want a click anywhere on the component to make that snapshot block the active block, so putting that in the `click` action makes sense.

cadeParade:
@maprules1000 updated!

maprules1000:
Hmm, one thing I just noticed is that the diffs for the Billing section Integration test are not freezing the dates anymore. I assume that worked previously?

cadeParade:
The flaky date is unrelated to this change and fixed here: https://github.com/percy/percy-web/pull/1036

PR #1027, opened 2019-02-22
Add keyboard nav event for snapshot list

cadeParade:
Does anyone use this feature? Let's find out.


app/components/snapshot-list.js
```diff
@@ -48,6 +49,11 @@ export default Component.extend({
     $(document).bind(
       'keydown.snapshots',
       function(e) {
+        this.get('analytics').track(
+          'Snapshot List Keyboard Navigated',
```
fotinakis: How about 'Snapshot List Navigated' and have a property like `type: keyboard`? We could add other things there over time. Also please add this to the Event Taxonomy spreadsheet when we decide.

cadeParade:
@fotinakis updated

app/components/snapshot-list.js
```diff
@@ -58,6 +59,15 @@ export default Component.extend({
             this.get('toggleAllDiffs')({trackSource: 'keypress'});
           }
         }
+        this.get('analytics').track(
+          'Snapshot List Navigated',
+          this.get('build.project.orgainzation'),
```
fotinakis: `orgainzation` --> `organization`

PR #1028, opened 2019-02-22
Update demo polling when no demos are available

maprules1000:
## Description
Some users got into a state where they would get much more than 1 demo project when requesting just one. This had to do with how polling was setup. In addition to percy/percy-api#845, which limited orgs to only one demo project on the backend, this updates the "polling" and save process to only make a request for one demo project.


app/routes/organizations/organization/projects/new-demo.js
```diff
@@ -2,55 +2,70 @@ import Ember from 'ember';
...
-        }
-      } else {
-        // if there is no error then clear errors and route to demo project
-        this.controller.set('hasError', false);
-        this.transitionToDemo(demoProject);
-        return;
-      }
+  // only for organization acceptance testing
+  _testingCanRefreshOnce: true,
+  _testingRefreshOnce() {
```
maprules1000: I'm not thrilled about this, but it was the only way to assure this was only called once in order for all the tests to work.

app/routes/organizations/organization/projects/new-demo.js
```diff
@@ -2,55 +2,70 @@ import Ember from 'ember';
...
   },
 
   // save in afterModel so we can render a loading template while we wait
-  afterModel(demoProject) {
-    this.saveProcess(demoProject);
+  afterModel(model) {
+    const organization = model.organization;
+    const projects = model.projects;
+
+    if (projects.length === 0) {
```
maprules1000: This covers the different states of an org: it has no projects, which means it is correctly requesting a demo, it has an unsaved project which is the initial demo request, it has only one project which is a demo, or it has more than one project which means it just needs to go to the project listing screen.

cadeParade: While it is likely that when an org has no projects, it is requesting a demo, it is _possible_ to get into a state where your org has a project that is _not_ a demo but now you're requesting a demo. In that case, this logic would be incorrect, right?

cadeParade: Ok we talked and apparently the demo button does not show up if an organization already has a project.

PR #1037, opened 2019-02-27
Freeze loader animation

faun:
## Description
Freeze loading animation in Percy to stabilize loading spinner across builds

[Clubhouse Card](https://app.clubhouse.io/percy/story/2658/freeze-svg-loader-animation)


app/styles/components/_svg.scss
```diff
@@ -15,7 +15,7 @@
 }
 
 @media only percy {
-  .loader path {
```
cadeParade: What does this do to the class `loader` which is still in `loader-icon` file?
If we do not need `loader` class anymore, let's remove it from the svg file?

faun: Fixed in 2bc2d6e

PR #1040, opened 2019-02-28
Toggle all diffs button works with snapshot groups

cadeParade:
This PR allows using the "Toggle all diffs" button to work with snapshot groups.
To do this, the action for toggling a single diff was moved out of `comparison-viewer` and up to the `snapshot-viewer`/`snapshot-group` component. It was moved because I wanted the group to remember what diff overlay state it had even when showing and hiding all snapshots. It would not "remember" what diff overlay state it had without moving the action and state out of  `comparison-viewer` because as you show and hide all snapshots in a group, the comparison viewer that should "remember" the state un-renders and re-renders, destroying all state that it used to have. So I've moved the action to the snapshot-list-item to remedy this. 

Additionally, since snapshot viewers are now contained within snapshot groups, the snapshot viewer diff overlay and snapshot group diff overlay state need to be able to be controlled separately (i.e. not in the shared `snapshot-list-item` component). So I've introduced `isSnapshotShowingDiffOverlay` and `isGroupShowingDiffOverlay` properties.

## Reviewing this PR

TODO in followup PRs
* Keyboard nav through groups with all snapshots visible
* Lazy loading for groups
* Design updates


app/components/snapshot-list-item.js
```diff
@@ -108,6 +108,19 @@ export default Component.extend(InViewportMixin, {
...
   }),
 
+  trackToggleOverlay() {
+    const build = get(this, 'build');
+    const organization = get(build, 'project.organization');
+    const eventProperties = {
+      project_id: get(build, 'project.id'),
+      project_slug: get(build, 'project.slug'),
+      build_id: get(build, 'id'),
+      state: get(get, 'showDiffOverlay') ? 'on' : 'off',
```
maprules1000: I think you want this to be `get(this, 'showDiffOverlay...`... actually I'm not totally sure where `showDiffOverlay` is coming from, but expect that it's not coming from Ember's `get`.

cadeParade: Yes, A+, thanks for this catch.

PR #1041, opened 2019-03-01
Groups are included in deferred rendering when too many snapshots are loaded

cadeParade:
## Description
When there are many snapshots on the page, we defer them rendering until they enter the viewport. Previously that was calculated based on how many snapshots are changed, but now that we have snapshot groups, that number is no longer an accurate count of how many things are displayed. This PR changes the `shouldDeferRendering` property to look at how many snapshot blocks we have, not the number of individual snapshots.
I've also separated out the lazy loading header into its own component that can, in the future be used for any other sort of item that would be displayed in the snapshot list.


tests/integration/components/snapshot-list-test.js
```diff
@@ -52,9 +52,9 @@ describe('Integration: SnapshotList', function() {
     });
 
     it('renders snapshot header placeholder', async function() {
-      expect(SnapshotList.snapshots.length).to.equal(numSnapshots);
-      SnapshotList.snapshots.forEach(snapshot => {
-        expect(snapshot.isLazyRenderHeaderVisible).to.equal(true);
+      expect(SnapshotList.snapshotBlocks.length).to.equal(numSnapshots + 1);
```
maprules1000: this `+ 1` refers to the group made on line 32, right?

cadeParade: Yep.

PR #1048, opened 2019-03-05
Allow keyboard nav to work when groups are showing all snapshots.

cadeParade:
## Description
Currently, keyboard nav in snapshot-list allows you only to keyboard nav between top-level snapshot-blocks. So if you had clicked "Expand Duplicates" of a group and then pushed the down arrow, you would be navigated to the next top-level snapshot-block (aka, the next group or snapshot). Instead, if the group is showing all snapshots, when you click the down arrow, you should be navigated to the first snapshot within the group.

To accomplish this, an action in snapshot-list is called when a group is toggled open or closed. These actions manually add the group's snapshots to the array that keyboard nav is calculated from. Because these actions alter the array in-place, the computed property that updates the keyboard nav index must be explicitly notified (with `notifyPropertyChange`)


app/components/snapshot-group.js
```diff
@@ -1,11 +1,12 @@
...
+import {computed, get, set} from '@ember/object';
 import {filterBy, readOnly} from '@ember/object/computed';
 import SnapshotListItem from 'percy-web/components/snapshot-list-item';
 
 export default SnapshotListItem.extend({
   snapshots: null,
-
   areAllSnapshotsExpanded: false,
+  allowGroupedSnapshotsInKeyboardNav() {},
+  removeGroupedSnapshotsFromKeyboardNav() {},
```
maprules1000: Having empty fns here means the component requires them right?

app/components/snapshot-list.js
```diff
@@ -78,6 +78,25 @@ export default Component.extend({
     updateActiveSnapshotBlockId(newSnapshotBlockId) {
       set(this, 'activeSnapshotBlockId', newSnapshotBlockId);
     },
+    allowGroupedSnapshotsInKeyboardNav(groupedSnapshots) {
```
maprules1000: This ends up overwriting the computed property with a static array. This is not great, but not inherently a problem if that computed prop isn't recalculated for some reason. 

I think a comment about overwriting the computed prop would be good here.

`_allVisibleSnapshotBlocks` shouldn't be recomputed for any reason, right?

cadeParade: Well, it would be recomputed when 
` 'snapshotBlocks.[]',
    'snapshotsUnchanged.[]',
    'isUnchangedSnapshotsVisible',` 
change -- aka when a user clicks "show unchanged snapshots" at the bottom of a build.
This is a good point, whenever a user clicks that, `_allVisibleSnapshotBlocks` will be overwritten. 

I'm going to hold off on this until further discussions about the state of the diff grouping feature.

cadeParade:
Closing for now, since this feature comes with a lot of edge-case-y things which may not be worth the maintenance. But may re-open if this feature becomes requested.

PR #1035, opened 2019-02-26
Org Dashboard

maprules1000:
## Description
This PR adds an organization dashboard which includes a new tab navigation and a project listing page.

The majority of this PR is style tweaks and component variable changes.

The new elements are a `dashboard-nav-wrapper` and an organization index route where the projects are listed.

PR #1053, opened 2019-03-06
Org Settings Refactor

maprules1000:
## Description
Refactor org settings to more easily work with the soon-to-land organization dashboard. This must be merged before #1035.

Many of the refactors in that PR are small. The big ones are the gitlab common refactor, and the pattern of using model hooks.


app/routes/organizations/organization/billing.js
```diff
@@ -33,16 +35,15 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
       currentUsageStats: model.usageStats,
       usageNotificationSetting: model.usageNotificationSetting,
     });
   },
 
   actions: {
     didTransition() {
-      this._super.apply(this, arguments);
-      let organization = this.modelFor('organizations.organization');
+      const organization = this.controller.get('organization');
```
cadeParade: Why no `super`?

app/routes/organizations/organization/integrations/index.js
```diff
@@ -4,22 +4,26 @@ import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-rout
 import {INTEGRATION_TYPES} from 'percy-web/lib/integration-types';
 
 export default Route.extend(AuthenticatedRouteMixin, {
-  setupController(controller, model) {
-    this._super(controller, model);
+  model() {
```
maprules1000: Adding model hooks like this provide very clear documentation about where data is coming from, and it's a pattern I'd like to switch to because it makes code easier to read without having to traverse a bunch of routes.

app/routes/organizations/organization/integrations/gitlab-common.js
```diff
@@ -1,47 +1,76 @@
...
 import Route from '@ember/routing/route';
 import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import {hash} from 'rsvp';
+import {INTEGRATION_TYPES} from 'percy-web/lib/integration-types';
 
 export default Route.extend(AuthenticatedRouteMixin, {
-  newGitlabIntegration: computed('currentIntegration', function() {
-    // This computed property relies on currentIntegration, because the current integration can
-    // change between routes, therefore requiring to recompute the property
-    return this.store.createRecord('version-control-integration', {
```
maprules1000: Creating a record on a computed property is not idea, so I worked to remove this. 

Computed properties are also not very common in routes since they are usually seen as template decorators, which commonly live in controllers. But we don't use controllers so I had to work around that and I was able to figure it out 😄

app/routes/organizations/organization/integrations/gitlab-common.js
```diff
@@ -1,47 +1,76 @@
...
+
     willTransition() {
       let model = this.store.peekAll('version-control-integration').findBy('isNew', true);
       if (model) {
         this.store.unloadRecord(model);
       }
     },
+
     didTransition() {
-      this._super.apply(this, arguments);
```
cadeParade: No super?

app/templates/organizations/organization/integrations/index.hbs
```diff
@@ -7,16 +7,16 @@
...
 
 <section data-test-all-integrations>
-  {{#if model.isVersionControlIntegrated}}
+  {{#if organization.isVersionControlIntegrated}}
     <section class="px-2">
       <div class="bg-white rounded shadow-1 p-2 mb-2">
         <h2 data-test-integrations-installed class="text-xl font-semibold mb-2">Installed Integrations</h2>
         <div class="bg-gray-000 rounded border border-gray-100">
-          {{#each model.versionControlIntegrations as |integration|}}
+          {{#each organization.versionControlIntegrations as |integration|}}
```
cadeParade: I know you didn't write this logic originally, but `versionControlIntegrations` is an async relationship and we are not waiting for it anywhere, so it makes me nervous. Why don't we add an organization's `versionControlIntegrations` to the `integrations/index` model, and then we know it will always be there. Also it makes sense that the version control integrations for an organization would _be_ the model for `integrations/index`.

app/routes/organizations/organization/settings.js
```diff
@@ -1,29 +1,38 @@
...
       this.analytics.track('Settings Viewed', organization);
     },
+
     organizationUpdated(organization) {
-      // If an organization slug changes, we prefer to reload the entire application
+      // If an organization slug changes, reload the entire application
       // to prevent odd bugs, since we rely on the org slug for basically everything.
-      let router = getOwner(this).lookup('router:main');
-      let destinationUrl = router.generate('organization', organization.get('slug'));
+      const router = getOwner(this).lookup('router:main');
```
cadeParade: Let's use the router service?

app/routes/organizations/organization/integrations/index.js
cadeParade: STRONG AGREE

app/routes/organizations/organization/integrations/gitlab-common.js
cadeParade: 💯

app/routes/organizations/organization/billing.js
maprules1000: So, there's no superseding `didTransition` hook with a default behavior, so `this._super` does nothing here and just isn't needed.

app/routes/organizations/organization/integrations/gitlab-common.js
maprules1000: No super needed.

app/routes/organizations/organization/settings.js
maprules1000: 👍

PR #1054, opened 2019-03-06
Diff grouping ui

PixelJanitor:


PixelJanitor:
We could, but then if it’s really long the gradient won’t be there for the bottom ones. We could adjust the gradient percentage, but I don’t think it would give the desired affect. Can try though!

PR #1074, opened 2019-03-22
Grouping Demo Tooltips

maprules1000:
## Description
This adds a new diff grouping tooltip and adds the existing tooltips to the diff grouping interface.

## The Catch
Right now 3 tooltips are only shown on the first snapshot comparison component. This was great, but now grouped diffs introduces a different snapshot comparison component, which becomes first and results in the other single comparison specific tooltips not being shown.

A follow up PR will fix this, but the solution is not currently clear. The problem is that there can be `n` diff groups before a single comparison, so the first single comparison needs to know that it is the first one, which is not easy to do in an elegant way.


app/templates/components/snapshot-group.hbs
```diff
@@ -52,6 +59,13 @@
             </button>
           </div>
           <div class="btn-group">
+            {{demo-tooltip
+              shouldShowTip=(and build.project.isDemo (eq index 0))
+              anchorPlacement='left-corner'
+              key="eight"
+              title="Approving screenshots"
+              message="You can review and approve each chaneg in a group one by one, or you can approve all changes in a group together. All snapshots in a build can be approved all at once with the “Approve All” button at the top of the page."
```
cadeParade:
```suggestion
              message="You can review and approve each change in a group one by one, or you can approve all changes in a group together. All snapshots in a build can be approved all at once with the “Approve All” button at the top of the page."
```

PR #1082, opened 2019-03-27
Linear Demo Tooltips

maprules1000:
## Description
Some users were confused by the order of the tooltips, and some did not know that the pulsating dots could be clicked on. This PR updates the demo tooltips to show in a linear order. Users can step through each one and dismiss them all at any time.

The tooltips change based on the interface for a build: if there is a diff group the tips are diffrerent that if there were no diffs at all. Check out this branch and then check out this build (locally of course): http://dev.percy.local:4200/My-Cool-Demo/Demo-Project-a4cba3534a/builds/1584267


tests/pages/components/demo-tooltip.js
```diff
@@ -2,19 +2,21 @@ import {create, clickable, isVisible} from 'ember-cli-page-object';
...
   TOOLTIP: '.ember-attacher',
-  DISMISS_BUTTON: '[data-test-tooltip-dismiss-button]',
+  NEXT_BUTTON: '[data-test-tooltip-next-button]',
   DISMISS_ALL_BUTTON: '[data-test-tooltip-dismiss-all-button]',
 };
 
 export const DemoTooltip = {
-  scope: SELECTORS.CONTAINER,
+  wrapperScope: SELECTORS.CONTAINER,
+  nextableScope: SELECTORS.NEXTABLE_CONTAINER,
```
maprules1000: There are two different scopes here because the final tooltip does not have a next button, but it was sometimes rendered to the page first. Sometimes. That made the test for the `next` buttion functionality flakey and that was not good.

Breaking this into a separate set of tooltips that can be next-ed fixes the flakey test.

tests/integration/components/demo-tooltip-test.js
```diff
@@ -57,20 +57,6 @@ describe('Integration: DemoTooltipComponent', function() {
       await percySnapshot(this.test);
     });
 
-    it('hides the tooltip and anchor element when dismissed', async function() {
```
maprules1000: Single tooltips can no longer be individually dismissed. This would have made maintaining a linear flow  unnecessarily difficult. Instead, they can be dismissed all at once, any time a user is looking at them.

tests/acceptance/build-test.js
```diff
@@ -777,23 +777,30 @@ describe('Acceptance: Demo Project Build', function() {
...
 
-    await BuildPage.demoTooltips.objectAt(0).clickAnchor();
-    await BuildPage.demoTooltips.objectAt(0).clickDismiss();
+    expect(firstTooltip.classList.contains('ember-attacher-show')).to.equal(true);
+    expect(secondTooltip.classList.contains('ember-attacher-hide')).to.equal(true);
 
-    expect(BuildPage.demoTooltips.objectAt(0).isAnchorVisible).to.equal(false);
+    await BuildPage.nextableDemoTooltips.objectAt(0).clickNext();
+
+    await expect(secondTooltip.classList.contains('ember-attacher-show')).to.equal(true);
```
maprules1000: The animation of the tooltips makes these tests a little flakey. I suspect it has to do with the class change. These tooltips are not easy to test anyway. 

Adding this `await` here removed the flakyness.

app/templates/components/snapshot-group.hbs
```diff
@@ -18,6 +18,7 @@
       </div>
 
       {{demo-tooltip
+        build=build
```
maprules1000: The tooltips need to know info about the build to know the order they will show up in.

app/templates/components/demo-tooltip.hbs
```diff
@@ -1,7 +1,8 @@
-<div data-test-tooltip-wrapper class="{{if shouldHideTooltip 'hidden'}}" style="position:relative;">
+<div data-test-tooltip-wrapper class="{{if shouldHideTooltip 'hidden'}} {{if (not isLastTooltip) 'data-test-nextable-tooltip-wrapper'}}" style="position:relative;">
   <div class="tip-anchor {{anchorPlacement}}" data-test-tooltip-anchor>
     {{#attach-popover
-      class='tip-container'
+      class=(if isLastTooltip 'tip-container' 'nextable tip-container')
```
maprules1000: This and the if statement above it are purely for testing

app/components/demo-tooltip.js
```diff
@@ -22,6 +22,11 @@ export default Component.extend({
 
   shouldHideTooltip: or('_isNotDemoProject', '_isAllHidden', '_isHidden'),
 
+  shouldShowTooltip: computed('tooltips.currentTooltipKey', function() {
```
maprules1000: Determines if a tooltip will be open when a user navigates to a build. Because of the reset, the first tooltip will always be open.

app/components/demo-tooltip.js
```diff
@@ -31,9 +36,27 @@ export default Component.extend({
...
+    if (this.get('tooltips.currentSequence')) {
+      return this.get('tooltips.currentSequence').lastObject === this.key;
+    } else {
+      return false;
+    }
+  }),
+
+  init() {
+    this._super(...arguments);
+    this.get('tooltips').setCurrentBuild(this.get('build'));
```
maprules1000: Set the build on the service so its there and nothing heavier has to be done. This will also help change the flow of the tooltips when the build changes.

app/components/demo-tooltip.js
```diff
@@ -31,9 +36,27 @@ export default Component.extend({
...
+  },
+
   actions: {
     onShow(visible) {
       if (visible) {
+        const currentKey = this.get('key');
+        const build = this.get('build');
+        // reset the canonical index
+        this.get('tooltips').resetTooltipFlow(currentKey, build);
```
maprules1000: A user might close a tooltip and then open some other tooltip and click the next button. Instead of creating a linked list of components (gross) the flow is just reset to the current tooltips every time a new one is opened. This is a noisy, but the `ember-attacher` library leaves a very narrow way to detect change.

app/templates/components/demo-tooltip.hbs
```diff
@@ -16,25 +17,36 @@
...
+            data-test-tooltip-dismiss-all-button
+          >
+            Got it!
+          </button>
+        {{else}}
+          <button
+            type="button"
+            name="button"
+            class="percy-btn"
+            {{action "hideAllTooltips" attacher.hide}}
```
cadeParade: I don't feel too strongly about this, but `hideAllTooltips` action calls `attacher.hide` and then `this.tooltips.hideAll()`. Another way you could write this is `{{action (pipe-action attacher.hide (action "hideAllTooltips"))}}`.
Pipe-action is from `ember-composable-helpers` https://github.com/DockYard/ember-composable-helpers#pipe

app/services/tooltips.js
```diff
@@ -1,22 +1,94 @@
...
 
+  currentSequence: null,
+  tooltipSequenceIndex: 0,
+  tooltipKey: 'one',
+
+  currentTooltipKey: computed('tooltipKey', function() {
+    return this.tooltipKey;
+  }),
+
+  isLastTooltip: computed('tooltipKey', function() {
```
cadeParade: Is it necessary to have `isLastTooltip` in both this service an the component?

app/services/tooltips.js
```diff
@@ -1,22 +1,94 @@
...
+  async next() {
+    // pick the correct scenario array for current scenario
+    const scenario = this.currentSequence;
+    // move to next tooltip by incrementing tooltipSequenceIndex by 1
+    const nextTooltip = scenario[this.tooltipSequenceIndex + 1];
+    this.set('tooltipKey', nextTooltip);
+  },
+
+  async _getCurrentSequence(build) {
+    const snapshots = await build.get('snapshots');
```
cadeParade: Does this need to be async?

app/services/tooltips.js
```diff
@@ -1,22 +1,94 @@
...
+    this.set('tooltipSequenceIndex', index);
+    // set tooltipKey to updated tooltip
+    this.set('tooltipKey', scenario[index]);
+  },
+
+  async next() {
+    // pick the correct scenario array for current scenario
+    const scenario = this.currentSequence;
+    // move to next tooltip by incrementing tooltipSequenceIndex by 1
+    const nextTooltip = scenario[this.tooltipSequenceIndex + 1];
```
cadeParade: What if there is not a next?

app/components/demo-tooltip.js
```diff
@@ -13,43 +12,64 @@ export default Component.extend({
...
+      return this.tooltips.currentSequence.lastObject === this.key;
+    } else {
+      return false;
+    }
+  }),
+
+  shouldShowTooltip: computed('tooltips.currentTooltipKey', function() {
+    // should be visible if isDemo, tooltips are not hidden, & is currentStep
+    const isCurrentStep = this.tooltips.currentTooltipKey === this.key;
+    const isDemo = this.get('build.project.isDemo');
```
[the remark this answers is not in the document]
maprules1000: So. I actually got a console error when I tried the updated syntax: 
```
Uncaught Error: Assertion Failed: You attempted to access the `project` property (of <(unknown):ember1193>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('project')` in this case.
```

So it looks like there are some rare cases where `this.get()` is still needed.

app/services/tooltips.js
maprules1000: Yeah, getting snapshots is unfortunately an async call.

app/services/tooltips.js
maprules1000: Then we should be on the final one and the next button will disappear

app/components/demo-tooltip.js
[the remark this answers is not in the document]
maprules1000: I'll add a comment about this

app/services/tooltips.js
maprules1000: Wow I did not realize that I totally duplicated that property across both 😅  Will fix

PR #1095, opened 2019-04-04
Make diff overlay purple.

timhaines:


timhaines:
This has at least two unexpected diffs on it right now.  https://percy.io/percy/percy-web/builds/1690682/view/110356186/1280?browser=chrome&mode=head and https://percy.io/percy/percy-web/builds/1690682/view/110359382/1280?browser=chrome&mode=diff

maprules1000:
What's up with some of these diffs being weirdly greyed out now? https://percy.io/percy/percy-web/builds/1690682/view/110359382/1280?browser=chrome&mode=head

PR #1096, opened 2019-04-05
Slack internal release

unicornzero:
## Description
This is the initial unpolished interface for the internal release of the Slack integration. The UI will change as we implement the new designs.

## Reviewing this PR
_**Is Covered by Automated Tests?:**_ NO, we'll add tests for the first customer-facing UI release


app/components/forms/slack-integration-config.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default BaseFormComponent.extend({
+  model: alias('slackIntegrationConfig'),
+
+  organization: null,
+  validator: null,
+  notificationTypes: null,
+
+  slackNotificationOptions: SLACK_NOTIFICATION_OPTIONS,
+  selectedProject: computed('model.selectedProjectId', 'organization.projects', function() {
```
cadeParade: You're passing `projects` in from the template and getting that value with `this.get('projects')` on line 22, but this dependency key is getting it from the organizations property. Which is it? Let's be consistent. 

I would lean towards always using the passed in value since `organization.projects` is an async relationship and you don't want to be fetching the projects accidentally. 
If you do that, you'd also want to add `projects: null` under line 16 probably.

app/components/forms/slack-integration-config.js
```diff
@@ -0,0 +1,36 @@
...
+    const projects = this.get('projects');
+    if (!projects) {
+      return null;
+    }
+    return projects.find(x => x.id == this.get('model.selectedProjectId'));
```
cadeParade: Please use a better var name than `x`.

app/components/forms/slack-integration-config.js
```diff
@@ -0,0 +1,36 @@
...
+  }),
+
+  actions: {
+    selectProject(project) {
+      if (project.id != ALL_PROJECTS_ID) {
+        this.get('model').set('projectId', project.id);
```
cadeParade: This is setting a property on the `slackIntegrationConfig` object. Should it be set on the changeset object (since this is a form that saves a changeset)?

app/components/organizations/integrations/slack-config-item.js
```diff
@@ -0,0 +1,33 @@
+import Component from '@ember/component';
+import utils from 'percy-web/lib/utils';
+import {computed} from '@ember/object';
+import {alias} from '@ember/object/computed';
+import {ALL_PROJECTS_ID, ALL_PROJECTS_STRING} from 'percy-web/models/slack-integration-config';
+
+export default Component.extend({
+  isNew: alias('slackIntegrationConfig.isNew'),
```
cadeParade: Prefer `readOnly` over `alias`

app/components/organizations/integrations/slack-config-item.js
```diff
@@ -0,0 +1,33 @@
...
+  projectName: computed('projects', 'slackIntegrationConfig.selectedProjectId', function() {
+    const projects = this.get('projects');
+    const selectedProjectId = this.get('slackIntegrationConfig.selectedProjectId');
+    if (selectedProjectId == ALL_PROJECTS_ID || selectedProjectId == 0) {
+      return ALL_PROJECTS_STRING;
+    }
+    if (!projects || !selectedProjectId) {
+      return null;
+    }
+    const matchingProject = projects.find(x => x.id == selectedProjectId);
```
cadeParade: Better var name than `x`, please.

app/components/organizations/integrations/slack-config-item.js
cadeParade: `projects` are passed into this component, so let's show that with a line `projects: null`.

app/components/organizations/integrations/slack-config-item.js
```diff
@@ -0,0 +1,33 @@
...
+    if (matchingProject) {
+      return matchingProject.name;
+    }
+  }),
+
+  actions: {
+    deleteSlackIntegrationConfig() {
```
cadeParade: Can we move this to the route and pass the actions down? (DDAU -- "data down, actions up" is an Ember pattern.

app/components/organizations/integrations/slack-integration-item.js
```diff
@@ -0,0 +1,35 @@
...
+
+  teamName: readOnly('slackIntegration.teamName'),
+  channelName: readOnly('slackIntegration.channelName'),
+  slackIntegrationConfigs: readOnly('slackIntegration.slackIntegrationConfigs'),
+  savedConfigs: computed('slackIntegrationConfigs', function() {
+    return this.get('slackIntegrationConfigs').filterBy('isNew', false);
+  }),
+
+  actions: {
+    createNewIntegrationConfig() {
```
cadeParade: Can we move this to the route and pass the actions down? (DDAU -- "data down, actions up" is an Ember pattern.

app/components/organizations/integrations/slack-integration-item.js
```diff
@@ -0,0 +1,35 @@
...
+      const newConfig = this.store.createRecord('slackIntegrationConfig', {
+        slackIntegration: this.get('slackIntegration'),
+        notificationTypes: [],
+      });
+      this.set('newConfig', newConfig);
+    },
+
+    deleteSlackIntegration() {
```
cadeParade: Can we move this to the route and pass the actions down? (DDAU -- "data down, actions up" is an Ember pattern.

app/components/organizations/integrations/slack-integration-item.js
```diff
@@ -0,0 +1,35 @@
...
+import {inject as service} from '@ember/service';
+
+export default Component.extend({
+  store: service(),
+  newConfig: null,
+
+  teamName: readOnly('slackIntegration.teamName'),
+  channelName: readOnly('slackIntegration.channelName'),
+  slackIntegrationConfigs: readOnly('slackIntegration.slackIntegrationConfigs'),
+  savedConfigs: computed('slackIntegrationConfigs', function() {
```
cadeParade: This property seems not to be used anywhere? Also there is a `filterBy` computed macro.

app/components/organizations/integrations/slack-settings.js
```diff
@@ -0,0 +1,27 @@
...
+  store: service(),
+  projects: computed('organization.projects', function() {
+    let orgProjects = this.get('organization.projects').map(project => {
+      return {id: project.get('id'), name: project.get('name')};
+    });
+    return [{id: ALL_PROJECTS_ID, name: ALL_PROJECTS_STRING}].concat(orgProjects);
+  }),
+
+  actions: {
+    connectSlackChannel() {
```
cadeParade: Can we move this to the route and pass the actions down? (DDAU -- "data down, actions up" is an Ember pattern.

app/components/organizations/integrations/slack-settings.js
```diff
@@ -0,0 +1,27 @@
+import $ from 'jquery';
+import Component from '@ember/component';
+import {computed} from '@ember/object';
+import {inject as service} from '@ember/service';
+import {ALL_PROJECTS_ID, ALL_PROJECTS_STRING} from 'percy-web/models/slack-integration-config';
+import utils from 'percy-web/lib/utils';
+
+export default Component.extend({
+  store: service(),
+  projects: computed('organization.projects', function() {
```
cadeParade: Two things about this:
- This property is called `projects` which, in this app, always refers to the objects that belong to an organization. In this property, we are not returning those things, we are returning partial objects of those things plus another fake thing. This is constructing a list of objects to use with a dropdown, right? So maybe we could call it `projectOptions` or something similar to differentiate this word from what we would normally think of as projects?
- Why are we computing this list of options here? It seems like it gets passed down through several components without any use. We should calculate it where it is needed.

app/components/organizations/integrations/slack-config-item.js
```diff
@@ -0,0 +1,33 @@
+import Component from '@ember/component';
+import utils from 'percy-web/lib/utils';
+import {computed} from '@ember/object';
+import {alias} from '@ember/object/computed';
+import {ALL_PROJECTS_ID, ALL_PROJECTS_STRING} from 'percy-web/models/slack-integration-config';
+
+export default Component.extend({
+  isNew: alias('slackIntegrationConfig.isNew'),
+  projectName: computed('projects', 'slackIntegrationConfig.selectedProjectId', function() {
```
cadeParade: Because
(1) in the template you only use `projectName` if `isNew` is false and
(2) if `isNew` is false, that means there is a `slackIntegrationConfig` which should always already have a project id associated with it, 
Could this be simplified to:
```
projectName: computed('slackIntegrationConfig.selectedProjectId', function() {
    const project = organization.get('projects').findBy('id', 'slackIntegrationConfig.selectedProjectId');
    return project.get('name');
})
```

Then you wouldn't have to mess about with `projects` property in this component at all, since the `projects` property is actually a set of POJOS for the dropdown in a child component of this? And the template doesn't need those projects unless `isNew` is true? If that's the case too, you could calculate the `projectsOptions` property in the form component where it is needed rather than passing it down through many components, unused.

I am not sure if this is totally comprehensible since these comments are referring to lots of different components, so I'm happy to talk over VC if you want!

app/components/forms/slack-integration-config.js
Robdel12: Also `===` over `==` 😊 (along with the other places)

app/components/forms/slack-integration-config.js
```diff
@@ -0,0 +1,36 @@
...
+    const projects = this.get('projects');
+    if (!projects) {
+      return null;
+    }
+    return projects.find(x => x.id == this.get('model.selectedProjectId'));
+  }),
+
+  actions: {
+    selectProject(project) {
+      if (project.id != ALL_PROJECTS_ID) {
```
Robdel12: Can we change `!=` to `!==`?

cadeParade: oh yeah, def, good catch with this and the above comment!

app/models/organization.js
```diff
@@ -68,11 +83,19 @@ export default DS.Model.extend({
...
+  isSlackIntegrated: gt('slackIntegrations.length', 0),
+  isNotSlackIntegrated: not('isSlackIntegrated'),
+  isSlackAllowed: computed(function() {
+    return this.get('launchDarkly').variation('slack-integration');
+  }),
   availableIntegrations: computed('versionControlIntegrations.[]', function() {
     let integrations = [];
     for (const key of Object.keys(INTEGRATION_TYPES)) {
       let item = INTEGRATION_TYPES[key];
+      if (key == 'slack') {
```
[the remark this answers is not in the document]
unicornzero: Yeah, I struggled trying to decide what we should do here. I ended up starting by doing what seemed like the smallest change, but am open to discussing alternatives. 

The main reason I'm excluding Slack here is to make it easier to have the feature flag around it.

app/components/forms/slack-integration-config.js
unicornzero: Oh, the evil loosey-goosey equality coercion in JS strikes again. *facepalm*

app/components/organizations/integrations/slack-config-item.js
unicornzero: Oh, right! Definitely.

app/components/organizations/integrations/slack-config-item.js
unicornzero: One gotcha is that we don't always have a `projectId`. The `projectId` is null if the config applies to All Projects.

unicornzero: This seems to work:
```
  projectName: computed(
    'organization.projects',
    'slackIntegrationConfig.selectedProjectId',
    function() {
      const selectedProjectId = this.get('slackIntegrationConfig.selectedProjectId');
      if (selectedProjectId == ALL_PROJECTS_ID || selectedProjectId == 0) {
        return ALL_PROJECTS_STRING;
      }
      const project = this.get('organization.projects').findBy('id', selectedProjectId);
      return project.get('name');
    },
  ),
```

Also, the shape of this component will change significantly before customers get access to this code as the new designs are moving the form to a separate route.

unicornzero: Thanks. I follow where you're going with this. I've played around with it a bit this evening and here are my findings:

* One gotcha is that we don't always have a projectId. The projectId is null if the config applies to All Projects.
* The component loads before the `projects` data is available, so I think a guard clause is handy to prevent console errors
* I'm getting undefined for `organization.projects' even after we've received the populated `projects`. Do I need to do something in the route or model to get that data and pass it down? Not sure what's going on there.

Please note that the shape of this component will change significantly before customers get access to this code as the new designs are moving the form to a separate route.

app/routes/organizations/organization/setup/slack-integration.js
```diff
@@ -0,0 +1,62 @@
...
+    afterSetup(organization) {
+      this.replaceWith('organizations.organization.integrations.slack', organization.get('slug'));
+    },
+  },
+
+  afterModel() {
+    const params = this.paramsFor(this.routeName);
+    const organization = this.modelFor('organizations.organization');
+    const slug = organization.get('slug');
+    if (params.error) {
```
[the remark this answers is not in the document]
unicornzero: Yes, if someone cancels out of the Slack auth screen.

app/routes/organizations/organization/setup/slack-integration.js
```diff
@@ -0,0 +1,62 @@
...
+      })
+      .fail(error => {
+        this.replaceWith('organizations.organization.integrations.slack', organization.get('slug'));
+        this.get('flashMessages').danger(`There was a problem connecting to Slack.: ${error}`);
+      });
+  },
+
+  setupController(controller, model) {
+    controller.setProperties({
+      organization: model.organization,
```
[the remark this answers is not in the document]
unicornzero: good find. deleting.

app/routes/organizations/organization/setup/slack-integration.js
```diff
@@ -0,0 +1,56 @@
...
+    const slug = organization.get('slug');
+    if (params.error) {
+      this.replaceWith('organizations.organization.integrations.slack', organization.get('slug'));
+      this.get('flashMessages').danger(`There was a problem connecting to Slack: ${params.error}`);
+    }
+
+    $.ajax({
+      type: 'POST',
+      dataType: 'json',
+      url: utils.buildApiUrl('slackIntegrations', slug),
```
cadeParade: Is it possible to replace these `buildApiUrl` requests with Ember calls? For instance, I believe this could be replaced with 
```
`this.store.createRecord('slackIntegration', {
  code: params.code,
  state: params.state,
  organization: organization
}
```
and write a few extra functions in the adapter?

As much as possible we should keep the use of `buildApiUrl` to the adapter.

then the followups would be 
```
this.store.create()
.then(() => {})
.catch(() => {})
```

cadeParade: There is one other instance in this PR which I think could be replaced like this.

app/models/slack-integration.js
```diff
@@ -0,0 +1,8 @@
+import DS from 'ember-data';
+
+export default DS.Model.extend({
+  organization: DS.belongsTo('organization', {async: false}),
+  slackIntegrationConfigs: DS.hasMany('slack-integration-configs', {async: true}),
```
cadeParade: So we've got a `slack-integration-config` which belongs to a `slack-integration` object. And a `slack-integration` object can have many `slack-integration-configs`. 

The async settings on these two properties are different which could be ok, but let's make sure everything is on purpose. The way it is right now, it is saying:

Whenever I ask for `slack-integration` objects I may not have the `slack-integration-config` objects that are related to it. I will sometimes have to make a separate query for those.

Whenever I ask for `slack-integration-config` objects, I will always have or get the related `slack-integration` object. I will never make a separate query for the `slack-integration` relationship.

Is this what is intended?

app/templates/components/organizations/integrations/slack-settings.hbs
```diff
@@ -0,0 +1,36 @@
...
+        deleteSlackIntegration=deleteSlackIntegration
+        createNewIntegrationConfig=createNewIntegrationConfig
+        deleteSlackIntegrationConfig=deleteSlackIntegrationConfig
+        slug=organization.slug}}
+    {{/each}}
+  {{/if}}
+
+  <div class="flex justify-around px-2 pb-2">
+    <p>
+      {{#percy-btn handleClick=(action "connectSlackChannel" class="percy-btn percy-btn-primary")}}
```
cadeParade: Do we need to pass these classes to our `percy-btn` component? 
It looks like it always has class `percy-btn` and you can pass `isPrimaryButton=true` to get the other class?

cadeParade: Also, the `class` argument is inside the `action` parens so I think it is not doing anything anyway?

app/routes/organizations/organization/integrations/slack.js
```diff
@@ -0,0 +1,53 @@
...
+export default Route.extend(AuthenticatedRouteMixin, {
+  model() {
+    return this.modelFor('organizations.organization');
+  },
+
+  setupController(controller, model) {
+    controller.set('organization', model);
+  },
+
+  actions: {
```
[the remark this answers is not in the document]
unicornzero: Agreed. Adding to the to-do list for the private beta release.

app/models/slack-integration.js
unicornzero: Yeah, that's how I expect it to behave at present. I'm serializing `slack-integrations` on the `organization` response. It didn't seem like we tend to include data two models deep on that response, so I figured we'd need to query for it. I guess I might be able to serialize the `slack-integration-configs` directly on the organization model to avoid the extra api calls. I can explore that option

app/templates/components/organizations/integrations/slack-settings.hbs
unicornzero: Ha! That's where my styling typo was, that errant paren. I overlooked that `isPrimaryButton` option before; love it!

app/models/slack-integration.js
unicornzero: Oh! Looks like I had resolved the serialization issue and no longer need `async: false`. Sweet!

app/routes/organizations/organization/setup/slack-integration.js
unicornzero: Note to self: API changes required.

unicornzero: @cadeParade the other instance of an ajax call is for the slack-integration-request endpoint for which there is no corresponding model. Can chat with you tomorrow to determine whether it's a good for converting to an Ember call or not.

app/routes/organizations/organization/integrations/slack.js
```diff
@@ -0,0 +1,53 @@
...
+    connectSlackChannel() {
+      $.ajax({
+        type: 'POST',
+        url: utils.buildApiUrl(
+          'slackIntegrationRequests',
+          this.paramsFor('organizations.organization').organization_id,
+        ),
+      }).done(response => {
```
unicornzero: @cadeParade this other ajax call has no corresponding model; its job is to fetch the `slack_auth_url` we use to kick-off the oauth handshake. That `slack_auth_url` has a unique `state` uuid in it we use to add security to our handshake process.

app/routes/organizations/organization/setup/slack-integration.js
unicornzero: Note to self: Got this slackIntegration save mostly working, though I've got to do some cleanup or a change so that the unsaved new model isn't loading on the page. (doing an org reload didn't work, so I probably need to filter the displayed collection by persisted record)

app/models/slack-integration.js
unicornzero: Egh, had to do API work to make that true. https://github.com/percy/percy-api/pull/908

unicornzero:
Dependent on https://github.com/percy/percy-api/pull/908

PR #1105, opened 2019-04-09
Sanitize verify-email patch request.

fotinakis:


fotinakis:
verbal LGTM from @cadeParade

PR #1106, opened 2019-04-09
Cleanup images.

fotinakis:
None of these are used anymore, they're all hosted on contentful.


fotinakis:
verbal LGTM from @cadeParade

PR #1107, opened 2019-04-11
fix: exclude redirectPath from silent auth parameters 

djones:
Don't supply a redirectPath for silent auth as it's not an allowed parameter

## Reviewing this PR


cadeParade:
This breaks a lot of stuff so closing for now.

PR #1109, opened 2019-04-12
Customize the install button on the Slack integration item

unicornzero:
## Description
The design for the new Slack integration has a "Connect to Slack" button instead of an "Install" button. This PR allows us to customize the text on that button while still using the same shared component.

## Pictures

image

## Reviewing this PR


app/components/organizations/integrations/integration-item.js
```diff
@@ -60,8 +60,13 @@ export default Component.extend({
     return `organizations.organization.integrations.${this.get('routeSlug')}`;
   }),
 
+  installButtonText: lookup('integrationName', INTEGRATIONS_LOOKUP, 'installButtonText'),
   buttonText: computed('isInstalled', function() {
```
cadeParade: This property should now also depend on `installButtonText`

unicornzero: Oh, totally! Thank you!

PR #1123, opened 2019-04-18
Hide archived projects

cadeParade:
## Description
No one really wants to see their archived projects. You archive a project to get rid of it.
This update hides archived projects by default. 

## Reviewing this PR


app/templates/components/organizations/archived-projects.hbs
```diff
@@ -0,0 +1,18 @@
...
+      {{organizations/project-card
+        organization=organization
+        project=project
+      }}
+    {{else}}
+      <div class="my-2" data-test-no-archived-projects>
+        This organization has no archived projects.
+      </div>
+    {{/each}}
+  {{else if getArchivedProjects.isRunning}}
```
maprules1000: This makes a good case for a component that runs concurrency tasks and shows the loading icon, but that is definitely outside of the scope of this PR.

app/components/organizations/archived-projects.js
```diff
@@ -0,0 +1,15 @@
...
+import {task} from 'ember-concurrency';
+
+export default Component.extend({
+  tagName: '',
+  projectQuery: service(),
+  organization: null,
+  projects: null,
+
+  getArchivedProjects: task(function*(organization) {
+    const projects = yield this.projectQuery.getDisabledProjects(organization);
```
maprules1000: I think this works for this use case, but I'm never crazy about data loading in the component, but otherwise it would _have_ to be loaded on the route, which is not what we want.

Could the task exist on the route and be passed down? I guess that's not really different or useful since it means keeping track of this task all the way down to this component. At least this load call is abstracted out into the service.

maprules1000: So, basically, in summary, this is ok.

app/routes/organization/index.js
```diff
@@ -5,10 +5,11 @@ import {inject as service} from '@ember/service';
 
 export default Route.extend({
   redirects: service(),
+  projectQuery: service(),
 
   model() {
     const organization = this.modelFor('organization');
-    const projects = this.store.query('project', {organization: organization});
+    const projects = this.projectQuery.getEnabledProjects(organization);
```
maprules1000: This is fine, but it could just as easily be `this.store.query('project', {organization})`. Any particular reason to throw this into the service?

app/components/organizations/archived-projects.js
cadeParade: I agree that this should not be on the component. But I think it should be in a controller. And we don't have controllers. The problem with this being on a route is 
(1) a task is not a route-action, so we can't really pass the task down in the right way
(2) the task must set something on its context, which if it it was in the route would be the route. Then we would have to update the controller from the task in the route with disabled projects, which I thought was less good than loading it in the component.

But agree, this is not ideal

app/routes/organization/index.js
cadeParade: It could be for sure. But as we allow more filter query params and dynamic includes, we'll probably want to re-use the same queries a bunch of times, so I'm starting the pattern here.

app/services/project-query.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+
+  getDisabledProjects(organization) {
+    return this.store.query('project', {
+      organization,
+      enabled: false,
+    });
+  },
+
+  getAllProjects(organization) {
```
maprules1000: 👍

PR #1127, opened 2019-04-19
Slack UI updates & config edit functionality

unicornzero:
This PR gets our Slack integration UI another step closer toward the new designs. It also adds the ability to edit 'project configs' (SlackIntegrationConfigs). Oh, and it adds a bunch of tests for new and existing functionality.

_Not included: These are not the final look/feel of these pages, but they're a big step toward the design comps. Design around the tags that show for the selected notification types as well as the notification type selector will be done separately. There should be room for improvement in error and error message handling that is not added with this PR; it's on the to-do list, though._


app/models/slack-integration.js
```diff
@@ -7,4 +10,12 @@ export default DS.Model.extend({
   channelName: DS.attr(),
   code: DS.attr(), // only used during create
   state: DS.attr(), // only used during create
+
+  projects: readOnly('organization.projects'),
+  projectOptions: computed('projects.@each.{id,name}', function() {
```
unicornzero: With the edit/create form moving to a new route, `projectOptions` are needed by two different routes. This seemed as good a place as any to put it because both components have access to this model. 

It does mean that if someone has multiple SlackIntegrations we create this on each one, but that doesn't seem a big trade-off, just something to note.

app/routes/organizations/organization/integrations/slack.js
```diff
@@ -1,57 +1,4 @@
...
-      const confirmationMessage =
-        'Are you sure you want remove this project from this Slack integration?';
-      if (!utils.confirmMessage(confirmationMessage)) {
-        return;
-      }
-      slackIntegrationConfig.destroyRecord();
-    },
-  },
-});
+export default Route.extend(AuthenticatedRouteMixin, {});
```
unicornzero: The work has been moved over to the `/index` and `slack-config` subroutes.

app/routes/organizations/organization/integrations/slack/index.js
```diff
@@ -0,0 +1,40 @@
+import Route from '@ember/routing/route';
+import {run} from '@ember/runloop';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import utils from 'percy-web/lib/utils';
+import $ from 'jquery';
+
+export default Route.extend(AuthenticatedRouteMixin, {
```
unicornzero: This is basically the Slack integrations list page.

mirage/config.js
```diff
@@ -2,7 +2,7 @@ import Mirage from 'ember-cli-mirage';
 
 export default function() {
   // Enable this to see verbose request logging from mirage:
-  // this.logging = true;
+  this.logging = true;
```
cadeParade: Don't commit this plz.

mirage/serializers/organization.js
```diff
@@ -14,6 +14,7 @@ export default JSONAPISerializer.extend({
         'githubIntegrationRequest.createdBy',
         'usageNotificationSettings',
         'slackIntegrations',
+        'projects',
```
cadeParade: Projects are not included in organization queries (check `SHARED_INCLUDES` in the API org controller -- this set of includes should match that array). So this is not a good thing to add to the mirage serializer. I believe it is because of the situation I commented on in app/models/slack-integration.js . We'll need to find another way to get the projects.

mirage/serializers/slack-integration.js
```diff
@@ -0,0 +1,5 @@
+import {JSONAPISerializer} from 'ember-cli-mirage';
+
+export default JSONAPISerializer.extend({
+  include: ['slackIntegrationConfigs'],
```
cadeParade: I also don't see in the API where `slackIntegrationConfigs` are included with the `slackIntegration` requests, so this also is perhaps not a good addition.

tests/acceptance/slack-integration-test.js
```diff
@@ -1,16 +1,18 @@
 import {beforeEach} from 'mocha';
 import {percySnapshot} from 'ember-percy';
+import sinon from 'sinon';
+import {currentRouteName} from '@ember/test-helpers';
 import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
-import SlackIntegrationPage from 'percy-web/tests/pages/organizations/slack-integration-page';
 import utils from 'percy-web/lib/utils';
-import sinon from 'sinon';
+import SlackIntegrationPage from 'percy-web/tests/pages/organizations/slack-integration-page';
```
cadeParade: If this slack config form is inside the slack integration page, we should import the slack config form into the slack integration page object so we would only need one page object in this file.

tests/pages/components/organizations/slack-config-item.js
```diff
@@ -0,0 +1,16 @@
...
+export const SELECTORS = {
+  CONTAINER: '[data-test-slack-config-item]',
+  PROJECT_NAME: '[data-test-project-name]',
+  EDIT_BUTTON: '[data-test-edit-button]',
+};
+
+export const SlackConfigItem = {
+  scope: SELECTORS.CONTAINER,
+  projectName: {scope: SELECTORS.PROJECT_NAME},
+  notificationTypes: collection('code'),
```
cadeParade: Could we use a more specific selector for this? And also maybe put it in the SELECTORS object?

tests/pages/components/organizations/slack-integration-item.js
```diff
@@ -12,6 +13,7 @@ export const SlackIntegrationItem = {
   reminder: {scope: SELECTORS.REMINDER},
   addProjectButton: {scope: SELECTORS.ADD_PROJECT_BUTTON},
   deleteIntegrationButton: {scope: SELECTORS.DELETE_INTEGRATION_BUTTON},
+  configItems: collection(SlackConfigItem.scope, SlackConfigItem),
```
cadeParade: Love it

app/components/forms/slack-integration-config.js
```diff
@@ -35,5 +42,26 @@ export default BaseFormComponent.extend({
       }
       this.send('save');
     },
+
+    saving(promise) {
```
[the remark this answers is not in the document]
unicornzero: Help would be appreciated. My challenge was figuring out how to do the routing right after the save as I wasn't seeing a way to do it right after the whole `save` function resolves. I saw that we did this in a slightly different case elsewhere, but I'd like to see what else we can come up with.

app/components/organizations/integrations/slack-config-item.js
```diff
@@ -9,22 +9,14 @@ export function selectedProjectOption(projectOptions, projectId) {
...
-  deleteSlackIntegrationConfig: null,
 
-  isNew: readOnly('slackIntegrationConfig.isNew'),
+  projectOptions: readOnly('slackIntegrationConfig.slackIntegration.projectOptions'),
 
   projectName: computed('projectOptions', 'slackIntegrationConfig.projectId', function() {
     const selectedProject = selectedProjectOption(
       this.get('projectOptions'),
-      this.get('slackIntegrationConfig.projectId'),
+      Number(this.get('slackIntegrationConfig.projectId')),
```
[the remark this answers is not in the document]
unicornzero: My acceptance test that hit this was failing because the ID we were getting (from mirage, I believe) was a string.

app/models/slack-integration.js
unicornzero: Finding a way to pass it down from the shared route eluded me (aka- my research failed me) but I'd love your help pointing me in the right direction.

app/routes/organizations/organization/integrations/slack.js
unicornzero: This is the shared route, perhaps there's a way to pass projectOptions down from it?

mirage/config.js
unicornzero: Sorry! Thought I already pushed an uncommit of that. goof!

unicornzero:
```suggestion
  // this.logging = true;
```

mirage/serializers/organization.js
unicornzero: Oh, that was silly of me not to check that! Bad assumption on my part; thanks for catching it.

Yes, another way to get projects is needed.

tests/acceptance/slack-integration-test.js
unicornzero: They are actually on separate routes. I just didn't make a wrapper page for the config form because it felt like overkill.

tests/pages/components/organizations/slack-config-item.js
unicornzero: Will do!

mirage/serializers/slack-integration.js
unicornzero: They're actually included on the organization but my attempts to include `slackIntegrationConfigs` or `slackIntegration.slackIntegrationConfigs` on the mirage organization serializer were unsuccessful

app/components/organizations/integrations/slack-details-container.js
```diff
@@ -0,0 +1,26 @@
...
+  showSlackIntegrationList: null,
+  connectSlackChannel: null,
+  deleteSlackIntegration: null,
+
+  // For displaying forms/slack-integration-config
+  showSlackIntegrationConfig: null,
+  deleteSlackIntegrationConfig: null,
+
+  projects: readOnly('organization.projects'),
+  projectOptions: computed('projects.@each.{id,name}', function() {
```
cadeParade: You'll still have to wait for this since `organization.projects` is an async call. I'd recommend using ember concurrency when fetching data outside of a route. You can see an example of this working here:
https://github.com/percy/percy-web/blob/master/app/components/user-menu.js#L25

This action calls perform on a task which sets a property on the component. The template watches the property that is set by the task and displays the property when it is loaded.
If you need it to query for projects on the load of the page, you can put the `perform` command in the component `init` hook.

Here's the docs for this: http://ember-concurrency.com/docs/task-function-syntax but let me or @maprules1000 know if you have any questions.

app/components/organizations/integrations/slack-config-item.js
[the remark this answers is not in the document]
unicornzero: I see that I didn't properly answer this question. Yes, projectId is expected to be a number or null. It is handling it well now, but I see how the code could be cleaner and am making a change now. Thanks for pointing out the ambiguity.

app/components/organizations/integrations/slack-details-container.js
unicornzero: Interesting! I took an initial stab at this locally and don't have it working yet, so I'll take another gander at it in the morning. Thanks!

app/components/forms/base.js
```diff
@@ -16,6 +16,7 @@ export default Component.extend({
   isSaveSuccessful: null,
   errorMessage: null,
   confirmationMessage: null,
+  saveSuccess: null,
```
cadeParade: P0: If you do this, you'll have to put an `if` around calling `saveSuccess`. Otherwise, if there is no `saveSucess` passed in, it will throw an error `null is not a function` or something like that. 

Another option is to initialize optional property with a reasonable default value. In this case, it would be 
`saveSuccess: () => {}`. 
This indicates that you can optionally pass in `saveSuccess` and nothing will break if `saveSuccess` is called if it is not passed in, because it is already a function.

app/components/forms/slack-integration-config.js
```diff
@@ -1,22 +1,30 @@
...
+  approved: 'Builds approved by anyone on the team',
+  finished_unreviewed_snapshots: 'Builds that finished but need review',
   finished_no_changes: 'Builds with no visual changes',
-  finished_auto_approved_branch: 'Builds auto-approved due to branch',
+  finished_auto_approved_branch: 'Builds auto-approved from your auto-approved branches.',
 };
 
 export default BaseFormComponent.extend({
   model: alias('slackIntegrationConfig'),
+  router: service(),
```
cadeParade: Looks like you don't need `router` anymore.

mirage/serializers/slack-integration.js
cadeParade: Let's try to resolve this. These serializers should reflect the API includes as much as possible. I'm happy to get on a call to work it out.

app/components/forms/base.js
unicornzero: Oh! An empty function is the right way to make it optional, like we've done with validators! Gotcha!

app/components/forms/slack-integration-config.js
unicornzero: good catch!

mirage/serializers/slack-integration.js
unicornzero: Since my timing was off with collaborating with you on this today, here's where I'm at:

When I include `'slackIntegrations.slackIntegrationConfigs',` on the mirage organization serializer, they appear in the mirage api response under `included` but they’re not getting associated to things in the data store and not available to my tests. I think this is because the mirage api response is not adding them to `relationships` and I don’t know how to make it do that. :(

(I also tried adding a link under `relationships` but that didn't seem to change things. The relationships in the Ember data models seem right to me. Not sure what to try next.)

image

image






image

I'm confused as to what I'm overlooking here. :(

cadeParade: I've investigated this a bit, and come to the conclusion that this is actually necessary. Organization doesn't have a direct relationship to `slack-integration-config`, but to `slack-integration`. Therefore, the serializer needs to know how to form a relationship between `slack-integration-config` and `slack-integration` which it would do in the `slack-integration-config` serializer rather than the `organization` serializer. 

I guess...:/

PR #1136, opened 2019-04-24
Split usage graph and usage notification blocks into org settings usage tab

cadeParade:
## Description
We're updating the billing page! As part of that, we're splitting the graph and usage notifications into a new org settings tab called "usage".
This PR takes the graph and notification form and puts it in a new usage route. 

The number of plans and the differences between them are really crazy, so there's some weird boolean combinations that are happening (and will happen a few more times in future PRs) :( 
It is not great, but at least is is no longer happening all in the template. 

## Reviewing this PR


app/components/organizations/usage-section.js
```diff
@@ -0,0 +1,22 @@
...
+
+  subscription: readOnly('organization.subscription'),
+  plan: readOnly('subscription.plan'),
+
+  shouldDisplayUsageStats: computed(
+    'organization.billingLocked',
+    'subscription.{isCustom,isTrialOrFree}',
+    'plan.{hasAmount,allowOverages,isStandardAndNotFree}',
+    function() {
+      const isStandardPlan = this.plan.isStandardAndNotFree;
```
cadeParade: This is all garbage but at least it's not in the template any more :(

tests/integration/components/organizations/billing-section-test.js
```diff
@@ -32,113 +32,99 @@ describe('Integration: BillingSection', function() {
...
+        await this.render(hbs`{{organizations/billing-section
+          organization=organization}}`);
 
-        // To prevent an API call and setup for admin tests:
-        organization.set('currentUserIsAdmin', true);
+        expect(BillingSection.billingCardUpdater.isCardUpdaterVisible).to.equal(false);
       });
 
-      it('renders Usage Notification settings', async function() {
+      it('does not display card updater when plan is trial', async function() {
```
maprules1000: It's awesome that the cc field is hidden and all these states are tested 💯

PR #1147, opened 2019-04-30
Redirect to config edit form after connecting a Slack channel

unicornzero:
## Description
After a user selects a Slack channel to connect (in the Slack OAuth screen), they are redirected to `.../setup/slack-integration` where we make a call to our API to complete the rest of the Slack OAuth handshake. When that completes, this PR ensures the user is routed to start configuring that inte

## Reviewing this PR


app/routes/organizations/organization/setup/slack-integration.js
```diff
@@ -32,8 +32,10 @@ export default Route.extend(AuthenticatedRouteMixin, {
       () => {
         organization.reload().then(() => {
           this.replaceWith(
-            'organizations.organization.integrations.slack',
-            organization.get('slug'),
+            'organizations.organization.integrations.slack.slack-config',
+            organization.get('id'),
```
cadeParade: This might need to be `organization.get('slug')`?

unicornzero: You're right, that's more consistent with the rest of our app to use the slug. Changed accordingly.

mirage/config.js
```diff
@@ -310,6 +311,13 @@ export default function() {
   this.get('/repos/:id');
   this.post('/reviews');
 
+  // Slack
+  this.post('/organizations/:organization_id/slack-integrations', function(schema) {
+    return schema.slackIntegrations.create({
+      teamName: faker.company.companyName(),
+      channelName: `#${faker.lorem.slug()}`,
```
cadeParade: Dunno if this is important but slack limits channel names to certain characters and lengths. Not sure if that matters here.

unicornzero: The the output of `faker.lorem.slug()` is very similar to channel names (I don't remember if the character restrictions are identical, but, if not, they're super close). We're not passing those values to Slack at all, I just needed some fake data for our visual diffs.

PR #1148, opened 2019-05-01
Add current plan box to billing page

cadeParade:
## Description
We want to revamp the billing page. Part of that is to update how the "Current plan" section of the page displays. Currently, it's very inconsistent how this displays -- for free and trial plans, there is no box, for self-serve plans, the subscription list shows and the selected plan button is disabled, for legacy and enterprise plans, there is a box with info similar to this update.

This update:
- Adds the same box for every plan and displays varying information based on what data the plan has. 
- Deletes the "Current usage" component and moves the information into the "Current plan" box. It has a much smaller footprint.
- Deletes parts of the subscription list component that display the notices that are now in the bottom of the "current plan" box.

## Pictures
Every combination
Build__6622_-_percy_-_Percy___Visual_testing_as_a_service

Example on a page: 
Build__6622_-_percy_-_Percy___Visual_testing_as_a_service

## Reviewing this PR
_**Is Covered by Automated Tests?: YES/**_
The testing for this feature is a bit unique. Since we have a set number of plans, I've tested those particular combinations of data rather than true unit tests where each conditional variation would be tested. I think the fixture tests make sense for this case since there are a set number of plans and they will not be changing often.


app/models/plan.js
```diff
@@ -1,45 +1,46 @@
...
   amount: DS.attr('number'),
+  intervalCount: DS.attr('number'),
   workerLimit: DS.attr('number'),
   usageIncluded: DS.attr('number'),
   historyLimitDays: DS.attr('number'),
   allowOverages: DS.attr('boolean'),
   overageUnitCost: DS.attr('number'),
+  userLimit: DS.attr('number'),
+  isDeprecated: DS.attr('boolean'),
+  // According to the API, this is true for plans with `trial_period_days` present.
```
maprules1000: Comments like these are awesome!

app/models/subscription.js
```diff
@@ -23,10 +23,9 @@ export default DS.Model.extend({
   }),
   trialStart: DS.attr('date'),
   trialEnd: DS.attr('date'),
-  isTrial: alias('plan.isTrial'),
-  isFree: alias('plan.isFree'),
-  isTrialOrFree: or('plan.isTrial', 'plan.isFree'),
-  isCustom: readOnly('plan.isCustom'),
+  isTrial: readOnly('plan.isTrial'),
+  isFree: readOnly('plan.isFree'),
+  isTrialOrFree: readOnly('plan.isTrialOrFree'),
```
maprules1000: `readOnly` 🙌

app/routes/organizations/organization/billing.js
```diff
@@ -27,7 +30,9 @@ export default Route.extend(AuthenticatedRouteMixin, {
 
         return {
           organization,
-          usageStats: organization.get('subscription.currentUsageStats'),
+          usageStats: shouldIncludeUsageStats
```
maprules1000: This is clever since the call is so expensive

tests/integration/components/organizations/current-plan-test.js
```diff
@@ -0,0 +1,166 @@
...
+        {{organizations/current-plan
+          organization=orgWSponsoredPlan
+        }}
+
+        <div class="m1 font-semibold">Enterprise plan</div>
+        {{organizations/current-plan
+          organization=orgWEnterprisePlan
+        }}
+      `);
+      await percySnapshot(this.test);
```
maprules1000: This is a great test

PR #1150, opened 2019-05-01
Return a default sequence when there is no build present

maprules1000:
## Description
For a brief moment when the demo tooltips are initialized for the first time, there will be no build present to calculate the correct sequence of demo tooltips. This is an ephemeral issue since the sequence will be refreshed when the next tooltip is opened. 

Returning a default sequence when a build is not present solves this issue.

The chosen default sequence matches up with the build that users are automatically redirected to after requesting a new demo project.


maprules1000:
I think I know the issue you're talking about. I'll see where its coming from.

PR #1158, opened 2019-05-03
Update billing page subscription-list to display radio buttons and allow email input.

cadeParade:
## Description
This PR is a complete rewrite of `subscription-list` component. It now always contains a list of three radio buttons, and depending on an org's existing plans either
(a) displays a credit card and email input when the orgs' plan is free or trial
(b) displays only the radio buttons if the org's plan is v3-self serve or "upgradeable" (according to list of upgradeable [plan ids in `subscription-data` file](https://github.com/percy/percy-web/blob/master/app/services/subscription-data.js#L10).

A few of the main strokes:
- The `subscription-list` component previously used a mixin called `billing-updater`. It used this because it was previously possible for more than one credit card + submit button form to be present on the page, and the saving/progress state of each needed to be displayed independently. This update only shows one card submit button at a time, so I've moved the save actions to the route, which is where they should be. This had the side effect of also slightly refactoring `billing-card-updater` to move away from the mixin and to use the route actions.

- Using stripe elements to submit a credit card number requires that the api call have a reference to the stripe element component. That means that for version (a) of subscription list -- where the credit card and email inputs are showing -- both inputs and the submit button need to be inside the yielded area of the stripe elements component.
It also means that version (b) of the form, without the stripe elements component, has to be duplicated to work without that component. That's why there's two submit buttons in the template.

- Visualizing the saving state of the actions was made a bit easier with ember-concurrency tasks, but a bit harder because we needed to know about the saving state in several component layers in the page. Since we don't have controllers, setting computed properties on the route, to be used by the template, doesn't work. So whenever the card/email save action is happening, I've updated the save state of the parent component (`billing-section`) by calling an action passed down from `billing-section`.
 
## Pictures
https://percy.io/percy/percy-web/builds/1830412/view/119061022/1280?browser=chrome&mode=head

## Reviewing this PR
_**Is Covered by Automated Tests?: YES, pretty hard**_


app/components/organizations/billing-card-updater.js
```diff
@@ -1,17 +1,37 @@
 import Component from '@ember/component';
+import StripeOptions from 'percy-web/lib/stripe-elements-options';
 import {readOnly} from '@ember/object/computed';
-import BillingUpdater from 'percy-web/mixins/billing-updater';
 
-export default Component.extend(BillingUpdater, {
+export default Component.extend({
```
cadeParade: This file was refactored to stop using `billingUpdater` mixin and to use the actions from the billing route.

app/components/organizations/current-plan.js
```diff
@@ -2,9 +2,8 @@ import Component from '@ember/component';
 import {computed} from '@ember/object';
 import {and, or, readOnly} from '@ember/object/computed';
 import {inject as service} from '@ember/service';
-import BillingUpdater from 'percy-web/mixins/billing-updater';
 
-export default Component.extend(BillingUpdater, {
+export default Component.extend({
```
cadeParade: This mixin was never needed for this file.

app/components/organizations/subscription-list.js
```diff
@@ -1,23 +1,97 @@
 import Component from '@ember/component';
```
cadeParade: A whole new subscription-list!

This component has two states:
(1) No plan is selected because the user is on a free or trial plan. If no plan is selected, the credit card and billing email inputs show up. This is complicated because in order for us to get the credit card save actions to work, we need the `stripeElement` object, exposed by the stripe card component. So anything that will be submitted with the card inputs needs to be rendered within the stripe element component 🙄

(2) There is already a plan on the org a person is viewing, so no inputs are displayed. Only the radio buttons and "Submit" button is displayed.

The complication added by (1) means that there are 2 submit buttons in the template and therefore two save state managements, and two sets of actions.

app/styles/components/organizations/_billing_section.scss
```diff
@@ -1,70 +1,5 @@
 /* stylelint-disable selector-max-compound-selectors */
 
-.bar {
```
cadeParade: I moved the graph to `usage-section` component, so I also moved the css.

tests/acceptance/billing-test.js
```diff
@@ -0,0 +1,83 @@
+import freezeMoment from '../helpers/freeze-moment';
```
cadeParade: These are not really new...they are moved from organization-test.

app/components/organizations/subscription-list.js
```diff
@@ -1,23 +1,110 @@
...
+  isSubmitWithInputsEnabled: and('isEmailValid', 'isCardValid', 'isNewPlanSelected'),
+
+  subscriptionSaveTask: null,
+  isSubscriptionSaving: readOnly('subscriptionSaveTask.isRunning'),
+
+  emailSaveTask: null,
+  cardSaveTask: null,
+  isEmailOrCardSaving: or('emailSaveTask.isRunning', 'cardSaveTask.isRunning'),
+
+  subscriptionChangeset: computed('subscription.billingEmail', function() {
```
djones: note, don't depend on `subscription.billingEmail` here

app/templates/components/organizations/subscription-list.hbs
```diff
@@ -1,35 +1,95 @@
...
+          <div class="items-start mr-3 pr-2 text-xl font-semibold BillingSection-left-col">
+            {{form-fields/radio
+              data-test-subscription-list-item
+              name='planSelector'
+              value=planData.id
+              title=planData.name
+              disabled=(eq planData.id plan.id)
+              shouldBeChecked=(eq planData.id selectedPlanId)
+              handleRadioClick=(action "clickNewPlan")
+              testLabel=planData.id
```
djones: take a look at just using `testLabel` or `data-test-subscription-list-item` but not both

mirage/factories/subscription.js
```diff
@@ -49,4 +49,18 @@ export default Factory.extend({
   withSponsoredPlan: trait({
     plan: association('sponsored'),
   }),
+
+  withExpiredTrialPlan: trait({
```
djones: this is an unused trait

tests/integration/components/organizations/billing-section-test.js
```diff
@@ -38,37 +42,75 @@ describe('Integration: BillingSection', function() {
...
+    describe('billing card updater', function() {
+      async function _expectBillingCardUpdater(
+        context,
+        planTrait,
+        {isVisible = false, isEmailOrCardSaving = false} = {},
+      ) {
+        organization.set('subscription', make('subscription', planTrait));
+        context.setProperties({isEmailOrCardSaving});
+        await context.render(hbs`{{organizations/billing-section
+          organization=organization
```
djones: note this organization is null.

tests/integration/components/organizations/billing-section-test.js
```diff
@@ -38,37 +42,75 @@ describe('Integration: BillingSection', function() {
...
+          isEmailOrCardSaving=isEmailOrCardSaving
+        }}`);
+
+        expect(BillingSection.billingCardUpdater.isCardUpdaterVisible).to.equal(isVisible);
+      }
+
+      it('does not display when plan is free', async function() {
+        await _expectBillingCardUpdater(this, 'withFreePlan');
```
djones: possibly add `{isVisible: false}` in here just for readability

tests/integration/components/organizations/subscription-list-test.js
```diff
@@ -15,39 +17,375 @@ describe('Integration: SubscriptionList', function() {
...
+        await this.render(hbs`{{organizations/subscription-list
+          organization=organization
+        }}`);
+      });
+
+      it('does not display plan info when there is no plan selected', async function() {
+        expect(SubscriptionList.isPlanInfoVisible).to.equal(false);
+      });
+
+      it('does displys plan info for selected plan when a is plan selected', async function() {
```
djones: typo 💃

tests/integration/components/organizations/subscription-list-test.js
```diff
@@ -15,39 +17,375 @@ describe('Integration: SubscriptionList', function() {
...
+          const flashMessageService = this.owner
+            .lookup('service:flash-messages')
+            .registerTypes(['success']);
+          const flashMessageSuccessStub = sinon.stub(flashMessageService, 'success');
+          this.set('actions', {
+            updateEmail: sinon.stub().returns(resolve()),
+            updateCreditCard: sinon.stub().returns(resolve()),
+          });
+
+          this.setProperties({updateCreditCard, updateEmail});
```
djones: do you need this line still?

tests/integration/components/organizations/subscription-list-test.js
```diff
@@ -15,39 +17,375 @@ describe('Integration: SubscriptionList', function() {
...
+        it('sends updateSubscription when entry is valid', async function() {
+          await SubscriptionList.selectLargePlan();
+          await SubscriptionList.submitNewPlan();
+
+          expect(updateEmail).to.not.have.been.called;
+          expect(updateCreditCard).to.not.have.been.called;
+          expect(updateSubscription).to.have.been.called;
+        });
+
+        it('sends updateSubscription when entry is not valid', async function() {
```
djones: "it does not send"

app/templates/components/organizations/subscription-list.hbs
cadeParade: Deleted `data-test-subscription-list-item`.

mirage/factories/subscription.js
cadeParade: Deleted.

tests/integration/components/organizations/billing-section-test.js
cadeParade: It turns out that it is set in the admin describe block.

PR #1167, opened 2019-05-10
Don't display unpersisted Slack configs or errors if canceling out of Slack OAuth

unicornzero:
## Description
If a user cancels out of authorizing Slack, we should not display an error. We also should return out of the error handler so that we don't try to complete OAuth when we haven't been authorized.

We also need to clear an unpersisted Slack configuration if the user cancels out of the form rather than saving the record.

## Reviewing this PR


app/components/forms/slack-integration-config.js
```diff
@@ -25,6 +25,11 @@ export default BaseFormComponent.extend({
     return selectedProjectOption(this.get('projectOptions'), this.get('changeset.projectId'));
   }),
 
+  willDestroyElement() {
```
cadeParade: P2: Is this the only way that someone might navigate away from this screen? My concern is that it will ALWAYS rollback everything whenever this component is un-rendered. Is that desireable? Is there any way we would want to keep the updates when the component is destroyed?

cadeParade: Is there a cancel action that could do this?

app/routes/organizations/organization/setup/slack-integration.js
```diff
@@ -18,9 +18,11 @@ export default Route.extend(AuthenticatedRouteMixin, {
   afterModel() {
     const params = this.paramsFor(this.routeName);
     const organization = this.modelFor('organizations.organization');
-    if (params.error) {
-      this.replaceWith('organizations.organization.integrations.slack', organization.get('slug'));
-      this.get('flashMessages').danger(`There with your Slack authorization: ${params.error}`);
+    if (params.error === 'access_denied') {
+      return this.replaceWith(
```
cadeParade: P1: Does this remove the flash message when there actually _is_ a problem with the slack integration? (also I just noticed that the previous flash message string is not actually a sentence).

tests/acceptance/slack-integration-test.js
```diff
@@ -171,6 +171,34 @@ describe('Acceptance: Slack Integration', function() {
       await percySnapshot(this.test.fullTitle());
     });
 
+    it('when you cancel out of adding a project, a new record is not displayed', async function() {
+      server.post('/slack-integrations/:slack_integration_id/slack-integration-configs', function(
```
cadeParade: P2: It looks like you're not asserting anything new in the custom endpoint here -- is it important that the project Id be null in this test?-- so you could use a generic one from `mirage/config.js`.

tests/acceptance/slack-integration-test.js
```diff
@@ -171,6 +171,34 @@ describe('Acceptance: Slack Integration', function() {
...
+      });
+
+      await SlackIntegrationPage.visitSlackIntegration({orgSlug: organization.slug});
+      expect(SlackIntegrationPage.integrationItems[0].configItems.length).to.equal(3);
+
+      await SlackIntegrationPage.integrationItems[0].addProjectButton.click();
+      expect(currentRouteName()).to.equal(
+        'organizations.organization.integrations.slack.slack-config',
+      );
+      await percySnapshot(this.test.fullTitle() + ' | new slack-config form');
```
cadeParade: P3: Do you expect this snapshot to be different than the one on line 166? If not, we probably don't need it.  Also, you've tested all these mechanics in the previous test, so we probably don't need to assert each route name again...if it goes to the wrong route, the previous test will fail and this one will fail too because it won't be able to find the elements its looking for.

app/routes/organizations/organization/setup/slack-integration.js
unicornzero: This is to handle the case when you cancel out of the Slack Auth screen. Payton pointed out that if a user intentionally backs out that we shouldn't throw an error.

cadeParade: Right. But What if there _is_ an error?

app/components/forms/slack-integration-config.js
unicornzero: I only want to persist the changes if they click the Save button. I can dig deeper into the existing code to try to find an alternative, though.

app/routes/organizations/organization/setup/slack-integration.js
unicornzero: Ooooooh, someone needs more coffee, and that someone is me.... Yes. You're totally right.

tests/acceptance/slack-integration-test.js
unicornzero: took out the ones in the above test. good point.

PR #1190, opened 2019-05-23
Comment components and integration tests

cadeParade:
## Description
This PR adds the components for comments, threads, and replies along with related integration tests.

Check out the tests for a list of behaviors covered in this PR. They are pretty comprehensive.

## Reviewing this PR


app/components/collaboration/new-thread.js
```diff
@@ -0,0 +1,24 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+  tagName: '',
+  shouldShowNewCommentInput: false,
+  isReply: false,
```
djones: This appears redundant

app/templates/components/collaboration/comment-thread.hbs
```diff
@@ -0,0 +1,37 @@
...
+
+    {{#if areCommentsCollapsed}}
+      {{collaboration/collaboration-comment
+        comment=commentThread.firstComment
+        isFirstComment=true
+      }}
+      <div class="collaboration-content-inset">
+        {{collaboration/collaboration-comment
+          comment=commentThread.secondComment
+          isFirstComment=false
```
djones: how about we make `isFirstComment` default to `false` and then only supply that arg if we need to flick it to `true`.

tests/integration/components/collaboration/collaboration-comment-reply-test.js
```diff
@@ -0,0 +1,88 @@
...
+
+      await CommentReply.expandTextarea();
+    });
+
+    it('submit button is disabled by default', async function() {
+      expect(CommentReply.submit.isDisabled).to.equal(true);
+      await percySnapshot(this.test);
+    });
+
+    it('submit button is enabled when comment body is not emtpy', async function() {
```
djones:
```suggestion
    it('submit button is enabled when comment body is not empty', async function() {
```

PR #1191, opened 2019-05-24
Add bitbucket cloud integration behind launch darkly flag

timhaines:
Adds a way to install and manager the Bitbucket Cloud integration.  A bitbucket-cloud-integration launch darkly flag controls whether it shows as an option or not.

**Important:** Currently this must be run against percy-api running locally, with ngrok running.

This PR previously contained a fix for Launch Darkly.  That's been moved to #1196 now.  Thanks @cadeParade for the great help getting that PR over the line.

Todos:
- [x] Fix the "Remove Integration" button
- [x] bitbucket cloud setup handler acceptance tests
- [x] integrations index page acceptance test
- [x] Finish setting up `bitbucketCloudUrls`


app/components/organizations/bitbucket-cloud-integrator.js
```diff
@@ -0,0 +1,23 @@
...
+  organization: null,
+  classes: null,
+
+  currentIntegration: alias('organization.bitbucketCloudIntegration'),
+
+  actions: {
+    triggerInstallation() {
+      let url = this.get('bitbucketCloudIntegrationUrl');
+      let organization = this.get('organization');
+      window.location.href = url.replace('%@', organization.slug);
```
cadeParade: We usually use one of the util functions here: 
https://github.com/percy/percy-web/blob/master/app/lib/utils.js#L67
or
https://github.com/percy/percy-web/blob/master/app/lib/utils.js#L77

so the url change can be stubbed in tests. Otherwise your tests will navigate to a new page and that will be the end of your tests.

app/components/organizations/integrations/bitbucket-cloud-settings.js
```diff
@@ -0,0 +1,11 @@
+import Component from '@ember/component';
+import {alias} from '@ember/object/computed';
+import {inject as service} from '@ember/service';
+
+export default Component.extend({
+  session: service(),
+
+  currentUser: alias('session.currentUser'),
```
cadeParade: We prefer `readOnly` to `alias`. `readOnly` is, as it says, a read operation only. Using `alias` allows editing the property it is aliasing. (There are still instances of `alias` used in percy-web though)

tests/acceptance/bitbucket-cloud-app-install-test.js
```diff
@@ -0,0 +1,78 @@
...
+      );
+    });
+
+    it('shows the installed state when the install is in successful', async function() {
+      await BitbucketCloudSettings.visitBitbucketCloudSettings({orgSlug: organization.slug});
+      expect(currentRouteName()).to.equal(
+        'organizations.organization.integrations.bitbucket-cloud',
+      );
+
+      expect(BitbucketCloudSettings.isBitbucketCloudAppSuccessStateVisable).to.equal(true);
```
cadeParade:
```suggestion
      expect(BitbucketCloudSettings.isBitbucketCloudAppSuccessStateVisible).to.equal(true);
```

cadeParade: (small typo)

app/components/organizations/integrations/bitbucket-cloud-settings.js
timhaines: Going to stick with `alias` here to keep it consistent with the other integrations components.  Perhaps they can all be updated together at some future point.

app/components/organizations/bitbucket-cloud-integrator.js
timhaines: Fixed.

tests/acceptance/bitbucket-cloud-app-install-test.js
timhaines: Fixed.

app/templates/components/organizations/bitbucket-cloud-integrator.hbs
```diff
@@ -0,0 +1,20 @@
...
+  {{organizations/bitbucket-cloud-integration organization=organization}}
+{{else}}
+  <div class="mb-2">
+    <a
+      data-test-bitbucket-cloud-app-install-button
+      href="{{bitbucketCloudIntegrationUrl}}"
+      class="percy-btn percy-btn-primary"
+      target="_blank"
+      rel="noopener"
+      {{action "triggerInstallation"}}>
```
cadeParade: Do you need both the `href` and the action?
Seems like they would both do the same thing? 
I think a better way to do it would be to have a computed property on the component, with the full string including the org name. Then, use that string as the `href` property and remove the action from this `a`.

ex:
```
bitbucketCloudIntegrationUrl: computed('organization.slug', function() {
  const url = config.APP.bitbucketCloudUrls.integration;
  return url.replace('%@', this.get('organization.slug'))
}
```

timhaines: Updated.  This is much nicer. Refrained from updating the GitHub integrator that was implemented in a similar way.

PR #1196, opened 2019-05-26
Fix Launch Darkly user identification

timhaines:
There's a current problem that if a user starts their session as an anonymous user, when they sign in, we don't re-identify them in the Launch Darkly Context.  This PR fixes that.


app/services/session.js
```diff
@@ -60,6 +61,7 @@ export default SessionService.extend({
       this._setupSentry(user);
       this._setupAnalytics(user);
       this._setupIntercom(user);
+      this._setupLaunchDarkly(user);
       resolve();
```
timhaines: This resolve call ensures the promise gets resolved immediately without waiting for any response.  I think this makes the promise redundant?  It's not adding anything?

app/services/session.js
```diff
@@ -18,7 +19,7 @@ export default SessionService.extend({
     if (this.get('isAuthenticated')) {
       return (
         this.forceReloadUser()
-          .then(user => {
+          .then(async user => {
             this.set('currentUser', user);
             this._setupThirdPartyUserContexts(user);
```
timhaines: Note, nothing is done with the promise returned from `_setupThirdPartyUserContexts`.

timhaines:
@cadeParade this is the refactor we wrote on Friday afternoon pulled out into it's own PR, as it's unrelated to bitbucket.

I've played around writing a test for it, but haven't got it to a good state yet, because of the async nature of `_setupLaunchDarkly` and the fact we're not waiting for it anywhere.

Can we safely return from `loadCurrentUser` without waiting for launchdarkly?  We weren't waiting for it previously either.

timhaines:
@cadeParade Have pushed some passing tests that exercise the right things now.  It is a bit weird that one tests a function gets called, and the other tests that function in turn calls identify.  Not sure if we could wrap this in to one test?

tests/unit/services/session-test.js
```diff
@@ -84,6 +84,32 @@ describe('SessionService', function() {
...
+        return promise.then(() => {
+          expect(_setupLaunchDarklyStub).to.have.been.calledWith(user);
+        });
+      });
+
+      it("calls Launch Darkly's identify via _setupLaunchDarkly", function() {
+        const identifyStub = sinon.stub();
+        subject.set('launchDarkly.identify', identifyStub);
+
+        const promise = subject._setupLaunchDarkly(user);
```
cadeParade:
```suggestion
        const promise = subject.loadCurrentUser(user);
```

Re: your question about two tests: 
I think the first test is not necessary...we don't care that it calls `_setupLaunchDarkly` specifically. What we care about is that when `loadCurrentUser` is called, it calls the right thing on the launch darkly service. So changing line 102 to call `loadCurrentUser` and checking that the identify method got called, seems sufficient to me.

timhaines: @cadeParade That test would fail.  The `loadCurrentUser` returns before the `_setupLaunchDarkly` async function is fully resolved.  Do you know how to nicely make it wait?

PR #1202, opened 2019-05-31
sort snapshots by commentThreads count

unicornzero:
## Description

We sort snapshots based on several criteria. First, snapshots with diffs are sorted before snapshots without diffs. Next, snapshots are then sorted by their comment thread count before they're sorted by existing width and diff ratio logic.

## Reviewing this PR


app/lib/snapshot-sort.js
```diff
@@ -23,7 +23,12 @@ export default function(snapshots, browser) {
     } else if (bHasDiffs && aHasNoDiffs) {
       return 1;
     }
-
+    // // Next prioritize snapshots with more comment-threads
+    const aCommentThreadsCount = get(a, 'commentThreads.length') || 0;
```
cadeParade: Why is this `|| 0`? What does this protect against?

unicornzero: That protects against `undefined` which our `>` evaluation doesn't handle as zero. So, if for some reason there wasn't a commentThreads attribute on a snapshot, it will still sort correctly.

unicornzero: We discussed in Slack that `get` returns undefined rather than erroring if `commentThreads` is not present, the `|| 0` handles the case where the `commentThreads` attribute is not on a snapshot for some reason.

PR #1216, opened 2019-06-06
group cover with comments

unicornzero:
## Description
* Displays all snapshots with comments in its group's cover
* Makes commenting work on snapshots within groups
* Changes some eslint comments that were problematic with prettier

## Pictures

image

## Reviewing this PR


app/components/snapshot-group.js
```diff
@@ -17,6 +20,8 @@ export default SnapshotListItem.extend({
   _unapprovedSnapshots: filterBy('snapshots', 'isUnreviewed'),
   numUnapprovedSnapshots: readOnly('_unapprovedSnapshots.length'),
   _isApproved: readOnly('isGroupApproved'),
+  openCommentThreadsSnapshots: filterBy('snapshots', 'hasOpenCommentThreads'),
+  hasOpenCommentThreadsSnapshots: gt('openCommentThreadsSnapshots.length', 0),
```
cadeParade: perhaps `hasSnapshotsWithOpenCommentThreads`?

app/models/snapshot.js
```diff
@@ -27,6 +27,8 @@ export default DS.Model.extend({
   }),
 
   commentThreads: DS.hasMany('commentThreads', {async: false}),
+  openCommentThreads: filterBy('commentThreads', 'isOpen'),
+  hasOpenCommentThreads: gt('openCommentThreads.length', 0),
```
cadeParade: This implementation is fine, just fyi there is also a macro `empty` and `notEmpty`. https://api.emberjs.com/ember/3.4/functions/@ember%2Fobject%2Fcomputed/notEmpty

app/templates/components/snapshot-group.hbs
```diff
@@ -85,40 +85,69 @@
...
-              <div class="absolute bg-white rounded border pin-t-3 pin-r-3 pin-l-1 pin-b-3 shadow-lg"></div>
-              <div class="relative pl-0 pb-2">
-                {{comparison-viewer
-                  comparison=selectedComparison
+        {{#if hasOpenCommentThreadsSnapshots}}
+          <div class="relative pt-2 bg-gray-000">
+            <div style="left: 28px;" class="snapshot-group-indent-line"></div>
+            {{#each openCommentThreadsSnapshots as |snapshot|}}
+              <div class="flex bg-gray-000">
+                <div style="padding: 11px 16px 10px;" class="mb-2">
```
cadeParade: What are these heights? (why are they different than before?)

app/templates/components/snapshot-group.hbs
```diff
@@ -85,40 +85,69 @@
...
-            <div class="relative flex-1 p-2">
-              <div class="absolute bg-white rounded border pin-t-4 pin-r-4 pin-l pin-b-2 shadow-lg"></div>
-              <div class="absolute bg-white rounded border pin-t-3 pin-r-3 pin-l-1 pin-b-3 shadow-lg"></div>
-              <div class="relative pl-0 pb-2">
-                {{comparison-viewer
-                  comparison=selectedComparison
+        {{#if hasOpenCommentThreadsSnapshots}}
+          <div class="relative pt-2 bg-gray-000">
+            <div style="left: 28px;" class="snapshot-group-indent-line"></div>
+            {{#each openCommentThreadsSnapshots as |snapshot|}}
```
cadeParade: Would `snapshot-list` component be helpful here? (I am not sure, but it lists snapshots...)

tests/acceptance/build-test.js
```diff
@@ -1,20 +1,75 @@
...
+      expect(firstSnapshot.collaborationPanel.isVisible).to.equal(true);
+      expect(firstSnapshot.commentThreads.length).to.equal(3);
+      expect(firstSnapshot.header.numOpenCommentThreads).to.equal('3');
+      await percySnapshot(this.test);
+    });
+  }
+
+  function createsCommentReplyOnFirstSnapshot() {
+    it('can create a new comment reply', async function() {
+      //this was falsely passing because the other comments were hidden
```
cadeParade: It is good if this is not falsely passing anymore, but maybe instead of this comment we could add either:
1) advice about how to make sure this test is not falsely passing or even better
2) write a test that will fail if it is broken, so then we don't need a comment.

tests/acceptance/build-test.js
```diff
@@ -99,7 +154,8 @@ describe('Acceptance: Build', function() {
       });
     });
 
-    it('displays snapshots in the correct order, before and after approval when build is finished', async function() { // eslint-disable-line
+    // eslint-disable-next-line
```
cadeParade: Running prettier on files with this line moves the comment to the next line...undoing the usefulness of the 'long line is ok' comment :(

In any case, this comment does not apply to anything anymore.

unicornzero: We talked through this one (regarding eslint-disable-line vs eslint-disable-next-line). Marking resolved.

tests/acceptance/build-test.js
unicornzero: ah, yes, that comment is to be deleted fo sho. will upgrade comment/code.

app/templates/components/snapshot-group.hbs
unicornzero: Looks like snapshot-list renders this component, so I think it's a parent component to this, not a child. #recursion

iterating through snapshots to display snapshot-viewer components is the existing pattern we had in this file for when a group is expanded.

app/models/snapshot.js
unicornzero: Thanks, I've changed my other use of `gt` in this PR to use `notEmpty`. In this case, I found it doesn't properly protect against if the `commentThreads` attribute is absent for some reason (as is the case with some of our snapshot ordering tests).

app/templates/components/snapshot-group.hbs
```diff
@@ -28,17 +28,15 @@
 
     <div class="flex justify-end">
       <div class="btn-toolbar">
-        {{#if (not areAllSnapshotsExpanded)}}
+
+        {{#if isGroupCollapsed}}
```
unicornzero: Not a *necessary* change, but this file was getting super unreadable and I hope this and some other naming changes help. I'm open to tweaking/changing names as desired.

unicornzero: And with the `isExpanded` variable name used below, I wanted to try to make it a little less name-overlappy.

app/templates/components/snapshot-group.hbs
unicornzero: Git's confusing us by which group it thinks we changed.

app/templates/components/snapshot-group.hbs
```diff
@@ -119,42 +107,47 @@
             </div>
           </div>
         </div>
+
       {{else}}
         <div class="relative pt-2 bg-gray-000">
           <div style="left: 28px;" class="snapshot-group-indent-line"></div>
-          {{#each snapshots as |snapshot|}}
+          {{#each snapshotsToDisplay as |snapshot|}}
```
unicornzero: This little `snapshotsToDisplay` can either show the commented snapshots for the cover or all snapshots if the group is expanded. Yay!

tests/pages/components/snapshot-block.js
```diff
@@ -14,7 +14,7 @@ export const snapshotBlock = {
   snapshotGroup: snapshotGroup,
   snapshotViewer: SnapshotViewer,
   _block: getter(function() {
-    return this.isSnapshot ? this.snapshotViewer : this.snapshotGroup;
+    return this.isGroup ? this.snapshotGroup : this.snapshotViewer;
```
unicornzero: This bugger threw things off in one of my tests. Works in more cases if we check group status first.

app/templates/components/snapshot-group.hbs
cadeParade: seems great.

app/templates/components/snapshot-group.hbs
cadeParade: Yay!

tests/pages/components/snapshot-block.js
unicornzero: We're all good. :) If I run into these tests again I might take you up on that later, though!

PR #1222, opened 2019-06-10
Cmd+Enter textarea submit

cadeParade:
## Description
Allow an optional `cmd+enter` to trigger an action on textareas. Also allow an optional action to be triggered by `escape`. Use these in new thread and comment reply textareas.
## Reviewing this PR


app/components/percy-textarea.js
```diff
@@ -0,0 +1,19 @@
+import TextArea from '@ember/component/text-area';
+import {EKMixin, keyDown} from 'ember-keyboard';
+import {on} from '@ember/object/evented';
+
+export default TextArea.extend(EKMixin, {
+  attributeBindings: ['data-test-percy-textarea'],
```
unicornzero: For my own learning, why have we added this line? Is it so that we can have this data attribute without needing a `.hbs` template for the component?

app/components/snapshot-list.js
```diff
@@ -80,21 +80,21 @@ export default Component.extend(EKMixin, {
     this.set('keyboardActivated', true);
   },
 
-  onDKeyPress: on(keyUp('KeyD'), function() {
+  onDKeyPress: on(keyDown('KeyD'), function() {
```
unicornzero: For my own learning, why do we prefer keyDown over keyUp?

app/templates/components/collaboration/new-thread.hbs
```diff
@@ -9,9 +9,11 @@
       </div>
       <div class="flex-1 collaboration-content-inset">
         <div class="relative">
-          {{textarea
+          {{percy-textarea
```
unicornzero: I'm digging this. Having our own textarea component with keybindings built in is super handy!

tests/pages/components/collaboration/collaboration-comment-reply.js
```diff
@@ -28,6 +29,8 @@ export const collaborationCommentReply = {
   isCollapsed: getter(function() {
     return !this.isExpanded;
   }),
+
+  percyTextarea,
```
unicornzero: 🔥

tests/pages/components/percy-textarea.js
```diff
@@ -0,0 +1,22 @@
...
+import {keyDown} from 'ember-keyboard/test-support/test-helpers';
+
+const SELECTORS = {
+  SCOPE: '[data-test-percy-textarea]',
+};
+
+export const percyTextarea = {
+  scope: SELECTORS.SCOPE,
+
+  async cmdEnter() {
```
unicornzero: Nice!

app/templates/components/collaboration/new-thread.hbs
cadeParade: Yeah! I named it more generically after thinking about maybe other extensions we could build into it like....emoji support 🤷‍♀

app/components/snapshot-list.js
cadeParade: Well, it is not a great story. I used `keyUp` previously on this component, but had trouble with `keyUp` on the textarea component. I think maybe `keyUp` event on mac on these damn touch bars doesn't work? I am not sure and also don't remember too much about my reasoning, but `keyDown seems to work more consistently and there's no reason not to chose one event to use regularly.

app/components/percy-textarea.js
cadeParade: Sure!
If you recall, an ember component (in the current world of ember) has a wrapper div associated with the component. This wrapper div does not appear in the templates anywhere. So to assign classes or attributes to this wrapper component you either need to do it in the js file, as I've done here (for classes it would be either `classNames: ['foo', 'whatever']`, or if they are conditional; `classNameBindings: ['isLoading:is-loading', etc]`) or you could pass the property in as:
```
{{percy-textarea
  data-test-percy-textarea=true
}}
```

Since I want `data-test-percy-textarea` to be true in every instance of this component, I put the attr binding in the js file.

unicornzero: Thanks!

app/components/snapshot-list.js
unicornzero: ok! :)

app/templates/components/collaboration/new-thread.hbs
unicornzero: 🤷‍♀🤷‍♀🤷‍♀

PR #1226, opened 2019-06-13
Initiate silent auth when in inconsistent session state

cadeParade:
## Description
When loading the app, figure out if we are in the inconsistent sessions state. If we are inconsistent because the API has a user and the FE doesn't have a user, initiate silent auth. If that fails or we are inconsistent in another way, log the user all the way out.
## Reviewing this PR


app/routes/organizations/organization/integrations/bitbucket-cloud.js
```diff
@@ -7,10 +7,10 @@ export default Route.extend(AuthenticatedRouteMixin, {
   session: service(),
   flashMessages: service(),
 
-  beforeModel() {
+  async beforeModel() {
     // If we don't force reload user on this page,
     // we don't get the associated Identities
-    return this.get('session').forceReloadUser();
+    return await this.get('session').forceReloadUser();
```
Robdel12: Hm, I don't expect this to change anything. I'm pretty sure returning an await is the same as returning a promise. 

And I think the model hooks are promise aware? Interesting stuff

PR #1239, opened 2019-06-20
move project integrations to separate integrations tab within projects

PixelJanitor:


app/router.js
```diff
@@ -115,11 +115,10 @@ Router.map(function() {
     // projects namespace. Org-level settings and such should go in the above "organizations" route.
     this.route('index', {path: '/'});
     this.route('project', {path: '/:project_id'}, function() {
-      this.route('settings', {}, function() {
-        this.route('integrations', {}, function() {
-          this.route('webhooks', {}, function() {
-            this.route('webhook-config', {path: '/:webhook_config_id'});
-          });
+      this.route('settings', {}, function() {});
+      this.route('integrations', {}, function() {
```
fotinakis: @cadeParade @PixelJanitor -- This change will break anything that references this route, right? which there are a bunch of things that reference `organization.project.settings.integrations`

fotinakis: Ah nevermind, my PR hadn't refreshed, looks like those got updated.

cadeParade: I've tried to update all of them. LMK if you see something that's off in the app.

PR #1240, opened 2019-06-20
@mentions

cadeParade:
## Description
This PR introduces
(1) `mentionable-textarea` component which is a yielded component that is made to wrap a textarea. This component requires a tribute config object (see (2)) and optionally accepts an action to take once a mentionable item is selected.
(2) A mentionables service which has methods for setting up a tribute. The main method right now is `generateOrgUserCollection`. 

-    When `generateOrgUserCollection` is called with an organization, it loads `organization-users` for that org. In particular, note that it uses the `load` method on `organization.hasMany` reference object. This allows it to load `organization-users` for an `organization` only once. If we did not use this method here, a new request to `organization-users` endpoint would be made for every comment textarea on the page. Using this API allows us to make only one request for `organization-users` per page load. https://api.emberjs.com/ember-data/3.10/classes/DS.HasManyReference/methods/push?anchor=load "Loads the relationship if it is not already loaded. If the relationship is already loaded this method does not trigger a new load. "
-    It also provides `verifyMentions` method which makes sure that when the request goes out, we are not tagging any users whose names were deleted from the text.

 
## Pictures
@mention

## Reviewing this PR


app/services/mentionables.js
```diff
@@ -0,0 +1,63 @@
+import Service from '@ember/service';
+import {task} from 'ember-concurrency';
+
```
djones: perhaps a link to https://github.com/zurb/tribute#a-collection would be useful since there are some library specific things being set here

app/components/collaboration/collaboration-comment-reply.js
```diff
@@ -2,19 +2,28 @@ import Component from '@ember/component';
...
   createComment: null,
+  mentionedUsers: null,
 
   session: service(),
   flashMessages: service(),
   currentUser: readOnly('session.currentUser'),
   isCommentSaving: readOnly('commentSaveTask.isRunning'),
 
+  // Setup @mentions
+  tributeCollections: computed(function() {
```
djones: `tributeCollections` -> `tributeConfigs`. They are strictly called collections in Tribute but that's pretty confusing. These collections include things like the template for rendering and other configuration options, so "configs" feels better to me

PR #1250, opened 2019-06-27
User notification settings form

cadeParade:
## Description
I present to you the latest in form technology: This form with two checkboxes. 
Yes, you may be wondering how it is possible to build a form with not one but two checkboxes, but friends, I am here to show you that we can do it.

Ok but for real, this form has some cool stuff.
(1) It is starting a new form pattern using contextual components. The new `percy-form` component yields out several form components. These components are already hooked up with a lot of cool stuff via `percy-form` so a developer can just use the yielded components and not think too much about what is inside these components. The yielded components are very thin wrappers on top of plain HTML elements and are pre-set-up with form submit actions, label linking and loading state. For a short introduction to this concept (and to discover I wholesale stole it from...) check out the first two videos of this ember map series: https://embermap.com/topics/forms

(2) We've got this array `notificationTypes` on the `userNotificationSetting` model, which is an array of strings. This is problematic for a couple reasons but one is that it comes in as a JSON array and is not converted to an Ember object. Therefore, we cannot observe properties on it or do any of the nice things that come with Ember objects. BUT, with a bit of cleverness, we can use custom computed properties (`twoWayArrayToggle`) to allow these options to be toggled as if they were ember objects via the checkbox component.

Other things of note:
- This `percy-form` actually uses the `form` element as it was meant to be used (as opposed to our current form implementations which disable the submit action (eg: `<form action="javascript:;">`. So when we click the submit button we call the form submit action AND when we push enter we call the form submit action. 
- The `percy-form` component expects an action to be passed in but inside the form, it calls the action within a task. Therefore, the save button and the rest of the form can be aware of whether the form is saving.
- The `percy-form` component can optionally accept a task by passing in a task with the ember-concurrency perform helper (as in this PR).
- We named this record type singular in the API (`user-notification-setting`) since there is only one per user. Ember Data expects resources to be named plurally, so I've added a line to the inflector that tells Ember to never make this model plural. This allows Mirage and other parts of the web-app to also respect the singular version of `user-notfication-setting`.
- There are no integration tests for this form, only acceptance tests. This is because (1) the form is fairly simple and the most important thing it does is save data correctly and (2) The action for this form is now contained within the form component rather than passed in from the route. This is a thing we can probably do now, but it means that integration testing is harder because we'd have to handle data loading and server calls in the test. Possible, but going to hold off for now.

## Pictures
https://percy.io/percy/percy-web/builds/2107419/view/136921930/1280?browser=chrome&mode=diff

## Reviewing this PR


app/components/percy-form.js
```diff
@@ -0,0 +1,20 @@
...
+  onSubmit() {},
+
+  submit(event) {
+    event.preventDefault();
+    this.submitTask.perform();
+  },
+
+  submitTask: task(function*() {
+    // This tells the task to wait at least 500ms for the task to complete, even if the actual
+    // action finishes sooner so the user has a nice indicator that the form is doing something.
```
unicornzero: <3

app/components/percy-form/checkbox.js
```diff
@@ -0,0 +1,18 @@
...
+  attributeBindings: ['name', 'type', 'checked'],
+  checked: false,
+
+  // Default behavior of checkbox component. When you click the checkbox, the checked value changes.
+  // This can be overwritten if an `onChange` action is passed.
+  onChange(isChecked) {
+    this.set('checked', isChecked);
+  },
+
+  change(e) {
```
unicornzero: is this the ember event system capturing a browser event, like it would with `submit`?

app/initializers/custom-inflector-rules.js
```diff
@@ -0,0 +1,13 @@
+import Inflector from 'ember-inflector';
+
+export function initialize(/* application */) {
+  const inflector = Inflector.inflector;
+
+  // Tell the inflector that the plural of "advice" is "advice"
+  inflector.uncountable('user-notification-setting');
```
unicornzero: fancy

app/models/user-notification-setting.js
```diff
@@ -0,0 +1,46 @@
...
+      return USER_NOTIFICATION_SETTING_OPTIONS.mapBy('value');
+    },
+  }),
+
+  isCommentCreatedEmailOn: twoWayArrayToggle('comment_created_email'),
+  isCommentMentionCreatedEmailOn: twoWayArrayToggle('comment_mention_created_email'),
+});
+
+// allows us to get and add/subtract an item in an array.
+function twoWayArrayToggle(settingName) {
```
unicornzero: aka `emberifiedArray` 😍
Very cool.

app/templates/components/forms/user-notification-setting.hbs
```diff
@@ -0,0 +1,33 @@
...
+  onSubmit=(perform saveUserNotificationSetting)
+  as |form|
+}}
+
+  <h2 class="text-xl font-semibold mb-2">Comment notifications</h2>
+
+  <div class="p-2 bg-gray-000 rounded border mb-2">
+    <fieldset class="mb-0">
+      {{#each userNotificationSettingOptions as |option|}}
+        {{#form.label data-test-notif-option=option.value class="flex mb-sm"}}
```
unicornzero: Does this lap the checkbox and divs all in a `label` tag? Does this have downsides? (like accessibility, perhaps?)

app/components/percy-form/checkbox.js
samselikoff: yep! Components get event handler methods to let you wire up handlers to the root element:

https://api.emberjs.com/ember/release/classes/Component#handling-browser-events

In the future these are kinda going to be going away, if and when Glimmer Components (which have no root element) become the default.

app/components/percy-form.js
```diff
@@ -0,0 +1,20 @@
...
+    yield all([timeout(500), this.onSubmit()]);
```
samselikoff: nice touch! Might want to replace with

```
const FEEDBACK_TIME = Ember.testing ? 10 : 500;
```

above and use this variable here to keep tests fast

cadeParade:
yessssss

PR #1251, opened 2019-06-28
Remove slack feature flag

cadeParade:
Remove slack feature flag, it is fully released.


app/templates/organizations/organization/integrations/index.hbs
```diff
@@ -19,12 +19,6 @@
               organization=organization
             }}
           {{/each}}
-          {{#if organization.isSlackIntegrated}}
```
unicornzero: Deleting this block breaks the view when the integration is installed. Weird that the tests didn't catch it. :(

image

unicornzero: Deleting this block breaks the view when the integration is installed. Weird that the tests didn't catch it. :(



image

PR #1286, opened 2019-07-26
Introduce rejected snapshots

cadeParade:
## Description
This PR adds the rejected properties to snapshot + build model, and also allows the build to fetch snapshots in state rejected.

This PR does not add any different display based on snapshot state rejected, nor does it sort the snapshots on the build page differently yet.

## Reviewing this PR
I've added a couple of rejected snapshots to tests that should test the order of snapshots. All the tests passing and no visual diffs related to snapshot ordering means that we successfully fetch rejected snapshots and do not sort them differently.


app/models/snapshot.js
```diff
@@ -53,12 +59,18 @@ export default DS.Model.extend({
...
   // - 'approved' --> 'auto_approved_branch': Automatically approved based on branch name
+  // - 'rejected' --> 'user_rejected': User clicked "Request changes" button in current build
+  // - 'rejected' --> 'user_rejected_previously':'rejected' status has carried forward
   reviewStateReason: DS.attr(),
   isApprovedByUser: equal('reviewStateReason', SNAPSHOT_REVIEW_STATE_REASONS.USER_APPROVED),
   isApprovedByUserPreviously: equal(
     'reviewStateReason',
     SNAPSHOT_REVIEW_STATE_REASONS.USER_APPROVED_PREVIOUSLY,
   ),
+  isRejectedPreviously: equal(
```
unicornzero: Should there also be an `isUserRejected` mapping to the `user_rejected` reviewState?

tests/factories/comment-thread.js
```diff
@@ -32,5 +32,10 @@ FactoryGuy.define('comment-thread', {
     note: {
       type: NOTE_COMMENT_TYPE,
     },
+
+    autoReview: {
+      type: REVIEW_COMMENT_TYPE,
+      comments: FactoryGuy.hasMany('comment', 1, {body: 'Changes were requested'}),
```
unicornzero: Might want to make the body blank later.

unicornzero: (To match what is happening in the API)

app/models/snapshot.js
cadeParade: Yes, but it is not needed for this PR. This property also, is not used at all in this PR, it is copied from my import. I can remove it, but it's going to be added back soon.

tests/factories/comment-thread.js
cadeParade: Sure.

PR #1296, opened 2019-07-31
Only reload snapshots with diffs after creating a review.

cadeParade:
Only reload snapshots with diffs after creating a review. Snapshots with no diffs cannot have a review made on them so we don't need to reload those to get their updated state. Also, fetch build and snapshot concurrently. No need to fetch them one after the other.


cadeParade:
This is spurred by testing the "Request changes" button in canary and seeing that it takes ages for the generated comment to reappear. 

Got some more ideas brewing also...

PR #1304, opened 2019-08-05
Specifically include sideloaded models we need to refresh snapshots after a review is made

cadeParade:
~Requires this API PR https://github.com/percy/percy-api/pull/1116/files~

After a review is made, we reload the build and snapshots that could have been affected. In both cases (and snapshots in particular) we are asking for way more data than we need to refresh reviews. For snapshots, we include all comparisons, and their images with each load, so reducing this should make the review creation faster.

This PR also removes the quick snapshot state change before the request has returned and adds a loading state to the "reject" button so that the "changes requested" badge on the snapshot shows up at the same time as the generated comment.


app/services/snapshot-query.js
```diff
@@ -20,4 +20,14 @@ export default Service.extend({
...
     });
   },
+
+  getChangedSnapshotsWithComments(build) {
+    return this.get('store').query('snapshot', {
+      filter: {
+        build: build.get('id'),
+        'review-state-reason': DIFF_REVIEW_STATE_REASONS.join(','),
+      },
+      include: 'comment-threads.comments.author',
```
djones: ✨

cadeParade: :D

PR #1321, opened 2019-08-16
Upgrade ember infinity

cadeParade:
## Description
Upgrade ember-infinity in preparation for Ember 3.11 upgrade. The old version was causing lots of deprecation warnings in 3.11.
`inViewport` property in `snapshot-list-item` was clobbering the `inViewport` service, so the name of that property has been changed.


cadeParade:
AND BEYOND!!!!

PR #1323, opened 2019-08-16
Add tooltip to demo for request changes.

cadeParade:
## Description
Add a tooltip dot to demo projects for the "request changes" button


app/services/tooltips.js
```diff
@@ -6,8 +6,19 @@ import {inject as service} from '@ember/service';
 
 export const TOOLTIP_MASTER_KEY = 'percy_demo_tooltips_hidden';
 
-const GROUP_SEQUENCE = ['one', 'seven', 'three', 'eight', 'five'];
-const NO_GROUP_SEQUENCE = ['one', 'two', 'three', 'four', 'five'];
+// one --> Build overview
```
unicornzero: Yessssss

cadeParade:
There is not a snapshot capturing the visual change and the reason is that these tooltips are really a pain to test due to the asynchronous and out-of-line rendering.

PR #1329, opened 2019-08-20
Button to access archived comments.

cadeParade:
## Description
It is hard to differentiate between open and archived comments. Also users will probably not care as much about archived comments as open comments. So put archived comments behind a button.
archived-comments

## Reviewing this PR


cadeParade:
Updated to be full width.

PR #1331, opened 2019-08-20
Allow users to set their default_branch_base

unicornzero:
This allows users to set the default branch that's used as the base build for comparisons.

API PR: https://github.com/percy/percy-api/pull/1151

Here's the most relevant Percy snapshot/diff: https://percy.io/percy/percy-web/builds/2430050/view/156822905/1280?browser=chrome&mode=diff


unicornzero:
Yes, just deployed API to prod.
image

PR #1342, opened 2019-08-26
Slack-style emoji in comments.

cadeParade:
## Description
Type a `:` character in a comment or comment-reply, and get a list of emoji to enter. 
The emojis included in the feature are from v11 and backwards. v12 is not supported on many devices still.


cadeParade:
I agree, but I couldn't figure out which rule made it mad, and also the
troublesome area became every single empty line in the file, so putting it
around that is the whole file :/

cadeParade:
I feel like maybe it is a big in eslint -- adding the special characters
anywhere in a file made all empty lines in the rest of the file errors 🤔

cadeParade:
*bug

PR #1350, opened 2019-08-30
Add option to snapshot header dropdown to go back to last changed snapshot.

cadeParade:
## Description
Quietly add an option to navigate to the latest snapshot with changes.

## Reviewing this PR

_**Is Covered by Automated Tests?: Yes*_


app/components/snapshot-viewer-header.js
```diff
@@ -78,6 +79,24 @@ export default Component.extend({
 
       utils.replaceWindowLocation(url);
     },
+    async goToLastChangedSnapshot() {
+      const latestChangedAncestorRef = this.snapshot.belongsTo('latestChangedAncestor');
```
unicornzero: Is this the only place that the `latestChangedAncestorRef` is defined? Since it's a snapshot under the hood we don't need a model definition for it?

PR #1356, opened 2019-09-03
Cut percy-web production build size

cadeParade:
## Description
Our current production build size is 1331KB (1.3MB) according to the techniques outlined here. 

Running this tool revealed that `faker.js` was being included in our production builds at a whopping 261kb and it is not used in app code at all. I've removed it from production builds.

Additionally, highlight.js is very large at 152kb, and used in a rarely used component-- of which all instances atm have the "nohighlight" option. I've removed this package altogether. 

Finally, `svgs.js` is rather large, and I've taken an easy step to compress the svgs so they are half the size. 

Overall, this update cuts the production build size from ~1.3MB to ~754kb or about ~42%.

Further optimization can be made by loading the emoji file lazily, but will require a bit more code change so I will follow up with this.

Cut_percy-web_production_build_size_by_cadeParade_·_Pull_Request__1356_·_percy_percy-web

This tool only analyzes files that flow through broccoli. This includes actual ember addons and the percy-web code. It does NOT include anything included by `ember-auto-import`, which is processed through webpack. Another way to check bundle size is look at the built file sizes in `dist/assets` directory after running `ember b -environment=production`. This shows very different values, but after the changes in this PR we do see a decrease of ~1.4 MB, or about 25% of build size.

image
image

## Reviewing this PR


package.json
```diff
@@ -39,6 +39,7 @@
     "chai-as-promised": "^7.1.1",
     "ember-ajax": "5.0.0",
     "ember-attacher": "^0.15.1",
+    "ember-auto-import": "^1.5.2",
```
Robdel12: 😍 I was wondering how we got the webpack build analyzer into ember hahahaha

cadeParade: It was actually this addon: https://github.com/kaliber5/ember-cli-bundle-analyzer 
But I didn't commit it because we don't need it all the time. `ember-auto-import` is for importing "vanilla" faker.js following the advice here: https://www.ember-cli-mirage.com/docs/getting-started/upgrade-guide

But anyway, `ember-auto-import` is great and we should have it!

package.json
```diff
@@ -105,6 +105,7 @@
     "eslint-config-prettier": "6.0.0",
     "eslint-plugin-ember": "^6.7.1",
     "eslint-plugin-prettier": "3.1.0",
+    "faker": "^4.1.0",
```
unicornzero: Is this package.json change what caused it to no longer be included in the prod bundle?

cadeParade: Looking at this further, it seems like that because it is only imported from test files, it is not included in the build. Packages that are requested from app code will make it into percy-web.js

Robdel12: Yeah, it seems the ember faker package was including all of its deps into the prod builds 👎 

Since we can use ember auto import now, we can import NPM packages directly into files. And since the test files don't get bundled in a prod build, faker shouldn't make it in anymore

ember-cli-build.js
```diff
@@ -40,7 +40,9 @@ module.exports = function(defaults) {
     },
     svg: {
       paths: ['public/images/icons', 'public/images/logos', 'public/images/icons/tech'],
-      optimize: false,
+      optimize: {
+        plugins: [{removeEmptyAttrs: false}, {cleanupIDs: {minify: false}}, {mergePaths: false}],
```
Robdel12: Curious if percy helped with this by surfacing diffs?

cadeParade: Yes, definitely. With just `optimize: true`, Percy revealed that the vertical line svg between the left and right side of the build header had disappeared. Another icon that I caught manually (because of fuzz factor it didn't show up in Percy) was broken also. 
This is why I've added these three random seeming rules, because we caught the faulty svgs in Percy. (and also, it was removing some test selectors that were used in tests).

PR #1359, opened 2019-09-05
Rk/websockets comment updates

unicornzero:
This PR provides instant UI updates for comments:

* Implements websockets with Pusher
* Uses an auth endpoint to connect to private websocket channels
* Receives websocket updates for the creation of comments and archiving of comment threads, and immediately updates the ember data store to trigger updates in the UI

## Reviewing this PR


package.json
```diff
@@ -133,5 +134,9 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
+  },
```
unicornzero: This seems the wrong place. They should be up in devDependencies, yes?

samselikoff:
This is a little hacky, but you can use Mirage to generate the JSON:API payload.

At the end of your `createWebhookPayload` hook I tried this:

```js
let lastComment = server.schema.comments.find(lastExistingCommentId)
server.serializerOrRegistry.serialize(lastComment, {
  queryParams: { include: 'author,comment-thread' }
})
```

I think this produces the payload you want. I can explain what's going on:

- `lastComment` is a Mirage Model instance, so if you update the `createWebhookPayload` to actually create a new Mirage model, you could use that instead
- `server.serializerOrRegistry` is kinda private but kinda public API, and if it ever changes it would be mentioned in the changelog.
- The `serialize` method takes in a Model, and the second arg is a request, so we're passing in a request-like object with `queryParams` key simulating a request that's like `GET /comments/1?include=author,comment-thread`

I can chat more about this later but this does seem to produce the payload you're after. If you omit the request-like 2nd argument Mirage will use the default serializers that are set up and you'll get the serialized comment with ALL its relationships (which might be ok for your use case too).

mirage/factories/comment-thread.js
```diff
@@ -5,7 +5,10 @@ import faker from 'faker';
 
 export default Factory.extend({
   type: REVIEW_COMMENT_TYPE,
-  createdAt: () => moment().subtract(22, 'hours'),
+  createdAt: () =>
+    moment()
+      .subtract(22, 'hours')
+      .format(),
```
unicornzero: This stringifies the value, allowing it to work in my tests without me having to run JSON.stringify and JSON.parse on the objects.

mirage/factories/comment.js
```diff
@@ -6,5 +6,5 @@ export default Factory.extend({
   body: () => faker.hacker.phrase(),
   author: association(),
   createdAt: () => moment().subtract(22, 'hours'),
-  commentThread: association('commentThread'),
+  commentThread: association(),
```
unicornzero: When testing something else (now deleted), I discovered the incorrect syntax here.

app/routes/organization.js
```diff
@@ -41,6 +43,12 @@ export default Route.extend({
     }
   },
 
+  async _setupWebsocket(organization) {
+    if (await isUserMemberPromise(organization)) {
```
cadeParade: Don't need to await this anymore...rebase?

app/services/websocket.js
```diff
@@ -0,0 +1,45 @@
...
+import utils from 'percy-web/lib/utils';
+import Pusher from 'pusher-js';
+
+export default Service.extend({
+  store: service(),
+
+  subscribeToOrganization(organization) {
+    const channelName = `private-organization-${organization.id}`;
+
+    if (this._client.channels[channelName]) {
```
cadeParade: Is it important that the key have any certain value, or only that the key be present in the `channels` hash?

app/services/websocket.js
```diff
@@ -0,0 +1,45 @@
...
+        cluster: config.APP.PUSHER_APP_CLUSTER,
+        authEndpoint: utils.buildApiUrl('websocketsAuth'),
+        auth: {
+          headers: {'X-CSRF-Token': csrfToken},
+        },
+      });
+    },
+    set(key, value) {
+      if (config.environment === 'test') {
+        return value;
```
cadeParade: Let's put that assert back in...we don't want anyone setting this value unless they're in a test.  https://api.emberjs.com/ember/release/functions/@ember%2Fdebug/assert

app/services/websocket.js
unicornzero: My intention is to not subscribe to the same organization's channel multiple times.

app/services/websocket.js
```diff
@@ -4,28 +4,45 @@ import config from '../config/environment';
...
-    });
+  disconnect() {
+    this._socket.disconnect();
   },
 
-  subscribe(channelName, event, callback) {
-    const channel = this._client.subscribe(channelName);
+  _bindEvents(channel, event, callback) {
     channel.bind(event, callback);
+    channel.bind('error', function(error) {
```
cadeParade: What does this do?

unicornzero: If Pusher sends an error, such as if we run out of connections for our account, I'd like us to log an entry in Sentry.

cadeParade: Have we confirmed that it works?

unicornzero: Discussed in chat. I ran into a snag and am pulling this out for now.

PR #1370, opened 2019-09-10
Websockets: handle when cookie absent and don't reinitialize connections

unicornzero:
### Description
Do not attempt to reinitialize a socket connection if one already exists or if the cookie is absent.

### Reviewing this PR

_**Is Covered by Automated Tests?: Partially**_ - It was not feasible to test whether Pusher.disconnect was called or manipulate the content of the cookie in the test context.


app/services/websocket.js
```diff
@@ -26,31 +28,38 @@ export default Service.extend({
...
   _isSubscribed(channelName) {
-    return isPresent(this._socket.channel(channelName));
+    return this._socket && isPresent(this._socket.channel(channelName));
   },
 
   _subscribe(channelName, event, callback) {
     if (this._isSubscribed(channelName)) {
       return;
     }
+    if (!this._socket) return;
```
cadeParade: Probably this line should go above line 35.

unicornzero: agreed

PR #1372, opened 2019-09-12
Improved CI run performance and resilience

djones:
This PR does two main things:

1) Parallelizes our test suite into 4 test runners on Buildkite. This results in our CI run times going from 8-10mins to 4-7mins
2) Makes our CI runs more resilient by adding retries when the preemptible VMs running the buildkite agent are shutdown during the test run. This prevents tests from randomly failing in this way.


package.json
```diff
@@ -19,8 +19,7 @@
     "test": "ember test",
     "start-production-for-testing": "ember server --environment=production",
     "test:node_debug": "node debug node_modules/ember-cli/bin/ember test --server",
-    "test:parallel": "PERCY_PARALLEL_TOTAL=-1 EMBER_ENV=test ember exam --environment=test --path dist --split=4 --parallel && yarn test:finalize",
-    "test:finalize": "PERCY_PARALLEL_TOTAL=-1 yarn percy finalize --all",
```
djones: the finalize step was moved into the Buildkite pipeline. This step does not require the percy-web codebase.

.buildkite/pipeline.yml
```diff
@@ -1,12 +1,41 @@
 steps:
-  - label: ":percy::hamster: test"
+  - label: ":percy::hamster: test runner #%n"
     commands:
       - "make build-test"
       - "make test"
+    parallelism: 4
```
djones: we could dial this up more if we wanted but it will have diminishing returns because a good portion of the run time is getting docker ready each time, rather than running the tests themselves.

.buildkite/pipeline.yml
```diff
@@ -1,12 +1,41 @@
...
+    commands:
+      - "percy finalize --all"
+    plugins:
+      - docker#v3.3.0:
+          image: "percyio/agent"
+          environment:
+            - "PERCY_TOKEN"
+            - "BUILDKITE"
+            - "BUILDKITE_BUILD_ID"
+            - "BUILDKITE_PARALLEL_JOB_COUNT"
```
timhaines: Suspect this one isn't needed -- but isn't harmful either.

cadeParade:
👏 😍

PR #1374, opened 2019-09-13
Update connected accounts panel to show okta identity

cadeParade:
## Description
If a user has an Okta identity, display it on the connected accounts page, but do not allow them to add/remove it.

## Reviewing this PR

### Relevant screenshot: 
[Click here](https://percy.io/percy/percy-web/builds/2587359/view/166366042/1280?browser=chrome&mode=diff)


app/models/identity.js
```diff
@@ -9,4 +12,12 @@ export default DS.Model.extend({
 
   isGithubIdentity: equal('provider', 'github'),
   isAuth0Identity: equal('provider', 'auth0'),
```
unicornzero: Is 'aut0' for user/password identities?

tests/factories/identity.js
```diff
@@ -9,13 +9,15 @@ FactoryGuy.define('identity', {
 
   traits: {
     githubProvider: {
-      provider: () => {
-        return 'github';
-      },
+      provider: 'github',
```
unicornzero: yay for succinct syntax!

app/models/identity.js
cadeParade: yes.

tests/factories/identity.js
cadeParade: No need for a function here, since it is returning the same string every time.

PR #1358, opened 2019-09-04
Bye bye bower

cadeParade:
## Description
Bower be gone.

requires #1356 to merge

PR #1375, opened 2019-09-13
Update baseimage-web

djones:
Related PRs:

* https://github.com/percy/percy-web/pull/1358
* https://github.com/percy/baseimages/pull/31


djones:
We suspect the visual diffs that came up here are due to Chrome being updated to the latest stable. Therefore they are accurate.

PR #1381, opened 2019-09-17
Disable removing users from org if they have SAML identity

cadeParade:
## Description
Do not allow users (admin or otherwise) to remove users from the org if they have a saml integration.

## Reviewing this PR


tests/integration/components/organizations/user-card-test.js
```diff
@@ -251,6 +277,16 @@ describe('Integration: UserCard', function() {
           expect(orgUserStub).to.have.been.called;
         });
       });
+
+      describe('Leave organization', function() {
+        it('requests confirmation and calls destroyRecord', async function() {
+          await requestsConfirmationAndCallsDestroyRecord();
```
unicornzero: Is there test setup to make this work for leaving the org you are in? I'm missing it if there is.

Looks like we've stubbed `currentUser` to `adminUser` and `requestsConfirmationAndCallsDestroyRecord` is setting `organizationUser=otherAdminOrganizationUser`. So, that test will always test the behavior for removing a user from an org that is not you.

`Leave organization` applies when it's your own organization user that you're trying to delete, but I see the definition for `requestsConfirmationAndCallsDestroyRecord` is setting `organizationUser=otherAdminOrganizationUser` and therefore not testing the behavior for removing your own user.

cadeParade: Thank you for noticing this. I've updated so it makes more sense with `currentUser`.

PR #1390, opened 2019-09-24
Remove skeleton.css

PixelJanitor:
Eliminating the legacy skeleton css from our codebase


unicornzero:
Can you please revisit the plan comparison table? https://percy.io/percy/percy-web/builds/2660287/view/170793950/1280?browser=chrome&mode=head

PR #1392, opened 2019-09-24
Further Improve CI Performance

djones:
Building upon https://github.com/percy/percy-web/pull/1372 I have ditched our own baseimage, which rarely gets updated, with [CircleCI's minimal node + browsers variant image](https://hub.docker.com/r/circleci/node). This also uses the Buildkite docker plugin to run it.

Using CircleCI's image means that the browsers we test against will be more regularly updated and has also lead to further performance wins.

The last 5 run times with this configuration are:

* 7m 28s
* 6m 49s
* 6m 46s
* 6m 32s
* 6m 3s

This PR also centralizes the retry code into a single reusable block.


.buildkite/pipeline.yml
```diff
@@ -1,24 +1,45 @@
+retry: &retry
```
djones: I stole this idea from @paulfri. Thanks Paul!

.buildkite/pipeline.yml
```diff
@@ -1,24 +1,45 @@
...
+      - BUILDKITE_BUILD_ID
+      - BUILDKITE_PARALLEL_JOB
+      - BUILDKITE_PULL_REQUEST
+      - PERCY_TOKEN
+    plugins:
+      - docker#v3.3.0:
+          image: "circleci/node:8-browsers"
+          workdir: "/home/circleci/percy-web"
+          propagate-environment: true
+          user: "root"
```
djones: without this, I was getting a permission denied error. `user: "circleci"` also was not effective even with the workdir running inside `/home/circleci`. This was a bit surprising to me.

djones:
@paulfri yeah I might address that in a follow up PR. My only hesitation around that is taking away `make test`. Not sure if that's something peeps are using at all

PR #1393, opened 2019-09-25
Screenshot the new snapshot state

timhaines:


timhaines:
I figured I could rely on the “new” trait for snapshots in tests/factories/snapshots.js where it exists.  But no, the new trait needs to be added to mirage/factories/snapshot.js, which will be a bit more fiddly. Not gonna do it this week.

PR #1401, opened 2019-10-04
Optimize marketing images with <picture> and Contentful's image api

djones:
Currently we are serving large images in a single format to all devices that load our marketing site.

This is problematic because better image formats are available in some browsers and smaller images should be served to mobile devices and so what happens is we end up sending way too much data over the line in a suboptimal format.

This PR uses the `<picture>` element to support 3 different breakpoints for responsive images. It also adds support for webp image format. And finally it adds support for 3 different screen densities: 1x, 1.5x and 2x. In the future we could add 3x but our source images are too small to bother with that right now.

It is limited to images served via Contentful which is basically the images that you see on our marketing site. This implementation works by making use of [Contentful's Image API](https://www.contentful.com/developers/docs/references/images-api/#/reference/changing-formats/image-format).


app/templates/components/marketing/logo-block.hbs
```diff
@@ -3,7 +3,7 @@
     <div class="flex flex-wrap items-center justify-center">
       {{#each logos as |customerLogo|}}
         <div class="w-1/2 sm:w-1/3 md:w-1/4 lg:w-1/5 text-center mb-2 px-4">
-          <img src="{{customerLogo.logo.file.url}}" alt="{{customerLogo.logo.description}}">
+          <img src="{{customerLogo.logo.file.url}}" alt="{{customerLogo.logo.description}}" />
```
cadeParade: Why not convert this one?

djones: ah thanks! I didn't do this one because they're SVGs, but I later built in automatic detection to leave SVGs alone so we can safely opt this one in now. Update coming soon.

djones: Fixed with e31a1e4c4370aff6bdaedb86a1883aa3e5e76428

djones:
I'm a bit concerned that the resized images served in the test suite are not deterministic. If that's the case we can work around it at the time.

PR #1403, opened 2019-10-05
Tailwind overhaul

PixelJanitor:
Converted as many css attrs to tailwind usages


.stylelintrc
```diff
@@ -1,24 +1,27 @@
...
   ],
-  "ignoreFiles": ["./app/styles/vendor/*.sass"],
+  "extends": "stylelint-config-recommended",
   "rules": {
     "at-rule-blacklist": ["extend"],
     "at-rule-name-case": "lower",
     "at-rule-name-space-after": "always-single-line",
+    "at-rule-no-unknown": [ true, {
+      "ignoreAtRules": [
+        "apply",
```
djones: I'm curious how you got around the use of `theme` for linting in the end

PixelJanitor: For some reason it just never came into play. I never had the IDE yell at me about it but I'm using a PostCSS file type extension too so it knew what it was as a function (a lot of postcss/sass plugins have `foo()` functions too). I'm assuming stylelint doesn't look for it to know it's not valid css.

PR #1404, opened 2019-10-08
feat: Add changelog show page

Robdel12:
## Description

This adds the ability to link directly to changelog entries. This is useful for marketing / SEO as well as linking directly to updates for users in support. 

## Pictures

image


app/routes/changelog/show.js
```diff
@@ -0,0 +1,27 @@
+import Route from '@ember/routing/route';
+import metaTagLookup from 'percy-web/lib/meta-tags';
+
+export default Route.extend({
+  headTags: metaTagLookup('changelog'),
+
+  async model({id}) {
+    const hero = await this.store.queryRecord('heroBlock', {
```
cadeParade: You can query for a single record by id. See here: https://github.com/davidpett/ember-data-contentful#usage
AFAICT, there's no reason to fetch every changelog post for a show page?

cadeParade: Why are we waiting for the hero block before asking for the posts? Any reason not to use RSVP.hash to make those requests in parallel?

cadeParade:
Two more questions:
- Do we care that the url will have the ugly contentful id? ex: http://percy.io/changelog/vcfhm2WxTAsKASVHEFTr8?
- Should we add a "back" link on the show page to get back to the main changelog page?

PR #1405, opened 2019-10-10
Surface removed snapshots

cadeParade:
## Description
This adds a new component, `removed-snapshots`, which lists the names of snapshots that were in the base build that are no longer present in the current build. 
Each snapshot name links to the fullscreen version of the snapshot from the base build. 
If there are over 5 (arbitrary choice) snapshots missing from the build, the box will only show the first 5 (in the order returned from the API) and show an option to expand the whole list of snapshots. A user can collapse the snapshot list once it is expanded.

The request for missing snapshots occurs once the build header is loaded. This means the page rendering is not blocked on the request, but it does show up slightly later than some of the page. Since the build page requests are already quite bloated, I didn't want to block anything on this request.

There is a (?) icon next to the title of the box that explains why a user might see this box:
Fullscreen_10_10_19__12_43_PM

The box is behind a feature flag `show-removed-snapshots`. The API request for removed snapshots will not be made if the flag is off.

## Pictures
The box looks like this: https://percy.io/percy/percy-web/builds/2793412/view/178681984/1280?browser=chrome&mode=diff

## Reviewing this PR


app/templates/components/build-container.hbs
```diff
@@ -18,6 +18,10 @@
     isBuildApprovable=isBuildApprovable
   }}
 
+  {{#if (and (variation 'show-removed-snapshots') build.isFinished)}}
+    <RemovedSnapshots @build={{build}} />
```
Robdel12: 👏

mirage/serializers/build.js
```diff
@@ -1,7 +1,7 @@
 import {JSONAPISerializer} from 'ember-cli-mirage';
 
 export default JSONAPISerializer.extend({
-  include: ['project', 'repo', 'approvedBy', 'commit', 'baseBuild', 'browsers'],
+  include: Object.freeze(['project', 'repo', 'approvedBy', 'commit', 'baseBuild', 'browsers']),
```
Robdel12: Good call

app/templates/components/build-container.hbs
djones: Hello angle brackets 👍

PR #1416, opened 2019-10-14
Disable invites on forced sso orgs

cadeParade:
## Description
If an org has a saml integration with force-sso = true, disable invite button on org users page and do not allow opening the invite form.

## Reviewing this PR


tests/factories/saml-integration.js
```diff
@@ -0,0 +1,20 @@
+import FactoryGuy from 'ember-data-factory-guy';
+import faker from 'faker';
+
+FactoryGuy.define('saml-integration', {
+  default: {
+    organization: FactoryGuy.belongsTo('organization'),
+    forceSso: false,
+    provider: () => 'okta',
+    providerLoginUrl() {
+      return `${faker.internet.domainWord()}.${faker.internet.domainName()}`;
```
unicornzero: Can we do `faker.internet.url()` here? https://github.com/Marak/faker.js/wiki/Internet

cadeParade: Yes. This format was a relic from a previous iteration of loginUrl.

cadeParade: Well, actually, I think I did it that way originally because that property wasn't going to be a full url. Is it a full url now?

cadeParade: Happy to change in a followup

unicornzero: I thought we had intended to put the okta login url there (though that won't do much good since we don't support IdP-Initiated SSO). I don't know what partial URL would go there.

cadeParade: Whenever I made this, the url  would be something like `percylindsay.okta.com` and then you could click that and log into okta there..i suppose that could be a full URL..I don't really recall why I didn't have it that way...but it seems like there are different plans for that field now?

PR #1420, opened 2019-10-16
Remove unused attributes

djones:
Avoid deserializing and storing these attributes in memory when we don't use them.

Related PR: https://github.com/percy/percy-api/pull/1224


app/models/organization.js
```diff
@@ -26,6 +26,7 @@ export default DS.Model.extend({
   slackIntegrations: DS.hasMany('slackIntegrations', {async: false}),
   versionControlIntegrations: DS.hasMany('version-control-integrations', {async: false}),
   samlIntegration: DS.belongsTo('samlIntegration', {async: false}),
+  // Todo: invites is actually never sent by the API, but will fail tests if you remove it
```
cadeParade: I don't think this comment is necessary. `invites` is an async relationship, so they don't need to be included in the organization request, but ember _does_ need to know about the relationship, which this line defines.

djones: Fixed with 30a53cf43cf873f37f770ef21fe6b550af765b2a

PR #1423, opened 2019-10-16
Do not allow removal of users if org has forced SSO SAML integration

cadeParade:


app/components/organizations/user-card.js
```diff
@@ -44,13 +46,15 @@ export default Component.extend({
   }),
 
   isRemoveDisabled: computed('isViewerAdmin', 'userHasSamlIdentity', function() {
-    return this.userHasSamlIdentity || !this.isViewerAdmin;
+    const isOrgForceSso =
+      this.userHasSamlIdentity && this.samlIntegration && this.samlIntegration.forceSso;
```
unicornzero: I'm not sure I follow why we have the `this.userHasSamlIdentity`. Is your thinking here that you want a forced sso org to be able to remove a user that doesn't have any saml identities?

cadeParade: Updated

PR #1424, opened 2019-10-16
Accommodate partial build flag

cadeParade:
## Description
With https://github.com/percy/percy-api/pull/1229, we have a new property on `build` object -- `partial`. When a build is partial, it will mean that a user has set that flag from their sdk and expects to run only a portion of their snapshots for that build.

When this is the case, there will be expected missing snapshots, so we will not make the missing snapshots API request, nor show the missing snapshots box. Additionally, there's a banner at the top of the build stating that the build is a partial build.

## Reviewing this PR


app/templates/components/build-container.hbs
```diff
@@ -18,7 +18,13 @@
...
   }}
 
-  {{#if (and (variation 'show-removed-snapshots') build.isFinished)}}
+  {{#if (and build.isFinished build.partial)}}
+    <div class="public-project-notice p-2 border border-gray-100 flex items-center font-semibold">
+      Partial build
+    </div>
+  {{/if}}
+
+  {{#if (and (variation 'show-removed-snapshots') build.isFinished (not build.partial))}}
```
timhaines: Does this prevent the request for removed snapshots from being sent to the server too?

cadeParade: Yep. The request is kicked off from within that component, so if it never renders, it never calls the API.

PR #1426, opened 2019-10-17
let admins always remove users

unicornzero:


tests/integration/components/organizations/user-card-test.js
```diff
@@ -314,55 +271,6 @@ describe('Integration: UserCard', function() {
...
-        let transitionToStub;
-
-        beforeEach(function() {
-          transitionToStub = sinon.stub();
-          stubService(this, 'router', 'router', {
-            transitionTo: transitionToStub,
-          });
-        });
-
-        it('requests confirmation and calls destroyRecord', async function() {
```
cadeParade: This test still seems relevant?

unicornzero:
I've been thinking about edge cases a lot and will probably want to show extra messaging if someone is removing a user from an SSO org. So, I'm going to hold off on merging this today as some of the code and test setup may carry over.

PR #1436, opened 2019-10-22
Demo-tooltip updates

cadeParade:
## Description
When I was debugging something recently, I got annoyed by all the tooltip elements being on the build page regardless of whether they would display on the current build (aka, I was not looking at a demo build but the tooltip elements were still there cluttering up the DOM). So I wanted to remove them when they weren't needed.
This led me to discover that tooltips were slightly broken for build #1 in demo projects -- it wanted to display the snapshot approval button tooltip, but since it is the first build and it's on master, they are all auto-approved, hence no approval button which broke the flow. 
This led me to centralize all the tooltip messages and also give them more descriptive keys. The number keys were all out of order and very difficult to reason about.

## Reviewing this PR

_**Is Covered by Automated Tests?: YES-ish**_


app/services/tooltips.js
```diff
@@ -5,21 +5,32 @@ import {computed} from '@ember/object';
...
-// six --> ??
-// seven --> Snapshot group
-// eight --> Group approval
-// nine --> Snapshot request changes button
-// ten --> Group request changes button
-
-const GROUP_SEQUENCE = ['one', 'seven', 'three', 'ten', 'eight', 'five'];
-const NO_GROUP_SEQUENCE = ['one', 'two', 'three', 'nine', 'four', 'five'];
-const NO_DIFF_SEQUENCE = ['one', 'two', 'four', 'five'];
+const GROUP_SEQUENCE = [
```
djones: nice

app/components/demo-tooltip.js
```diff
@@ -3,18 +3,78 @@ import {computed} from '@ember/object';
 import {not, or, readOnly} from '@ember/object/computed';
 import {inject as service} from '@ember/service';
 
+/* eslint-disable max-len */
+const TOOLTIP_DATA = {
```
djones: nice to have this content centralized :+1:

PR #1445, opened 2019-10-29
[2/3] Decouple org relationships

cadeParade:
This PR depends on https://github.com/percy/percy-api/pull/1254
and should be merged very close together with https://github.com/percy/percy-api/pull/1257

## Description
This PR finds all the places where things side-loaded on organization are required, and only asks for them when required.
This results in many more requests to `organizations/:id` endpoint, but each request is much smaller, and will result, in most cases, that less data overall will be loaded into the app. For example, while the build page does require the organization to be present in the app, the build page does not require: org subscription, version control integrations, slack integrations, or github integration requests. Most folks look only at the build page when using percy, so all of this stuff is not usually needed and now, not loaded.

One thing of note is that I've added the addon [`ember-data-storefront`](https://embermap.github.io/ember-data-storefront/docs). This addon does many things but the features that this PR takes advantage of are:
- using `loadRecord` rather than `findRecord`. Using `loadRecord` will always run the query the first time, then cache that query-with-includes combination. If the query was run before, the query promise will always immediately resolve and the fetch will happen again in the background. This means that after the request has been made once, it is never again blocking.
- using `organization.sideload`. You can use `sideload` when you already have the object that you need to reload. This is basically syntax sugar that allows replacing:
    ```
     this.store.findRecord('organization', organization.id, {
        include: 'some.stuff',
        reload: true,
    });
    ```  
    with
    ```
    organization.sideload('some.stuff');
    ```
    Requests made with `sideload` are also cached and provide the same benefits outlined above.

## Reviewing this PR

Thank you testing, you caught many things.


app/components/notice-bar.js
```diff
@@ -3,20 +3,54 @@ import isUserMember from 'percy-web/lib/is-user-member-of-org';
...
+  currentUsagePercentage: readOnly('subscription.currentUsagePercentage'),
+  hasRemainingSnapshots: bool('subscription.hasIncludedSnapshotsRemaining'),
+  trialDaysRemaining: readOnly('subscription.trialDaysRemaining'),
   isTrialEndingToday: equal('trialDaysRemaining', 0),
 
   showNoticeBar: computed('organization', 'session.currentUser', function() {
     return isUserMember(this.session.currentUser, this.organization);
   }),
+
+  fetchOrgWithSubscription: task(function*(organization) {
```
cadeParade: This component is displayed on many pages, so it is difficult to depend on the route for the subscription information. So now this component fetches the subscription information for itself.

app/routes/organizations/organization.js
```diff
@@ -8,6 +8,12 @@ export default Route.extend(AuthenticatedRouteMixin, {
   session: service(),
   currentUser: alias('session.currentUser'),
 
+  model(params) {
+    return this.store.loadRecord('organization', params.organization_id, {
```
cadeParade: The `organizations/organization` routes contain all the org settings route of which many depend on subscription data, so load that here.

app/templates/components/build-toolbar.hbs
```diff
@@ -17,8 +17,10 @@
         {{project.name}}
       {{/link-to}}
     </div>
-    {{#if organization.isSponsored}}
-      {{fa-icon data-test-build-public-project "globe" class="text-blue-500 ml-sm"}}
+    {{#if isBuildApprovable}}
+      {{#if organization.isSponsored}}
```
cadeParade: This always caused problems locally because `organization.isSponsored` requires knowing about whether an organization's plan is sponsored. Knowing whether a plan is sponsored requires being allowed to get the plan data. When you are not logged in and viewing a public org, the API will not return subscription or plan data. So this line would throw an error. 
This update checks for the plan if the user is a member of the org they're viewing. This means that anonymous users will not see the sponsored icon.

app/models/organization.js
```diff
@@ -17,7 +18,7 @@ const DISPLAY_NAMES = {
   gitlabSelfHosted: 'GitLab Self-Managed',
 };
 
-export default DS.Model.extend({
+export default DS.Model.extend(LoadableModel, {
```
unicornzero: What does this do?

app/routes/organization/project/integrations/index.js
```diff
@@ -9,10 +9,13 @@ export default Route.extend(AuthenticatedRouteMixin, {
   intercom: service(),
 
   model() {
+    const organization = this.store.loadRecord('organization', this.modelFor('organization').id, {
```
unicornzero: I assume we make an api call when using `this.store.loadRecord` with the include; are we also making one with `this.modelFor('organization')`, and if so, is that just how it works or is there a way around it?

app/routes/organizations/organization/billing.js
```diff
@@ -17,26 +17,22 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
-            : null,
-        };
-      });
+    return organization.sideload(includes.join(',')).then(organization => {
+      // If you want to access more relationships that belong to the
+      // organization in this route, you must set them in setupController
+      // or, for some reason, the relationship will be overwritten or dropped
+      return {
+        organization,
+        usageStats: shouldIncludeUsageStats ? organization.subscription.currentUsageStats : null,
```
unicornzero: Cleaner! Nice!

app/routes/organizations/organization/integrations/index.js
```diff
@@ -5,15 +5,15 @@ import {INTEGRATION_TYPES} from 'percy-web/lib/integration-types';
...
 export default Route.extend(AuthenticatedRouteMixin, {
   model() {
-    return this.modelFor('organizations.organization');
+    return this.modelFor('organizations.organization.integrations');
   },
 
   setupController(controller, model) {
     controller.setProperties({
       integrationItems: INTEGRATION_TYPES,
       organization: model,
```
unicornzero: It looks like model is now `integrations` and not `organization`

mirage/serializers/organization.js
```diff
@@ -1,25 +1,6 @@
 import {JSONAPISerializer} from 'ember-cli-mirage';
 
 export default JSONAPISerializer.extend({
-  include: function() {
```
unicornzero: 🔥🔥🔥

app/components/organizations/github-integrator.js
```diff
@@ -15,20 +15,22 @@ export default Component.extend(PollingMixin, {
...
+        let githubIntegration = organization.githubIntegration;
+        let githubIntegrationRequest = organization.githubIntegrationRequest;
 
-      // If the has been fully installed or uninstalled, stop polling for updates.
-      if (githubIntegration || (!githubIntegration && !githubIntegrationRequest)) {
-        this.pollingTask.cancel();
-      }
-    });
+        // If the has been fully installed or uninstalled, stop polling for updates.
+        if (githubIntegration || (!githubIntegration && !githubIntegrationRequest)) {
```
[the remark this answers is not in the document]
cadeParade: It has been a long time since I loaded github integration installation logic into my head, but the comment says its testing for full installation (presence of `githubIntegration`) or full uninstallation (absence of both). So your proposed `(githubIntegration || !githubIntegrationRequest)` does not quite do the same thing, as it would not proceed if they were both false.

app/components/setup-github-app.js
```diff
@@ -9,27 +9,34 @@ export default Component.extend(PollingMixin, {
...
-        let installationId = this.installationId;
-        let organization = orgs.find(org => {
-          return areInstallationIdsEqual(
-            org.get('githubIntegration.githubInstallationId'),
-            installationId,
-          );
-        });
+    const orgUsers = this.session.currentUser.organizationUsers;
+    const orgPromises = orgUsers.map(orgUser => {
+      const orgId = orgUser.belongsTo('organization').id();
```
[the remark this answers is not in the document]
cadeParade: This is the belongsTo reference api we discussed with Sam a long time ago. What this does is it tries to find the `organization.id` explicitly without fetching the organization. If the organization is not present, it returns null. This is a defensive way of getting an id of something...it never causes a request to be made, and it safely returns null if it doesn't know the answer. 
It is also possible for ember to know the id of an object accessed in this way without that object being actually loaded into the store. This would be through the linkage data returned with some API requests.

So the answer is, I used `belongsTo` here to not make the assumption that the org was already loaded, since all we know is that we have organization users. If I were being truly defensive, I would put `this.store.loadRecord` inside a check to make sure `orgId` exists.  So I'll probably make that correction.

app/models/organization.js
cadeParade: This is from `ember-data-storefront` and is how I can use  `organization.sideload`.

app/routes/organization/project/integrations/index.js
cadeParade: Yes, `this.store.loadRecord` makes an api call.

`this.modelFor` does _not_ make an API call. It should be used when you are in a child route and want to know what the model for a parent route resolved to. Since you are in a child route, the parent route's model hook has already been resolved, so it does not make another API request. If you try to request `modelFor` a route that is not part of the child's route heirarchy, it will return nothing.

app/routes/organizations/organization/integrations/index.js
cadeParade: Hmm, yes, you're correct. I changed it to `organizations.organization.integrations` on purpose, because that route now loads all the integrations sideloaded onto org. So it still returns an org but implies the integrations are sideloaded. I thought it made more sense to get the model from `organizations.organization.integrations` for `organizations.organization.integrations.index` because they are closer, but in reality it doesn't really matter which one it draws from because they are all children of each other and everything comes from the organization endpoint (so will return an organization).
¯\_(ツ)_/¯

app/components/organizations/github-integrator.js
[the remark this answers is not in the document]
unicornzero: So, if `githubIntegration` is present, it will return true before getting to the `||` statement, and if it's false we move onto `(!githubIntegration && !githubIntegrationRequest)`

In `(!githubIntegration && !githubIntegrationRequest)` if `githubIntegration` is true, it returns false, but we wouldn't have gotten to it anyway, and it it returns false, we move onto `!githubIntegrationRequest`

So, I think `(githubIntegration || (!githubIntegration && !githubIntegrationRequest))` and `(githubIntegration || !githubIntegrationRequest)` are the same.

app/routes/organizations/organization/integrations/index.js
unicornzero: So, is `organization: model,` setting organization equal to integrations? Sorry, we're talking in the call and I'm not sure.

cadeParade: I don't really understand your `organization: model` syntax, but the `organizations.organization.integrations` model returns an organization (see [here](https://github.com/percy/percy-web/blob/d5b0c06f69e33c9a2dee90fee2dfb825681aee2c/app/routes/organizations/organization/integrations.js#L13).
`organizations.organization` also returns an organization (see [here](https://github.com/percy/percy-web/blob/d5b0c06f69e33c9a2dee90fee2dfb825681aee2c/app/routes/organizations/organization.js#L12)
They will always be the same organization. There is not a functional difference between returning those, just the name of the route.

app/components/setup-github-app.js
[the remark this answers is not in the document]
unicornzero: Oh! That sounds very good.

app/routes/organizations/organization/integrations/index.js
unicornzero: I mean on line 14 when we set `organization` to equal `model`, if model is now `integrations`  (per line 8) are we incorrectly setting `organization` to equal `integrations`, or is ember cleverly knowing that we want the organization set to the organization associated with line 8's integrations?

cadeParade: Model is not `integrations`. The model from `integrations` route [returns an organization](https://github.com/percy/percy-web/blob/d5b0c06f69e33c9a2dee90fee2dfb825681aee2c/app/routes/organizations/organization/integrations.js#L13) that happens to sideload some integrations. Integration models are not really being passed around or acknowledged by ember, just sideloaded in the json response and silently entered into the store.

PR #1454, opened 2019-11-06
Actual date tooltips on build page

PixelJanitor:


PixelJanitor:
Real date tooltip:
image
image

Full date in build meta info instead of time_ago:
image

cadeParade:
Can we add a timestamp also? time of day... hmm what timezone...utc i guess :/

PixelJanitor:
Updated with time info - 
image

And in dropdown:
image

PR #1455, opened 2019-11-06
Sticky fullscreen header

PixelJanitor:


PixelJanitor:
Full screen snapshot header currently scrolls with the page unlike the other snapshot headers. This PR allows the snapshot header in full screen mode to stay at the top of the screen while scrolling.

PR #1456, opened 2019-11-06
Diff overlay adjustments

PixelJanitor:


PixelJanitor:
This PR aims to help users realize they can click on div overlays, as well as a slight blur on non-hover to help discover the changed areas even more. The blur goes away on hover.

Once this merges it will live on canary for a little bit for some of us to check out.

image

cadeParade:
cool!!

PR #1457, opened 2019-11-06
only allow one identity per saml integration per user

unicornzero:
If a Percy user is already connected to one user in a company Okta instance, we should not allow them to login to the same Percy account with a second Okta user from the same integration. (This can happen if the organization has shared pairing workstations another user might have logged into, or if they had a shared login account before setting up the Okta integration).

Screen Shot 2019-11-06 at 12 29 52 PM

API PR: https://github.com/percy/percy-api/pull/1281


tests/acceptance/project-test.js
```diff
@@ -313,8 +313,6 @@ describe('Acceptance: Project', function() {
       });
 
       it('transitions to webhook config form', async function() {
-        window.crypto.getRandomValues = sinon.fake.returns(new Uint8Array(32));
```
unicornzero: This was causing a slack test to fail in CI and is not necessary for the tests to pass.

If we do need to fake this method, we could instead do this, but I'm not sure that it's necessary:
```
        var fake = sinon.fake.returns(new Uint8Array(32));
        sinon.replace(window.crypto, 'getRandomValues', fake);
        sinon.restore();
```

tests/acceptance/project-test.js
```diff
@@ -313,7 +313,8 @@ describe('Acceptance: Project', function() {
...
+        var fake = sinon.fake.returns(new Uint8Array(32));
+        sinon.replace(window.crypto, 'getRandomValues', fake);
```
unicornzero: This was causing a slack test to fail.

tests/acceptance/second-saml-identity-error-test.js
```diff
@@ -0,0 +1,49 @@
...
+  let supportMessage;
+
+  setupSession(function(server) {
+    this.loginUser = false;
+    const organization = server.create('organization');
+    const samlIntegration = server.create('samlIntegration', {organization});
+    organizationName = organization.name;
+    connectionName = samlIntegration.connectionName;
+    newSsoProfileEmail = faker.internet.email();
+    userEmail = faker.internet.email();
```
cadeParade: Does this produce consistent emails for percy?

unicornzero: yep :)

PR #1465, opened 2019-11-12
Width switcher v2

PixelJanitor:


app/templates/components/snapshot-viewer-header.hbs
```diff
@@ -165,16 +165,17 @@
...
-          {{collaboration/toggle-button
-            toggleCollaborationPanel=toggleCollaborationPanel
-            isResolved=false
-            unresolvedCommentThreadCount=unresolvedCommentThreadCount
-          }}
-        {{/if}}
-
-
         <div class="btn-toolbar" data-test-snapshot-approval-button>
+          {{#if snapshot.build.isFinished}}
```
cadeParade: Why this line? It should have this filter much further up, in `build-container.hbs` I think.

app/components/snapshot-viewer-full.js
```diff
@@ -9,7 +9,7 @@ const KEYS = {
   RIGHT_ARROW: 39,
 };
 
-const GALLERY_MAP = ['base', 'diff', 'head'];
+const GALLERY_MAP = ['diff', 'base', 'head'];
```
cadeParade: Why change the order here?

app/templates/components/snapshot-viewer-header.hbs
PixelJanitor: Good catch. I was working from two branches and bringing these small changes in from a larger PR. Some wires must have gotten crossed. Thanks

app/components/snapshot-viewer-full.js
PixelJanitor: Mike and I have talked about this for a while now. When in full screen mode and switching from base and head you currently have to press right to go to the first item and then left to go back to the last item. This order makes more sense. You are initially presented with the dif, and then a right arrow will take you to the base, and then right again shows you the head. From there its left to go back and right to go forwards. Finally making this fix with this UI update!

PR #1474, opened 2019-11-20
Dark build page

PixelJanitor:


app/templates/components/theme-svg.hbs
```diff
@@ -0,0 +1,5 @@
+{{#if isDarkMode}}
```
samselikoff: No need for the extra `isDarkMode` computed property on the theme-svg component, I would just reference the `theme` service directly here:

```hbs
{{#if this.theme.isDarkMode}}
  {{inline-svg (concat svg "-dark-mode")}}
{{else}}
  {{inline-svg svg}}
{{/if}}
```

tests/helpers/percy-snapshot-dark-mode.js
```diff
@@ -0,0 +1,25 @@
+import {percySnapshot} from 'ember-percy';
+
+const DARK_MODE_CLASS_NAME = 'theme-dark';
+
+export async function percySnapshotWithDarkMode(rawName, options) {
```
cadeParade: @djones Can we up the timeout for tests that use `percySnapshotWithDarkMode` rather than globally? 
Here is an example of how you can do that with `this.timeout`: https://mochajs.org/#test-level

tests/test-helper.js
```diff
@@ -7,7 +7,7 @@ import {setApplication} from '@ember/test-helpers';
 import start from 'ember-exam/test-support/start';
 
 mocha.setup({
-  timeout: 10000,
+  timeout: 20000,
```
cadeParade: Not a big fan of this. If a test fails badly, it will cause the test run to take 30 min to fail, as you were seeing earlier this week. Would rather not need to wait that long if we can help it.

cadeParade:
@djones We also should probably put a flag around dark mode -- could you add that in the dark-mode component?

tests/test-helper.js
djones: yep, I think we're safe to revert this now knowing more about what was happening with the test runs.

app/templates/components/build-overview.hbs
```diff
@@ -19,7 +19,7 @@
       {{else if build.isRejected}}
         {{pluralize build.totalSnapshotsRequestingChanges "requested change"}}
       {{else}}
-        {{pluralize build.totalSnapshotsUnreviewed "unreviewed change"}}
+        <span class="font-semibold">{{build.totalSnapshotsUnreviewed}}</span> {{pluralize build.totalSnapshotsUnreviewed "unreviewed change" without-count=true}}
```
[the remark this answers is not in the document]
PixelJanitor: The text styling of the words next to the number have a different font weight styling so it's wrapped in a span. There isn't another number because of the `without-count=true`. It just uses the pluralized words based on the number

app/templates/components/snapshot-group.hbs
```diff
@@ -3,22 +3,11 @@
...
 }}
   <header
-    style="grid-template-columns: 1fr auto;"
-    class="SnapshotViewer-header sticky grid gap-4 px-2 py-1 bg-gray-000 border-b border-t"
+    class="SnapshotViewer-header sticky grid gap-4 px-2 py-1 bg-secondary border-b border-t"
     {{action "expandBlock"}}
     data-test-snapshot-group-header>
     <div style="min-width: 0;" class="flex items-center">
       <div class="Viewer-header-title SnapshotViewer-title flex items-center overflow-hidden cursor-default font-bold" title="{{snapshots.length}} matching changes" data-test-snapshot-group-name>
-        <figure class="svg-container mr-1">
```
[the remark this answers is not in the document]
PixelJanitor: We removed snapshot and group header icons based on feedback that there are too many of them on the page. This buys us some more horizontal space for titles in the headers as well.

app/templates/components/snapshot-viewer-header.hbs
```diff
@@ -12,16 +12,6 @@
     {{/if}}
 
     <div class="flex items-center overflow-hidden">
-      <figure class="snapshot-icon svg-container mr-1" data-test-SnapshotViewer-comparisonIcon>
```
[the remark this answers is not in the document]
PixelJanitor: Yes - see previous comment above

tests/test-helper.js
PixelJanitor: Done

PR #1481, opened 2019-11-26
Sort snapshots on `totalOpenComments`

cadeParade:
Intermediate step in comment count cache project.
Put `totalOpenComments` on snapshot model and make it a computed property calculated from the `commentThreads` and `comments` which are currently synchronously loaded.

The only effect that this could have is that we are currently sorting on number of comment threads, and this update sorts on number of comments. 
In most cases there should be no difference between these sorts, and if they are different, it probably doesn't matter much.


cadeParade:
I don't actually know what advantage it will serve to change this earlier than we get the prop from the API.

PR #1489, opened 2019-12-05
Sort snapshots on totalOpenComments field

cadeParade:
## Description
This PR accepts the new `totalOpenComments` count on snapshots and uses it for the snapshot sort algorithm.

The build page fetches comment threads in its model hook, which means that the page is blocked on the comments fetch, but I think it's a nice stepping stone to where we might want to go in the future -- that is to fetch comment threads without blocking any rendering.

## Reviewing this PR


app/lib/snapshot-sort.js
```diff
@@ -23,13 +23,11 @@ export default function(snapshots, browser) {
     } else if (bHasDiffs && aHasNoDiffs) {
       return 1;
     }
-    // Next prioritize snapshots with more comment-threads
-    const aCommentThreadsCount = get(a, 'commentThreads.length') || 0;
-    const bCommentThreadsCount = get(b, 'commentThreads.length') || 0;
-    if (aCommentThreadsCount || bCommentThreadsCount) {
-      return aCommentThreadsCount > bCommentThreadsCount ? -1 : 1;
+    // Next, prioritize snapshots with more open comments
+    if (a.totalOpenComments || b.totalOpenComments) {
```
djones: nice pretty clear change here. No more looking at the comment threads association!

mirage/serializers/snapshot.js
```diff
@@ -12,7 +12,6 @@ export default JSONAPISerializer.extend({
     'comparisons.baseSnapshot',
     'comparisons.headSnapshot',
     'comparisons.diffImage',
-    'commentThreads.comments',
```
djones: 🎉

PR #1492, opened 2019-12-05
Move noBuilds query param to controller

cadeParade:
## Description
Well now I've done it. I have created THE SECOND CONTROLLER in our app.
The time of no controllers is at an end.
Controllers 4 lyfe.

## Reviewing this PR


app/routes/organization/project/index.js
```diff
@@ -52,36 +43,14 @@ export default Route.extend(ResetScrollMixin, {
       projects: model.projects,
       infinityBuilds: model.infinityBuilds,
       isUserMember: model.isUserMemberOfOrg,
+      noBuilds: model.infinityBuilds.length === 0,
```
samselikoff: Can `model.infinityBuilds.length` change while the user is inside the `organization.project` route?

(IOW, if `model.infinityBuilds` changes the controller property will become stale... but if it's not a problem then it's fine.)

PR #1506, opened 2019-12-18
deps: Upgrade Ember SDK to v2

Robdel12:
## What is this?

This upgrades the Ember SDK to the v2 beta release. Everything should be stable, but I want to let it ride out for the rest of 2019 before cutting a v2 stable release. 

The FF snapshots with the missing images will be fixed once we upgrade Percy's FF version (need webp support)

See https://github.com/percy/ember-percy/pull/163 for the SDK PR.


Robdel12:
shipping time

PR #1526, opened 2020-01-16
Add ember template lint

cadeParade:
## Description
Template linting rules help keep our templates more consistent and also enforce better html practices.

Additionally, a new feature in Ember Octane is using `this` in templates to indicate when a property used in the template is defined in the component's js file rather than passed in. There is a codemod to convert all templates to using `this` and the linter will help us keep it consistent going forward.

I have left all the default rules from the linter on and corrected our templates with the exception of `no-inline-styles` -- I defer to @PixelJanitor on this -- he prefers to be able to occasionally use inline styles when they are minor and one-off and I don't really have a problem with this.

There are also 3 files where a rule is skipped: `no-nested-interactive` -- This rule forbids an interactive element within another interactive element for accessibility reasons. Fixing these three instances would require a bit of product discussion since we would lose some functionality. So I've skipped these files for now so going forward we can enforce the rule but not get too involved right now.


app/templates/components/collaboration/new-thread.hbs
```diff
@@ -65,7 +65,7 @@
     </div>
   </div>
 {{else}}
-  <div class="flex items-center mb-3 cursor-pointer" {{action "toggleShouldShowNewCommentInput"}} data-test-new-comment>
+  <div class="flex items-center mb-3 cursor-pointer" {{action "toggleShouldShowNewCommentInput"}} role="button"data-test-new-comment>
```
djones:
```suggestion
  <div class="flex items-center mb-3 cursor-pointer" {{action "toggleShouldShowNewCommentInput"}} role="button" data-test-new-comment>
```

PR #1527, opened 2020-01-16
Restrict project settings editing to admins via feature flag

unicornzero:
[Clubhouse ticket](https://app.clubhouse.io/percy/story/3061/make-browser-upgrades-admin-only-for-canva)

## Description
Some organizations would like to ensure only admin users can edit project settings. This prevents users who are less familiar with how the settings affect their CI setup from unintentionally causing disruptive changes.

For now, this feature will only be available via a feature flag but we may make it a configurable setting in the UI in the future.

When the feature flag is on, users without admin rights will be unable to see or edit the contents of the project settings or project integrations pages.

## Reviewing this PR


app/templates/components/forms/slack-integration-config.hbs
```diff
@@ -32,7 +32,7 @@
         <span class="label-body font-semibold">
           Select a Percy project
         </span>
-        <span class="repo-selector flex justify-start flex-1 mb-2">
+        <span class="slack-selector flex justify-start flex-1 mb-2">
```
unicornzero: I stumbled across this when making the other changes and want to ensure this class is named correctly.

tests/acceptance/login-test.js
```diff
@@ -11,7 +11,7 @@ describe('Acceptance: Login', function() {
     this.server = server;
   });
 
-  it('should login and logout user', async function() {
+  it('should login user', async function() {
```
unicornzero: I'm updating the name of this test because it does not actually verify that a user can logout.

tests/acceptance/project-test.js
```diff
@@ -134,7 +134,7 @@ describe('Acceptance: Project', function() {
     let repos;
     let webhookConfig;
 
-    setupSession(function(server) {
+    beforeEach(function() {
```
unicornzero: We should use `setupSession` when we're setting `this.loginUser` and stick with plain `beforeEach` blocks otherwise to avoid making it more complicated for user-specific nested tests to be setup correctly.

tests/helpers/setup-acceptance.js
```diff
@@ -32,30 +32,43 @@ export default function setupAcceptance({authenticate = true} = {}) {
...
 // Example:
-//  setupSession(function() {
-//    let user = server.create('user', {name: 'Test user', id: 'test_user'});
-//    let organization = server.create('organization', {name: 'Test org'});
-//    server.create('organizationUser', {user: user, organization: organization});
-//    this.loginUser = user;
-//  })
+//
+// let user;
+// beforeEach(function() {
```
unicornzero: I thought it may be helpful to put a note here about what I think should be best practices for test setup. Let me know if you have a different perspective on this.

tests/helpers/setup-acceptance.js
```diff
@@ -32,30 +32,43 @@ export default function setupAcceptance({authenticate = true} = {}) {
...
+//   server.create('organizationUser', {user: user, organization: organization});
+// });
+// setupSession(function() {
+//   this.loginUser = user;
+// });
 export function setupSession(createData) {
   beforeEach(function() {
     createData.bind(this)(server);
+
+    // If no loginUser is setup and there is only one user created, log that user in
```
unicornzero: I thought about ensuring there are no other users with `_currentLoginInTest` set to `true` here (I ran into some weird test setup cases where there was some state bleed), but figured I could leave that out for now.

app/templates/organization/project/settings/index.hbs
```diff
@@ -14,65 +14,73 @@
...
-    <div class="bg-white rounded shadow-1 p-2 mb-2">
-      <h2 class="text-xl font-semibold mb-sm">Project token</h2>
-      <p class="text-secondary mb-2">See the {{#link-to "docs" class="font-semibold"}}docs{{/link-to}} for how to set up client authentication for your app.</p>
-      <div class="p-2 bg-gray-000 rounded border">
-        <div class="mb-2">
-          {{projects/env-vars project=project}}
+    {{#if (and (variation 'only-admins-edit-settings') (not isUserOrgAdmin))}}
+      <div data-test-member-view class="Alert Alert--error">
+        <strong>Editing this organization's project settings requires organization admin permissions.</strong>
+        {{project.organization.currentUserIsAdmin}}
```
cadeParade: This will render true/false to the template?

tests/acceptance/login-test.js
cadeParade: fair enough

tests/acceptance/project-test.js
```diff
@@ -145,7 +145,80 @@ describe('Acceptance: Project', function() {
...
+      beforeEach(function() {
+        withVariation(this.owner, 'only-admins-edit-settings', true);
+      });
+
+      afterEach(function() {
+        withVariation(this.owner, 'only-admins-edit-settings', false);
+      });
+
+      describe('user is member', function() {
+        setupSession(function() {});
```
cadeParade: I would guess that you've confirmed that calling `setupSession` at all is necessary in this case?
If so an alternative syntax would be setupSession(() => {}).

app/templates/organization/project/integrations/index.hbs
```diff
@@ -5,47 +5,53 @@
...
-
         <div>This is the integrations page for a <strong>demo project</strong>. The demo settings are not editable here, but <strong>{{#link-to 'organizations.organization.projects.new' project.organization.slug}} create your own project {{/link-to}}</strong> to start customizing your Percy experience.
         </div>
       </div>
     {{/if}}
 
+    {{#if (and (variation 'only-admins-edit-settings') (not isUserOrgAdmin))}}
+      <div data-test-member-view class="Alert Alert--error">
+        <strong>Editing this organization's integration settings requires organization admin permissions.</strong>
+        {{project.organization.currentUserIsAdmin}}
```
cadeParade: This will render true/false to the template?

tests/acceptance/project-test.js
cadeParade: It probably makes sense in the long term to rename `setupSession` to something related to logged in user and maybe update it to not accept a function. Currently I believe the passed function is often used to create factory data, but we don't need it to do that, it's purely for setting current user. Anyway, outside the scope of this PR but thanks for digging through these issues.

tests/acceptance/project-test.js
```diff
@@ -195,6 +268,8 @@ describe('Acceptance: Project', function() {
           },
         };
 
+        setupSession(function() {});
```
cadeParade: Sorry I think I'm missing something. Why is `setupSession` being called so many times? Do we need to change the logged in user in every test block?

tests/acceptance/project-test.js
unicornzero: Agreed; that makes sense to rename later. Thanks!

app/templates/organization/project/settings/index.hbs
unicornzero: oh crud, forgot to delete that. thank you for catching it.

tests/acceptance/project-test.js
unicornzero: That's cleaner. Thanks! Yeah, we still need it to setup a user as the logged in user.

app/templates/organization/project/integrations/index.hbs
unicornzero: eep, yep, leftover debugging. deleting... thanks!

tests/acceptance/project-test.js
unicornzero: We need to setup a logged-in user and if we don't do it at a higher level test block then we have to do it in the lower level ones. If we want to setup a default for the whole bigger block and just override it as needed, then we could use that helper that I wrote but have deleted:

```
export function changeLoggedInUser(context, newUser) {
  server.schema.users.where({_currentLoginInTest: true}).update('_currentLoginInTest', false);
  context.loginUser = newUser;
  newUser.update({_currentLoginInTest: true});
}
```

tests/acceptance/project-test.js
```diff
@@ -195,6 +268,8 @@ describe('Acceptance: Project', function() {
           },
         };
 
+        setupSession(() => {});
```
cadeParade: I'm just going on indents here but it looks like this one and the one on line 255 are inside the scope of `setupSession` on 220, so maybe aren't needed?

unicornzero: Great eye/catch,  thanks!!!

PR #1528, opened 2020-01-18
Refactor fullscreen route

cadeParade:
## Description
This PR rearranges the fullscreen snapshot route to be at 
`/:org/:project/builds/:build_id/view/snapshot_id?browser=:browser&mode=:mode&width=:width`
instead of:
`/:org/:project/builds/:build_id/view/snapshot_id/:width?browser=:browser&mode=:mode`

The width parameter has been moved from the last required param to a basically optional query param. This allows us a few things:
- Move most of `builds/:build_id` route to `builds/:build_id/index` route. This allows us to **view a fullscreen snapshot without loading every snapshot in the build first**
- Set a default width param if we don't know all the widths of a snapshot. So we can say "Go to fullscreen snapshot for this snapshot and we **don't need to know ahead of time what widths it has.**
- Remove the strange modal switching that happened when switching between fullscreen and build view -- the toggle button is now an anchor tag and comes with anchor tag things like **cmd+click to open new tab.**

This PR also incidentally consolidated some of the snapshot loading logic so we are **repeating queries less** (using ember-storefront `loadRecord/s`).

Followups:
- Add [`ember-router-scroll`](https://github.com/DockYard/ember-router-scroll) which will give us scroll position retention.
- Put back fullscreen up/down snapshot nav.
- Nicer loading state for fullscreen comments.
- Write a redirect service so people with old fullscreen links can get to their old snapshots?

## Reviewing this PR

thank you all the tests ❤️


app/adapters/application.js
```diff
@@ -100,7 +100,6 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
 
     if (requestType === 'query' && modelName === 'snapshot') {
       const buildId = query.filter.build;
-      delete query.filter.build;
```
cadeParade: This line (and 144) prevented ember-data-storefront from caching the request responses. Without these lines, there seems to be no functional loss and we get query caching, so, winning?

app/components/snapshot-viewer-full.js
```diff
@@ -84,7 +86,10 @@ export default Component.extend(EKMixin, {
   },
 
   onEscKeyPress: on(keyDown('Escape'), function() {
-    this.closeSnapshotFullModal();
+    this.router.transitionTo(
```
cadeParade: Since this is a real route now, we can use regular `transitionTo` to move between the two screens.

app/models/snapshot.js
```diff
@@ -25,7 +26,7 @@ export const DIFF_REVIEW_STATE_REASONS = [
   SNAPSHOT_REVIEW_STATE_REASONS.USER_REJECTED_PREVIOUSLY,
 ];
 
-export default DS.Model.extend({
+export default DS.Model.extend(LoadableModel, {
```
cadeParade: Required to use storefront methods on snapshot models.

app/models/comment-thread.js
```diff
@@ -1,10 +1,11 @@
 import DS from 'ember-data';
 import {and, empty, equal, notEmpty} from '@ember/object/computed';
+import LoadableModel from 'ember-data-storefront/mixins/loadable-model';
 
 export const REVIEW_COMMENT_TYPE = 'request_changes';
 export const NOTE_COMMENT_TYPE = 'note';
 
-export default DS.Model.extend({
+export default DS.Model.extend(LoadableModel, {
```
cadeParade: Required to use storefront methods on comment-thread models.

app/routes/organization/project/builds/build.js
```diff
@@ -1,113 +1,22 @@
...
-    const org = this.modelFor('organization');
-    return hash({
-      // Note: passing {reload: true} here to ensure a full reload including all sideloaded data.
-      build: this.store.findRecord('build', params.build_id, {reload: true}),
-      isUserMember: isUserMember(this.session.currentUser, org),
-      commentThreads: this.commentThreadQuery.getCommentsForBuild(params.build_id),
-    });
-  },
-
-  afterModel(model) {
```
cadeParade: All of this got moved to `builds/build/index.js`. Or pushed to a service (review service or comment thread service) which takes care of the nasty logic.

app/routes/organization/project/builds/build/index.js
```diff
@@ -0,0 +1,64 @@
+import Route from '@ember/routing/route';
```
cadeParade: This is a pared down copy of the old `builds/build.js`

app/routes/organization/project/builds/build/snapshot.js
```diff
@@ -23,28 +25,55 @@ export default Route.extend(ResetScrollMixin, {
     });
   },
 
+  afterModel(resolvedModel) {
+    const snapshotId = resolvedModel.snapshot.id;
+    const build = this.modelFor('organization.project.builds.build');
+    // do not return from this method so comments load in the background and
+    // does not block template rendering.
+    this.commentThreads.getCommentsForSnapshotIds([snapshotId], build);
```
cadeParade: This is new behavior -- the comment threads will be queried separately and be displayed on the screen whenever they load. A followup will be to add a nicer loading state for waiting on the comments in fullscreen mode.

app/routes/organization/project/builds/build/snapshot.js
```diff
@@ -104,59 +129,6 @@ export default Route.extend(ResetScrollMixin, {
       this._updateQueryParams({newWidth: width});
       this._track('Fullscreen: Width Switched', {width});
     },
-
-    // This is how navigating through snapshots is determined in full screen snapshot mode.
```
cadeParade: Temporarily (?) removed.

app/services/comment-threads.js
```diff
@@ -0,0 +1,74 @@
+import Service, {inject as service} from '@ember/service';
```
cadeParade: This file was renamed from `comment-thread-query` to `comment-threads`

cadeParade: All the logic for creating comments/comment threads was moved into here also, out of the route. Making the route much cleaner and all logic related to comment threads centralized and accessible from anywhere.

app/services/reviews.js
```diff
@@ -2,12 +2,31 @@ import Service from '@ember/service';
...
 
 export default Service.extend({
   store: service(),
   analytics: service(),
   snapshotQuery: service(),
-  commentThreadQuery: service(),
+  commentThreads: service(),
+  confirm: service(),
+
+  createReview: task(function*({snapshots, build, eventData}) {
```
cadeParade: Moved the logic for creating reviews into a service so all interaction with reviews and the API are centralized and available from anywhere.

app/services/snapshot-query.js
```diff
@@ -1,10 +1,19 @@
 import Service, {inject as service} from '@ember/service';
 import {SNAPSHOT_REVIEW_STATE_REASONS, DIFF_REVIEW_STATE_REASONS} from 'percy-web/models/snapshot';
 
+const SNAPSHOT_COMPARISON_INCLUDES = Object.freeze([
```
cadeParade: Start explicitly including things we want on snapshots. Setting us up in the future for when we might not want to include the world with snapshots.

app/services/snapshot-query.js
```diff
@@ -1,10 +1,19 @@
...
+  'comparisons.base-screenshot.lossy-image',
+  'comparisons.diff-image',
+  'comparisons.browser',
+]);
+
 export default Service.extend({
   store: service(),
   getUnchangedSnapshots(build) {
-    return this.store.query('snapshot', {
+    return this.store.loadRecords('snapshot', {
```
cadeParade: Using the ember-storefront `loadRecord` method instead of `query` allows us to never make this request more than once in a blocking manner. It will make the snapshot request one time as a blocking promise and after that, if we ever ask for it again, it will automatically load the data we got back the first time and refresh the data in the background. This is a pretty big win imo.

app/services/tooltips.js
```diff
@@ -100,7 +101,7 @@ export default Service.extend({
       return GROUP_SEQUENCE;
     }
 
-    const snapshots = await build.get('snapshots');
+    const snapshots = await this.snapshotQuery.getChangedSnapshots(build);
```
cadeParade: Not technically necessary, but I noticed that we were making a request for `build/id/snapshots` sometimes, which is a huge payload. Nothing was blocked on it but no need to get all that data. Also, since this is using the `snapshotQuery` method which now uses `loadRecords`, it will not make an additional request 👍

app/templates/components/snapshot-group.hbs
```diff
@@ -30,13 +30,22 @@
...
         {{#if isGroupCollapsed}}
-          <div class="btn-group" role="group" aria-label="Second Group">
-            <button type="button" class="ToggleFullViewButton btn btn-square maximize-snapshot" {{action showSnapshotFullModalTriggered snapshots.firstObject.id blockSelectedWidth activeBrowser}} data-test-snapshot-header-dropdown-toggle>
-              <figure class="svg-container">
-                <ThemeSvg @svg="maximize-icon" />
-              </figure>
-            </button>
-          </div>
+
+          <LinkTo
```
cadeParade: Yessss, that button is an `<a>` tag now! Now you can `cmd+click` and open a fullscreen snapshot in a new tab 👍👍👍👍👍👍👍👍👍

app/templates/components/snapshot-viewer-header.hbs
```diff
@@ -1,17 +1,6 @@
 {{!-- template-lint-disable no-nested-interactive --}}
 <header class="SnapshotViewer-header {{if fullscreen 'SnapshotViewer-header-full w-screen z-2'}}" data-test-SnapshotViewer-header {{action expandSnapshot}} role="button">
   <div style="min-width: 0;" class="flex items-center">
-    {{#if fullscreen}}
```
cadeParade: Temporarily removing these nav buttons

app/templates/components/snapshot-viewer-header.hbs
```diff
@@ -135,15 +124,35 @@
       </div>
 
       <div class="btn-group" role="group" aria-label="Second Group">
-        <button type="button" class="ToggleFullViewButton btn btn-square {{if fullscreen 'minimize-snapshot' 'maximize-snapshot'}}" {{action toggleViewMode}} aria-label="{{if fullscreen "Minimize" "Maximize"}} snapshot" data-test-SnapshotViewer-toggleFullScreen>
-          <figure class="svg-container">
-            {{#if fullscreen}}
+        {{#if fullscreen}}
+          <LinkTo
```
cadeParade: Yessss, that button is an <a> tag now! Now you can cmd+click and open a fullscreen snapshot in a new tab 👍👍👍👍👍👍👍👍👍

tests/acceptance/build-test.js
```diff
@@ -1130,68 +1128,6 @@ describe('Acceptance: Fullscreen Snapshot', function() {
     expect(BuildPage.snapshotFullscreen.diffImageUrl).to.equal(TEST_IMAGE_URLS.DIFF_URL);
   });
 
-  describe('next/previous snapshot navigation', function() {
```
cadeParade: Temporarily removed.

app/adapters/application.js
cadeParade: I don't think this is a problem in the way you're imagining. The purpose of this method is to get some data together to push over to `environment.js` `apiUrls` property. So what this does is get the build id from the filter query param [here (for example)](https://github.com/percy/percy-web/pull/1528/files#diff-d2752d44b2e9a8619416b574de7611e8R27)  and shoves that id into the url `buildSnapshots` (`'/api/v1/builds/%@/snapshots'`). 
**TLDR**: This file is unrelated to any state of any models, it solely exists to builds a url.

Storefront uses the query params to make sure it loads from cache the data from the right query (and it does this after the `delete` line, so when the delete was there, it wasn't caching correctly). This, combined with the fact that nothing is returned from `buildURL` method, (and also nothing broke) seems like it's ok to remove these lines. 
`buildURL` method does call `super` at the end of the method and passes the query object.  But we don't see any ill effects from removing these two `delete`s ...

This file should eventually be replaced with smart use of json api links. This is what the links property on json api payloads are for. I think that this pattern was probably started before Ember got smart to JSON API so was probably a workaround from a long time ago.

cadeParade: Oh yeah, and slugs

PR #1534, opened 2020-01-21
Sanitize @mention name

cadeParade:
## Description
Ember automatically sanitizes strings rendered through `{{}}` but this string goes through `tribute.js` land and they render the html we give it straight to the page. Sanitize these strings so we can render it straight to the page. 

## Reviewing this PR


cadeParade:
Yes, this is the only place it's being used.

PR #1536, opened 2020-01-22
Replace descriptor_uri with addon_key

faun:
Description
===

Per the instructions [here](https://developer.atlassian.com/cloud/bitbucket/install-an-app-from-your-site/), it is required to use the `addon_key` query param instead of `descriptor_uri` in the link to connect a bitbucket app that has been registered.

[Change notice](https://developer.atlassian.com/cloud/bitbucket/2019-11-18-installation-url/)

This change updates the integration URL for the production "Install Bitbucket Cloud App" link.

[Clubhouse card](https://app.clubhouse.io/percy/story/3067/bitbucket-integrations-are-broken)

Before:

Before

After:

After


faun:
https://app.clubhouse.io/percy/story/3067/bitbucket-integrations-are-broken

PR #1537, opened 2020-01-22
Add logical wrapping boundaries to build created in build dropdown

faun:
This prevents wrapping issues like [this one](https://percy.io/percy/percy-web/builds/3879867/view/231673006/375?browser=firefox&mode=diff)

This breaks the word wrap between the date and the time only. Otherwise, there is no wrapping.

Wrapping:
Screen Shot 2020-01-22 at 11 41 18 AM

Full width:
Screen Shot 2020-01-22 at 11 41 49 AM


cadeParade:
@faun Thank you for doing this 🙏

PR #1539, opened 2020-01-22
Update organization users after invite.

cadeParade:
## Description
When a user accepts an invite, they are forwarded to the invite's organization page and can view the projects, but can only see the projects as a non-member viewer. This is because after the invite is accepted, the `organization-user` relationships were not being refreshed in the browser, so the app did not know that the user was now part of the organization. 
To address this, we reload the user, which includes `organization-user` after an invite is accepted.


tests/acceptance/join-test.js
```diff
@@ -38,9 +43,16 @@ describe('Acceptance: Join', function() {
...
-    expect(currentRouteName()).to.equal('organizations.organization.projects.new');
-
-    await percySnapshot(this.test.fullTitle() + ' | Into organization');
+    await click('[data-test-percy-btn-label=invite-accept-button]');
+    expect(currentRouteName()).to.equal('organization.index');
+    expect(OrganizationDashboard.projects.length).to.equal(1);
+    await percySnapshot(this.test.fullTitle(), ' | Into organization');
+
+    await OrganizationDashboard.projects[0].clickLink();
+    await this.pauseTest();
```
djones: hangover from testing?

app/components/invitation-handler.js
```diff
@@ -1,32 +1,34 @@
...
   session: service(),
+  router: service(),
   currentUser: alias('session.currentUser'),
 
   flashMessages: service(),
+
+  accept: task(function*() {
+    try {
+      const model = yield this.invitation.save();
+      yield all([timeout(500), model]);
```
djones: maybe centralize the 500ms wait into a meaningfully named constant? Something like:

```suggestion
      yield all([timeout(MIN_FEEDBACK_DELAY_MS), model]);
```

Naming TBD 😄

PR #1540, opened 2020-01-23
Fix stripe card syntax :(

cadeParade:
## Description
Where the tests don't reach...there is pain :(
Can't write a test for this because the stripe elements form won't load in tests.


cadeParade:
🤦‍♀

PR #1544, opened 2020-01-24
When hitting deprecated fullscreen snapshot route, forward to new route.

cadeParade:
Redirect to new snapshot route when an old one is asked for. 
Also use storefront to cache comment thread queries.


app/routes/organization/project/builds/build/deprecated-snapshot.js
```diff
@@ -0,0 +1,19 @@
...
+    activeBrowserFamilySlug: {as: 'browser', refreshModel: true},
+  },
+
+  redirect() {
+    const params = this.paramsFor(this.routeName);
+    this.transitionTo('organization.project.builds.build.snapshot', params.snapshot_id, {
+      queryParams: {
+        mode: params.comparisonMode,
+        activeBrowserFamilySlug: params.activeBrowserFamilySlug,
+        width: params.width,
```
unicornzero: Do we not need this in `queryParams` because it's part of the route? Is it called a route param?

cadeParade: I am not sure I understand your question. `queryParams` property to `transitionTo` is how you tell ember what query params to put on the end of your route. (ex: https://guides.emberjs.com/release/routing/query-params/#toc_transitionto)

unicornzero: Oops, I should have clarified I meant this:
Do we not need width in the queryParams block on lines 4-7 because width is coming in as a route param?

cadeParade: Ahh, yes, that's right. The old route did not use width as a query param, it was a dynamic segment (I think maybe that was your original question? What it is called?). But you can get dynamic segments and query params both through the `params` object. But query params in the url only show up in `params` object if they are also defined explicitly as query params, as you see on lines 4-7. 
Does that answer your question?

unicornzero: "Dynamic segment" not "route param" is the lingo, gotcha! Yes, fully answers my question, thanks!

PR #1551, opened 2020-01-28
Template lint -- no implicit this

cadeParade:
## Description
In preparation for glimmer components, templates need to declare `this` before variables used in the template. 
This means that the only time there is a `{{foo}}` used in a template it will be a helper (built-in or something in app/helpers directory). 
Glimmer components will throw an error if you try to refer to a js variable without `this`. So enforcing this now will ease the transition to full glimmer components in the future.


app/templates/components/forms/organization-invite.hbs
```diff
@@ -20,22 +20,27 @@
...
-    {{#link-to data-test-cancel-button "organizations.organization.users" organization.slug class="percy-btn"}}
+    <LinkTo
+      @route="organizations.organization.users"
+      @model={{this.organization.slug}}
+      @data-test-cancel-button={{true}}
+      class="percy-btn"
+    >
       Cancel
-    {{/link-to}}
+    </LinkTo>
```
djones: ha, how'd that sneak through the previous PR?

PR #1554, opened 2020-01-28
Do not autoplay video in tests. Try to hide animated border of mascot.

cadeParade:
## Description
Chrome 78 made one of the tests real flaky. Turns out the youtube video is playing in the tests which causes the play bar to be flaky. Also there's an animation that plays around the percy mascot bubble which flakes. Attempt to hide this element so we don't get the flakes.


paulfri:
We attempt to correct for this in the rendering environment. I'm curious why the autoplay attribute isn't being stripped.

PR #1559, opened 2020-01-31
Convert not-component js files to native classes

cadeParade:
## Description
Convert all not-component js files to use native classes. This is a step to bringing the app up to the new Ember Octane standards. 
Most code changes were generated with [ember-native-class-codemod](https://github.com/ember-codemods/ember-native-class-codemod). The things I did on top of that are:
- Remove the `@classic` decorator where possible/easy
- Convert some `this.get`s and `this.set`s to `get/set(this, 'whatever)`
- Add a comment above remaining instances of `@classic` decorator indicating when we can remove it. 

Here is a nice site that I've been using as a reference if you're interested 
https://www.notion.so/classic-Decorator-Workflow-f3f30898679c4f44b33179929aa127f1
Here also is a thread about decorator usage: https://discuss.emberjs.com/t/question-about-the-native-decorator-support-feature/17474

Next step is probably converting components to native classes/glimmer components. This is a much larger step because much more will have to change in component files when converting to glimmer components. This update will probably be slower and more gradual due to the added complexity.


app/adapters/block.js
```diff
@@ -1,3 +1,3 @@
 import ContentfulAdapter from 'ember-data-contentful/adapters/contentful';
 
-export default ContentfulAdapter.extend({});
+export default class Block extends ContentfulAdapter {}
```
djones: Definitely reads better

PR #1571, opened 2020-02-07
Refactor fullscreen query params

cadeParade:
## Description
This PR serves two purposes:
- Fix bug with sometimes going back to a previous comparison in fullscreen
- Refactor fullscreen route query params to be more ember-y

The bug was triggered by going to fullscreen, updating browser/width/comparison mode, then clicking close fullscreen button. This would, under the hood, call `window.history.back` and go back to the latest set of query params. To fix this, the query params have an arg `replace: true` so it does not add any state to the history. When it goes back, it goes to the build page.

The refactoring was was triggered by trying to add the above `replace: true` params. It was not working correctly with the previous query param setup and moving the query param controls to the controller is a more idiomatic way to manage query params anyway.

One thing to note is that this PR removes the update to the url when a user is forwarded to an incorrect browser. A user's screen will still be forwarded to the correct browser and the UI will reflect such, but the url does not change with this update. This only would happen if they type a wrong browser in the URL manually. The app would not forward a user to an incorrect browser. Similar with default comparison mode. I have not quite figured out how to get the redirects working correctly in these cases, but they are not hit often and I will figure it out soon. Better to fix the bigger bug affecting users right now.

## Reviewing this PR


tests/acceptance/build-test.js
```diff
@@ -1519,8 +1519,10 @@ describe('Acceptance: Fullscreen Snapshot', function() {
   });
 
   it('forwards to fullscreen snapshot when hitting old route', async function() {
-    // eslint-disable-next-line
-    await visit(`/${urlParams.orgSlug}/${urlParams.projectSlug}/builds/${urlParams.buildId}/view/${urlParams.snapshotId}/${urlParams.width}?mode=${urlParams.mode}&browser=${urlParams.browser}`);
+    await visit(
+      // eslint-disable-next-line
```
cadeParade: This diff was prettier

PR #1591, opened 2020-03-09
Server sorted snapshot list

cadeParade:
## Description
Currently, loading times for large builds is extremely slow because we send all snapshots in a build in one initial request on page load. The reason for this is because we need to show snapshots to users in the correct "order" -- more "important" snapshots should be sorted first. "Importance" is calculated on many things, including relationships (comparisons, comments) which means that we also need to have all the relevant relationships loaded when calculating the sort order. This causes initial loads for a build to be enormous and slow.

This PR updates the build page to load snapshots incrementally using a new sort metadata object coming from the API. The sort metadata object lists all snapshot ids and groups in the build in the correct order so the build page can pull them in as needed. 

There were two main categories of problems to solve for this: 
(1) displaying/interacting with snapshots incrementally (infinite scroll, pagination, keyboard nav, etc)
(2) how to handle write actions when not every snapshot is in the store (build approve all, counting unreviewed snapshots, etc)

### (1) Infinite scroll
The main logic for infinite scroll is based on the ember-map infinite scroll concept and is found in `infinite-snapshot-page.js` and `infinte-snapshot-list.hbs`. 
`infinite-snapshot-list` loops over a range from 1-`page` and for each item in the range renders a `InfiniteSnapshotPage` block. The `InfiniteSnapshotPage` block takes the page and the sort metadata, figures out which snapshots to load, and returns them to the template to be iterated over. 
There is an invisible sentinel at the top of each `InfiniteSnapshotPage` block (starts with `{{in-viewport`) that increments the page, which in turn increments the range we're keeping track of at the top, and adds another `InfiniteSnapshotPage` block to the end of the list.

`infinite-snapshot-list` renders `infinite-snaphshot-group` and `infinite-snapshot-viewer`. These are mostly copies of their more plain siblings, `snapshot-group` and `snapshot-viewer`. But there were enough small differences between them that putting them all behind the flag was too tedious so I copied them. Only the JS files are duplicated, the `infinite` versions of `snapshot-viewer` and `snapshot-group` use the same templates as the originals.
The differences between plain and infinite versions were related to: (a) lazy loading and (b) keyboard nav. Lazy loading could just plain go away (🎉) and keyboard nav got much easier since we can use the sort metadata provided index to calculate previous and next snapshots. 

#### Known/accepted regressions related to infinite scroll:
- When unchanged snapshots are visible, keyboard nav will not be able to transition from changed snapshots to unchanged snapshots. You will have to click the first unchanged snapshot to start keyboard nav for unchanged snapshots.
- Going from build list to fullscreen back to build list will lose your position 😢 

### (2) Handling when not every snapshot is in the store/ two sources of truth
The main problem encountered here is related to things that require knowing something about every snapshot. The two biggest problems were: (a) approving an entire build, (b) counting unreviewed snapshots after a snapshot state had changed.

(a) When approving an entire build, if there are rejected snapshots, we would like to show a modal asking if the user wants to override the rejected state. To know whether there are any snapshots with changes requested we check all the loaded snapshots first, then use the sort metadata to find out if any unloaded snapshots are rejected. We delete loaded snapshots from the unloaded snapshot check so we do not get conflicting states for the same snapshot.

(b) At the top of a build with more than one browser, there is a browser switcher component with a count of how many unreviewed snapshots are left in each browser. This number should update if a snapshot is reviewed.
The initial count can be calculated by the snapshots and states within the sort metadata object, but once a snapshot is approved locally, the number should be udpated and can no longer depend directly on the sort metadata (since that does not get updated) (but I have some ideas). To update the numbers, we call an update action after the review is saved which (1) calculates unapproved snapshots for build in the store, (2) subtracts those snapshots from the metadata list, (2) adds unapproved snapshots from metadata list.

#### Known/accepted regressions related to store state
- We will no longer block approving a build if there are open review comments but the snapshot is not rejected (possibly temporary)

Finally, there is a new file, `sort-metadata.js` which is a utility class that stores and processes the data in the sort metadata object. It is not a beautiful file but hopefully since all the logic related to the metadata is contained there, it will be improved over time.


tests/acceptance/billing-test.js
```diff
@@ -31,7 +31,7 @@ describe('Acceptance: Billing', function () {
 
     it('shows the correct card expiration date', async function () {
       // Static date at beginning of month to ensure time localization doesn't change display date
-      const expirationDate = new Date('2055-11-1 UTC');
+      const expirationDate = new Date('Nov 1 2055 00:00:00 UTC');
```
cadeParade: IDK the old way of declaring the date fails in firefox??? ¯\_(ツ)_/¯

tests/integration/components/snapshot-list-test.js
```diff
@@ -209,15 +207,6 @@ describe('Integration: SnapshotList', function () {
       expect(secondSnapshotBlock.isFocused).to.equal(false);
       expect(lastSnapshotBlock.isFocused).to.equal(false);
     });
-
-    it('does not send keyboard actions when isKeyboardNavEnabled is false', async function () {
```
cadeParade: `isKeyboardNavEnabled` is never false, or true, it no longer exists and hasn't for a while.

app/components/comparison-viewer.js
```diff
@@ -21,6 +21,10 @@ export default Component.extend({
     this.set('showDiffOverlay', this.allDiffsShown);
   }),
 
+  headSnapshot: computed('comparison.id', function () {
```
cadeParade: This sometimes caused  n+1 requests because it was asking for `headSnapshot`  we are no longer sending

app/components/infinite-snapshot-group.js
```diff
@@ -0,0 +1,97 @@
+import {computed, get} from '@ember/object';
```
cadeParade: Mostly a copy of `snapshot-group.js` but inherits from `InfiniteSnapshotListItem`.

app/components/infinite-snapshot-list-item.js
```diff
@@ -0,0 +1,130 @@
+import Ember from 'ember';
```
cadeParade: Much of this is the same as `snapshot-list-item` but keyboard nav, lazy loading, active state, etc have changed.

app/components/infinite-snapshot-list.js
```diff
@@ -0,0 +1,89 @@
+import Component from '@ember/component';
```
cadeParade: Much the same as `snapshot-list` but lazy loading, keyboard nav, active state, etc has changed.

app/components/infinite-snapshot-viewer.js
```diff
@@ -0,0 +1,80 @@
+import {filterBy, notEmpty, or, readOnly} from '@ember/object/computed';
```
cadeParade: Mostly a copy of `snapshot-viewer.js` but inherits from `InfiniteSnapshotListItem`

app/components/snapshot-list.js
```diff
@@ -135,24 +134,18 @@ export default Component.extend(EKMixin, {
   },
 
   onDKeyPress: on(keyDown('KeyD'), function () {
-    if (this.isKeyboardNavEnabled) {
-      this.toggleAllDiffs({trackSource: 'keypress'});
-      this._trackKeyPress();
-    }
+    this.toggleAllDiffs({trackSource: 'keypress'});
```
cadeParade: `isKeyobardNavEnabled` is no longer a thing. Just cleaning up here.

app/controllers/organization/project/builds/build/index.js
```diff
@@ -4,17 +4,41 @@ import Controller from '@ember/controller';
...
+  @action
+  async fetchSnapshots(build) {
+    await this.fetchSnapshotsWithSortOrder(build);
+  }
+
+  async fetchSnapshotsWithSortOrder(build) {
+    const snapshotsAndMeta = await this.snapshotQuery.getSnapshotsWithSortMeta(build);
+    const meta = snapshotsAndMeta.meta['sorted-items'];
+    const sortObject = metadataSort.create({sortData: meta, build, store: this.store});
+    this.build.setProperties({
```
cadeParade: Set the sort metadata object on the build so we can access it wherever we have the build.

app/models/build.js
```diff
@@ -373,12 +377,13 @@ export default class Build extends Model {
   @computed('loadedSnapshots.@each.{isUnreviewed,isUnchanged}', 'browsers.[]')
   get unapprovedSnapshotsWithDiffForBrowsers() {
     const loadedSnapshotsForBuild = this.loadedSnapshots;
-    const unreviewedSnapshotsWithDiffs = loadedSnapshotsForBuild.reduce((acc, snapshot) => {
-      if (snapshot.get('isUnreviewed') && !snapshot.get('isUnchanged')) {
-        acc.push(snapshot);
-      }
-      return acc;
-    }, []);
+    const unreviewedSnapshotsWithDiffs = loadedSnapshotsForBuild.filter(snapshot => {
```
cadeParade: The previous reduce was just a filter so this is slightly simpler.

app/templates/components/snapshot-group.hbs
```diff
@@ -164,26 +164,43 @@
...
-                @updateActiveSnapshotBlockId={{this.updateActiveSnapshotBlockId}}
-                @activeBrowser={{this.activeBrowser}}
-                @updateActiveBrowser={{this.updateActiveBrowser}}
-                @isBuildApprovable={{this.isBuildApprovable}}
-                @toggleOverlay={{action "toggleGroupOverlay"}}
-                @showDiffOverlay={{this.isGroupShowingDiffOverlay}}
-                @index={{index}}
-                @class="snapshot-group-snapshot rounded-l flex-1 border-l"
-              />
+              {{#if (variation 'snapshot-sort-api')}}
```
cadeParade: Because I am still using the original templates for `infinite-snapshot-group` and `infinite-snapshot-viewer`, we need to check in this template whether the flag is enabled and change `snapshot-viewer` to `infinite-snapshot-viewer` if a group is expanded.

mirage/config.js
```diff
@@ -391,40 +402,53 @@ export default function () {
...
-    const snapshots = schema.snapshots.find(attrs.snapshotIds);
     const reviewState =
       attrs.action === REVIEW_ACTIONS.APPROVE ? SNAPSHOT_APPROVED_STATE : SNAPSHOT_REJECTED_STATE;
     const reviewStateReason =
       attrs.action === REVIEW_ACTIONS.APPROVE
         ? SNAPSHOT_REVIEW_STATE_REASONS.USER_APPROVED
         : SNAPSHOT_REVIEW_STATE_REASONS.USER_REJECTED;
 
-    snapshots.models.forEach(snapshot => {
+    const snapshots = (() => {
```
cadeParade: Update review endpoint to handle a build approval when snapshot ids are not included.

tests/acceptance/build-test.js
```diff
@@ -382,28 +382,6 @@ describe('Acceptance: Build', function () {
       expect(BuildPage.isConfirmDialogVisible).to.equal(false);
       expect(firstSnapshot.isApproved).to.equal(true);
     });
-
-    // eslint-disable-next-line
-    it('blocks approval of build if there are open review threads on build', async function() {
```
cadeParade: Not doing this for builds for now.

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
+import setupAcceptance, {setupSession} from '../helpers/setup-acceptance';
```
cadeParade: These tests are copied from `build-test` but updated to work with infinite loading snapshot list. The tests from `build-test` can be deleted when we fully switch over to this loading method.

tests/helpers/setup-acceptance.js
```diff
@@ -30,7 +30,6 @@ export default function setupAcceptance({authenticate = true} = {}) {
       server.shutdown();
     }
   });
-
```
cadeParade: ¯\_(ツ)_/¯

tests/integration/components/infinite-snapshot-group-test.js
```diff
@@ -0,0 +1,382 @@
+/* jshint expr:true */
```
cadeParade: These tests are copied from `snapshot-group-test` and updated to work with new scheme. The old tests can be deleted once we switch over.

tests/integration/components/infinite-snapshot-list-test.js
```diff
@@ -0,0 +1,307 @@
+/* jshint expr:true */
```
cadeParade: These tests are copied from snapshot-list-test and updated to work with new scheme. The old tests can be deleted once we switch over.

tests/integration/components/infinite-snapshot-viewer-test.js
```diff
@@ -0,0 +1,344 @@
+/* jshint expr:true */
```
cadeParade: These tests are copied from snapshot-viewer-test and updated to work with new scheme. The old tests can be deleted once we switch over.

app/components/build-container.js
```diff
@@ -36,21 +38,33 @@ export default Component.extend(PollingMixin, {
...
-      return browserWithMostDiffs;
-    } else if (chromeBrowser) {
-      return chromeBrowser;
+  defaultBrowser: computed('_browsers', 'build.isFinished', 'browserWithMostDiffs', function () {
+    if (
+      this.launchDarkly.variation('snapshot-sort-api') &&
+      this.build.isFinished &&
+      this.build.sortMetadata
+    ) {
+      const defaultBrowserSlug = this.build.sortMetadata.defaultBrowserSlug;
```
djones: yay for getting this from the API now

unicornzero:
Would it be helpful to have the review endpoint return metadata if your request has a param asking for it?

cadeParade:
Well, the point of having sort metadata object was so that it was not refreshed until a page reload right? I have fixed my complaint with the unreviewed snapshot count, found a way to observe snapshots in the store, so that's ok for now. 
I do think as we continue to refine this system I'll need to come up with a smarter way of handling the two-sources-of-truth deal that's going on but I have some ideas...

app/components/infinite-snapshot-list.js
```diff
@@ -0,0 +1,89 @@
...
+    // TODO(sort): handle keyboard nav for unchanged snapshots when they have indexes
+    const numItems = this.blockItems.length;
+    if (!this.isActiveSnapshotIndex) {
+      this._updateActiveBlockIndex(0);
+    } else {
+      let currentIndex = this.activeSnapshotBlockIndex;
+
+      const newIndex = isNext ? currentIndex + 1 : currentIndex - 1;
+      if (newIndex < numItems && newIndex >= 0) {
+        this._updateActiveBlockIndex(newIndex);
```
unicornzero: Ah, nice. So it simply stops if you get to the beginning or end without worrying about wrapping.

app/components/infinite-snapshot-list.js
```diff
@@ -0,0 +1,89 @@
...
+      }
+    }
+    this._trackKeyPress();
+  },
+
+  _updateActiveBlockIndex(newIndex) {
```
unicornzero: I'm curious, is this an Ember best practice, to provide a helper method around the setter or is it something you simply find more readable?

app/components/infinite-snapshot-list.js
```diff
@@ -0,0 +1,89 @@
...
+    updateActiveSnapshotBlockIndex(newIndex) {
+      this._updateActiveBlockIndex(newIndex);
+    },
+    // TODO(sort) remove when unchanged items have indexes
+    updateActiveSnapshotBlockId() {},
+  },
+
+  init() {
+    this._super(...arguments);
+    this.blockItems = this.blockItems || [];
```
unicornzero: Nice. This version reads a lot simpler not having to track all the different block types separately.

app/components/infinite-snapshot-page.js
```diff
@@ -0,0 +1,34 @@
...
+  snapshotQuery: service(),
+  limit: SNAPSHOT_PAGINATION_COUNT,
+
+  didReceiveAttrs() {
+    this._super(...arguments);
+
+    this.query.perform();
+  },
+
+  query: task(function* () {
```
unicornzero: This function is a really clean read. Nice naming and abstractions.

app/lib/sort-metadata.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  // This computed property allows properties that observe it to basically
+  // observe the store. This is recalculated whenever a snapshot changes in the store.
+  @computed()
+  get _allSnapshots() {
+    return this.store.peekAll('snapshot');
+  }
+
+  @computed('_allSnapshots.@each.reviewState', 'build.id')
```
unicornzero: Could you add a comment that provides a quick overview of the loaded/unloaded flow? I'll dig in to this more tomorrow, but I'm a little confused by it right now.

cadeParade: Added

cadeParade: https://github.com/percy/percy-web/pull/1591/commits/4f2cfdad5e10ef47b85442767cc7820d6b420473

unicornzero: Thanks!! Reading it now.

app/services/reviews.js
```diff
@@ -67,16 +63,11 @@ export default class ReviewsService extends Service {
...
-      reload: true,
-      include: 'approved-by',
-    });
-    const refreshedSnapshots = this.snapshotQuery.getChangedSnapshots(build);
-    const snapshotsComments = this.commentThreads.getCommentsForSnapshotIds(
-      review.snapshots.mapBy('id'),
-      build,
-    );
-    await Promise.all([refreshedBuild, refreshedSnapshots, snapshotsComments]);
+    const refreshedBuild = this._refreshBuild(build);
```
unicornzero: ❤️ cleaner! Nice!

app/templates/components/infinite-snapshot-list.hbs
```diff
@@ -0,0 +1,98 @@
...
+  <InfiniteSnapshotPage
+    @build={{this.build}}
+    @blockItems={{this.blockItems}}
+    @page={{pageToLoad}}
+    as |blocks|
+  >
+    {{#each blocks as |block|}}
+      {{#if (gt block.snapshots.length 1)}}
+        {{#if (eq this.page pageToLoad)}}
+          <div {{in-viewport onEnter=(action (mut this.page) (inc 1 this.page))}}></div>
```
unicornzero: I'm not sure if I'm reading this correctly... so, we're loading pages 0 through this.page, so if pageToLoad equals this.page, then that means we're on the last of the pages we're queued up to load and it's time to load another page once this page is in viewport? Is that right?

unicornzero: Totally minor: It looks like lines 10-12 are identical to lines 31-33. Would it make sense to move them up out of these if blocks to the block above and deduplicate? Not sure if this is a style preference for readability, a functional requirement, or both.

app/components/infinite-snapshot-list.js
cadeParade: There are a few places in the file that update the active block index and I thought having a helper was easier to read...
`_updateActiveBlockIndex(3)` is better than
`this.set('activeSnapshotBlockIndex, 3)` IMO

app/services/comment-threads.js
```diff
@@ -17,6 +17,12 @@ export default class CommentThreadsService extends Service {
   }
 
   getCommentsForSnapshotIds(snapshotIds, build) {
+    const numSnapshots = snapshotIds.length;
+    // If we request too many snapshot ids, the API responds with 400 request too long
+    if (numSnapshots === 0 || numSnapshots > 30) {
+      return this.getCommentsForBuild(build.get('id'));
```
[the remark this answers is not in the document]
cadeParade: Not quite. If there are no snapshots or more than 30 we get ALL comments for build -- don't specify snapshot ids. Otherwise it continues to line 26 and we ask for comments for the specific snapshots provided.

app/models/build.js
cadeParade: I don't really know why that would be needed, but since this property is completely going away when we are fully switched to the new logic, I'm inclined to leave it.

app/lib/sort-metadata.js
```diff
@@ -0,0 +1,215 @@
...
+    this.loadedSnapshots.mapBy('id').forEach(id => {
+      delete allSnapshotItemsById[id];
+    });
+    return allSnapshotItemsById;
+  }
+
+  @computed('loadedSnapshots.@each.reviewState')
+  get unreviewedLoadedSnapshots() {
+    return this.loadedSnapshots.filter(snapshot => {
+      return snapshot.get('isUnreviewed') && !snapshot.get('isUnchanged');
```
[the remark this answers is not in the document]
cadeParade: Have removed.

app/templates/components/infinite-snapshot-list.hbs
cadeParade: Yes, your logic sounds correct. If we're on the last page, we need to load another page.
For code duplication, you are correct, I've consolidated those lines.

app/lib/sort-metadata.js
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  @computed('unloadedSnapshotItemsById.[]')
+  get anyUnloadedSnapshotItemsRejected() {
+    return Object.values(this.unloadedSnapshotItemsById).any(isSnapshotItemRejected);
+  }
+
+  // Returns: {chrome: [snapshotItem, snapshotItem], firefox: [snapshotItem]}
+  @computed('loadedSnapshots.@each.reviewState')
+  get unapprovedSnapshotsCountForBrowsers() {
```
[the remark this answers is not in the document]
cadeParade: Updated

tests/integration/components/infinite-snapshot-group-test.js
```diff
@@ -0,0 +1,382 @@
...
+  describe('width switcher', function () {
+    beforeEach(async function () {
+      // set the widest width comparison to have no diffs to have interesting test behavior.
+      this.get('snapshots').forEach(snapshot => {
+        snapshot.get('comparisons').findBy('width', 1024).set('diffRatio', 0);
+      });
+    });
+
+    // eslint-disable-next-line
+    it('shows widest width with diff as active by default when some comparisons have diffs', async function() {
```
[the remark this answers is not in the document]
cadeParade: This test checks which comparison within a snapshot displays by default -- like maybe the comparison at 1280px did not have a diff but the one at 375px does. This PR checks that the 375px comparison is showing within the snapshot.
The order of the snapshot within the list is not tested in this test.

tests/integration/components/infinite-snapshot-group-test.js
```diff
@@ -0,0 +1,382 @@
...
+      await percySnapshot(this.test, {darkMode: true});
+    });
+
+    it('is collapsed by default when the group is approved', async function () {
+      this.set('snapshots', approvedSnapshots);
+      expect(SnapshotGroup.isExpanded).to.equal(false);
+    });
+
+    it('is expanded when build is approved', async function () {
+      this.set('snapshots', approvedSnapshots);
```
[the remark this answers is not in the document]
cadeParade: Well I think we imagine if we are looking at a build after the build is reviewed, then we probably still want to see the snapshots as opposed to having to click to each one you might be interested in. These decisions were made a long time ago though, so if you have a good argument to change it it's probably worthwhile to discuss again.

tests/integration/components/infinite-snapshot-group-test.js
```diff
@@ -0,0 +1,382 @@
...
+      this.set('build.reviewState', 'approved');
+
+      expect(SnapshotGroup.isExpanded).to.equal(true);
+    });
+
+    it("is expanded when activeSnapshotBlockIndex is to the group's index", async function () {
```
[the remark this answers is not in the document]
cadeParade: yes

tests/integration/components/infinite-snapshot-group-test.js
```diff
@@ -0,0 +1,382 @@
...
+        @activeBrowser={{browser}}
+        @isBuildApprovable={{isBuildApprovable}}
+      />`);
+
+      expect(SnapshotGroup.snapshots.length).to.equal(2);
+      await percySnapshot(this.test, {darkMode: true});
+    });
+
+    // eslint-disable-next-line
+    it('shows unreviewed snapshot without comment before approved snapshot with open comment', async function() {
```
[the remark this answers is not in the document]
cadeParade: Great point, I think I can delete this test???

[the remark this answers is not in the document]
cadeParade: Updated!

tests/integration/components/infinite-snapshot-group-test.js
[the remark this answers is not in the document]
cadeParade: Updated!

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
...
+          build.update({state: BUILD_STATES.FINISHED});
+          return build;
+        }
+      });
+
+      await BuildPage.visitBuild(urlParams);
+
+      // We approved the snapshot that would normally be seen as first (default snapshot).
+      // So the normal second snapshot (twoWidthsSnapshot) will now be first, and defaultSnapshot
+      // will be second.
```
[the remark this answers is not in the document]
cadeParade: This whole file has the flag turned on here: https://github.com/percy/percy-web/pull/1591/files/4f2cfdad5e10ef47b85442767cc7820d6b420473#diff-b78cca0826f952facbc0a22a2bd204e5R95

The snapshots for this test are being sorted via the terrible snapshot sorting logic I wrote to be returned from mirage endpoints (https://github.com/percy/percy-web/pull/1591/files/4f2cfdad5e10ef47b85442767cc7820d6b420473#diff-f2a16300c690ebbd321f2b8198a0656e) (added automatically for snapshot requests asking for metadata here: https://github.com/percy/percy-web/pull/1591/files/4f2cfdad5e10ef47b85442767cc7820d6b420473

The sort logic i wrote for the tests is not as sophisticated as the whole of the logic in the API, but it does sort per browser on approval state, groups, and diff ratio. These were the smallest set of criteria I could find that would allow most existing tests to pass while not rewriting the entire sort algorithm.

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
...
+      );
+
+      await firstCommentThread.goToOriginatingSnapshot();
+
+      expect(currentRouteName()).to.equal('organization.project.builds.build.snapshot');
+      await percySnapshot(`${this.test.fullTitle()} | after following originating snapshot link`);
+    });
+
+    // eslint-disable-next-line
+    it('blocks approval of snapshot if there are open review threads on snapshot', async function () {
```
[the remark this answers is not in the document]
cadeParade: Yes, the open review thread check still works for snapshots in the store, but will not block on snapshots not in the store. It is a rare case, but is technically possible.

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
...
+      await BuildPage.visitBuild(urlParams);
+      expect(BuildPage.browserSwitcher.chromeButton.diffCount).to.equal('3');
+      expect(BuildPage.browserSwitcher.firefoxButton.diffCount).to.equal('3');
+      await percySnapshot(this.test.fullTitle() + ' before switching browsers');
+      await BuildPage.browserSwitcher.switchBrowser();
+      await percySnapshot(this.test.fullTitle() + ' after switching browsers');
+    });
+
+    it('sorts snapshots correctly when switching to another browser', async function () {
+      // Change diff ratio in one browser so the sort behavior is different in the other browser.
```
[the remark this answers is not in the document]
cadeParade: I've written the test sorting algorithm to take into account diff ratio, so this will be ok when the other is deleted. The mirage endpoint should return different sort metadata per browser based on diff ratio so this test is just making sure the page updates correctly when switching browsers based on the sort metadata

tests/acceptance/infinity-build-test.js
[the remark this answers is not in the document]
cadeParade: Actually, this will always work because this is on the snapshot level. If you are clicking "Approve" on an individual snapshot, it will always be in the store so we will always know whether there are open review threads on it.
The weird case comes when we are approving an entire build.

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
...
+        );
+        await thirdSnapshot.collaborationPanel.newComment.submitNewThread();
+
+        expect(thirdSnapshot.commentThreads.length).to.equal(1);
+        expect(thirdSnapshot.header.numOpenCommentThreads).to.equal('1');
+
+        await percySnapshot(this.test, {darkMode: true});
+      });
+
+      it('blocks approval of group if there are open review threads on group', async function () {
```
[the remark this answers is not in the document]
cadeParade: This is on the group level. So if you are clicking "Approve" on a group, you will always have all the snapshots in the group in the store. So this test is still valid. As above, it is the approving an entire build case that gets weird.

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
...
+        });
+        await BuildPage.visitBuild(urlParams);
+        await percySnapshot(this.test, {darkMode: true});
+      });
+    });
+
+    describe('when there is one snapshot missing', function () {
+      it('displays correctly', async function () {
+        await BuildPage.visitBuild(urlParams);
+        await percySnapshot(this.test, {darkMode: true});
```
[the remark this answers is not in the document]
cadeParade: Yeah fair enough. The title is usually sufficient for these cases imo, but I do like usually having a selector in addition.

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
...
+    expect(snapshotReview.action).to.equal('request_changes');
+    expect(snapshotReview.buildId).to.equal(build.id);
+    expect(snapshotReview.snapshotIds).to.eql([defaultSnapshot.id]);
+    expect(firstSnapshot.commentThreads.length).to.equal(1);
+    expect(firstSnapshot.commentThreads[0].wasRejectedPreviously).to.equal(false);
+    await percySnapshot(this.test, {darkMode: true});
+  });
+
+  it('blocks approval of snapshot when snapshot is rejected', async function () {
+    await BuildPage.visitBuild(urlParams);
```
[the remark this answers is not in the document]
cadeParade: As above, this is related to a single snapshot approval, so it will always be in the store, so it is safe to assume we will know it's state.

tests/acceptance/infinity-build-test.js
```diff
@@ -0,0 +1,1093 @@
...
+    await percySnapshot(this.test, {darkMode: true});
+  });
+
+  it('displays reintroduced snapshot', async function () {
+    server.create('snapshot', 'new', 'reintroduced', {build, name: 'ohai'});
+    await BuildPage.visitBuild(urlParams);
+    await percySnapshot(this.test, {darkMode: true});
+  });
+
+  it.skip('loads more snapshots on scroll', async function () {
```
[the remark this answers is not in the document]
cadeParade: uhhh shoot this test 🤦‍♀

PR #1595, opened 2020-03-12
Remove snapshot endpoint url construction from application adapter

cadeParade:
## Description
In a very long term project to remove all URL construction blocks from [application/adapter.js](https://github.com/percy/percy-web/blob/master/app/adapters/application.js#L34) file, I have taken the first step and removed the snapshot endpoint construction.

This was made possible by a small API change: https://github.com/percy/percy-api/pull/1395/files#diff-21497849d8f00507c9c8dcaf6288b136R84


app/services/tooltips.js
```diff
@@ -104,8 +107,7 @@ export default class TooltipsService extends Service {
     if (!build) {
       return GROUP_SEQUENCE;
     }
-
-    const snapshots = await build.get('snapshots');
```
cadeParade: This was calling `api/v1/builds/id/snapshots` every time you transitioned away from a build page. That means, it was asking for _every_ snapshot in a build whenever you go back to the project page. The request is non-blocking but as we know, _every_ snapshot in a build is usually a LOT and there's no need for it here. I'm not sure what this is for tbh, but at least I have severely clamped down on the scope of the query in the update.

app/adapters/application.js
```diff
@@ -99,11 +99,6 @@ export default class Application extends JSONAPIAdapter.extend(DataAdapterMixin)
       return utils.buildApiUrl('subscription', snapshot.record.get('organization.slug'));
     }
 
-    if (requestType === 'query' && modelName === 'snapshot') {
```
unicornzero: wooot! <3

PR #1620, opened 2020-04-06
Paginated unchanged snapshots

cadeParade:
## Description
The first iteration of the API snapshot sort update was to get changed snapshots paginating via sort metadata.
This update allows unchanged snapshots to be paginated via sort metadata. When a user click to show unchanged snapshots, we make a request to the API for unchanged snapshots to include sort metadata. The unchanged snapshot list then uses the same infinite scroll "page" technique as the change snapshots to load 30 into the store at a time. 

I chose 30 to load because unchanged snapshots are collapsed by default so many more fit on the screen at once.

The list keyboard nav has been updated to work with the unchanged snapshots also. Since the sort metadata includes an index, it is easy to know which snapshot should be "next" and "previous" in the list.
To have the keyboard nav work with unchanged snapshots, after we fetch the unchanged sort metadata, we update the index of the unchanged items to start at the last index + 1 of changed items.


app/components/build-container.js
```diff
@@ -112,15 +126,23 @@ export default Component.extend(PollingMixin, {
...
     this.set('isUnchangedSnapshotsVisible', false);
   },
 
   init() {
     this._super(...arguments);
+    // TODO(sort) Remove this when old method is deprecated.
     this.snapshotsUnchanged = this.snapshotsUnchanged || [];
   },
 
+  fetchUnchangedSnapshots: task(function* () {
```
unicornzero, line 140: Just noting that we chatted on our call about possibly moving this method to the controller.

app/lib/sort-metadata.js
```diff
@@ -79,7 +80,7 @@ export default class MetadataSort extends EmberObject {
     return defaultBrowserData ? defaultBrowserData.browser_family_slug : 'chrome';
   }
 
-  @computed()
+  @computed('sortData')
```
unicornzero, line 83: Noting that we chatted about maybe changing the property name to indicate this is for changed items

app/lib/sort-metadata.js
```diff
@@ -192,6 +204,27 @@ export default class MetadataSort extends EmberObject {
       };
     });
   }
+
+  handleUnchangedSortData(unchangedSortData) {
+    unchangedSortData.forEach(browserData => {
+      const browserSlug = browserData.browser_family_slug;
+      const changedSortData = this.sortData.findBy('browser_family_slug', browserSlug);
+      const lastIndex = this._lastIndex(changedSortData);
```
unicornzero, line 212: What would you think of changing this to:
`const startingIndex = this._lastIndex(changedSortData) + 1;`

This should allow simplifying line 214 to:
`set(item, 'index', startingIndex + i );`

app/lib/sort-metadata.js
```diff
@@ -192,6 +204,27 @@ export default class MetadataSort extends EmberObject {
...
+      browserData.items.forEach((item, i) => {
+        set(item, 'index', lastIndex + i + 1);
+      });
+    });
+    set(this, 'unchangedSortData', unchangedSortData);
+  }
+
+  _lastIndex(changedSortData) {
```
unicornzero, line 220: As an alternative to my above suggestion, this could change to `_nextIndex`

app/templates/components/infinite-snapshot-list.hbs
```diff
@@ -54,28 +46,45 @@
...
-          @updateActiveSnapshotBlockId={{action "updateActiveSnapshotBlockId"}}
-          @activeBrowser={{this.activeBrowser}}
-          @updateActiveBrowser={{this.updateActiveBrowser}}
-          @isBuildApprovable={{this.isBuildApprovable}}
-          @closeCommentThread={{this.closeCommentThread}}
-          @createCommentThread={{this.createCommentThread}}
-          @createComment={{this.createComment}}
-        />
-      </div>
+    {{#each (range 0 this.noDiffsPage true) as |noDiffsPageToLoad|}}
```
unicornzero, line 49: Does `this.noDiffsPage` mean the last noDiffs page to load in this batch? If so, what about renaming this (and the equivalent for changed snapshots) to something like `this.noDiffsPageLimit` or `this.lastNoDiffsPageToLoad`? not sure... :/

unicornzero: We might want to try to consider trying to phase out either `noDiffs` or `unchanged` in our naming. Maybe.

app/templates/organization/project/builds/build/index.hbs
```diff
@@ -8,5 +8,6 @@
   @isBuildApprovable={{this.isBuildApprovable}}
   @notifyOfUnchangedSnapshots={{action "notifyOfUnchangedSnapshots"}}
   @fetchSnapshots={{action "fetchSnapshots"}}
+  @fetchUnchangedSnapshots={{this.fetchUnchangedSnapshots}}
```
unicornzero, line 11: Totally optional: For parity, what would you think of changing `fetchSnapshots` to `fetchChangedSnapshots`? Worth the effort?

app/lib/sort-metadata.js
```diff
@@ -94,7 +106,7 @@ export default class MetadataSort extends EmberObject {
   // }
   @computed()
   get allSnapshotItemsById() {
```
unicornzero, line 108: Totally optional: Is this always specific to changed snapshots? If so, would it make sense to name it `allChangedSnapshotItemsById`? Maybe my suggestion is too verbose. :/

PR #1626, opened 2020-04-10
Update group indexes in snapshot list.

cadeParade:
## Description
If you go to https://percy.io/percy/percy-web/builds/4837745 and click the diff toggle for the second group, it will send your scroll position to the wrong place. I am attempting to fix this with this PR and guessing that the indexes within the groups are interfering with the overall index from sort metadata.
Unfortunately I cannot reproduce this locally, so I am not sure it will work. 
The problem does not occur all the time though, it is only when
1) they have the api sort feature flag AND
2) there is an expanded group


app/components/infinite-snapshot-list-item.js
```diff
@@ -49,7 +49,9 @@ export default Component.extend({
   shouldFullyRender: true,
 
   isActiveSnapshotBlock: computed('activeSnapshotBlockIndex', 'index', function () {
-    return this.activeSnapshotBlockIndex === this.index;
+    const activeIndex = this.activeSnapshotBlockIndex;
```
cadeParade, line 52: Do not return true if both `activeSnapshotBlockIndex` and `index` are undefined.

